PR #113, opened 2015-10-22
Removes the EMAIL and URL fields from tcl-commands.doc

vanosg:
Removes the EMAIL and URL fields from tcl-commands.doc . these were removed
some time ago in code, but documentation appears not to have been updated
along with it. 

Example:

---

<user> .tcl getuser user LASTON

<bot> Tcl: 1413684437 partyline

<user> .tcl getuser user EMAIL

<bot> Tcl error: No such info type: EMAIL

<user> .tcl getuser user URL

<bot> Tcl error: No such info type: URL

### DIFF
--- a/doc/Changes1.8
+++ b/doc/Changes1.8
@@ -5,6 +5,9 @@ Eggdrop Changes (since version 1.8.0)
     _____________________________________________________________________
 
 1.8.0 (CVS):
+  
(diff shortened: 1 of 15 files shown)


thommey:
Those were moved to scripts/userinfo.tcl, perhaps just clarify in documentation? It ships with eggdrop.

Issue #92, opened 2012-02-08
autoconf TCLLIB detection might fail


thommey:
| Trac | Data |
| --: | :-- |
| Ticket | 92 |
| Reported by | @thommey |
| Status | new |
| Component | Build process |
| Priority | major |
| Milestone | 1.8.0 |
| Version | 1.8.0 CVS |

"/usr/local/lib/libtcl85.so.1.1" is an example filename which will cause us to link -ltcl85.so.1

proposed fix: either #91 (preferred), or something like (aclocal.m4)

 TCLLIBEXT=`echo $LIB | awk '{print substr($1, match($1,/\.so|\.a/))}'`
instead of

 TCLLIBEXT=.`echo $LIB | awk '{j=split($1, i, "."); print i[j]}'`

reported by stu

Issue #91, opened 2012-02-08
use tclConfig.sh to find tcl


thommey:
| Trac | Data |
| --: | :-- |
| Ticket | 91 |
| Reported by | @thommey |
| Status | assigned |
| Component | Build process |
| Priority | major |
| Milestone | 1.8.0 |
| Version | 1.8.0 CVS |

we should be using tclConfig.sh, where available, to locate tcl headers and the library

thommey:
@vanosg:
http://lists.eggheads.org/pipermail/patches/2015-February/001455.html

PR #117, opened 2015-10-22
Incorporate tclConfig.sh

vanosg:
Issue #91 #92 

This patch incorporates the tcl.m4 file provided by the TCL maintainers into
the autoconf process. It uses these macros, via tclConfig.sh (standard on
TCL installs), to locate where libraries are installed on the local machine.
In theory, this will negate the need for users to ever manually specify the
paths to their TCL install (and would have avoided the issues with Ubuntu
maintainers changing the default install path in 14.4, for example). 

The -with-tcllib and --with-tclinc flags were left in for backwards
compatibility with any install scripts people might be using (that's a
stretch, but it's there), or for example, if the user had a local install of
TCL and wanted to use that path over the default paths. If used, these flags
will override the paths specified in tclConfig.sh (Of course, the user could
also just point to the tclConfig.sh from that installation as well, but hey,
flexibility)

By incorporating this patch, eggdrop NOW REQUIRES a tclConfig.sh file to
compile. Because the tcl.m4 file completely exits the autoconf process if
tclConfig.sh is not found, it must exist for configure to complete. I don't
think requiring this should be an issue, as it is placed as part of the
install process, you would have to proactively remove it (why, I don't know)
in order for it to not be present if TCL is installed. As of now, we're just
importing the code tcl.m4 file directly as provided with no modification,
but if this is something that becomes an issue, we would simply have to
modify the exit code in the eggdrop-used tcl.m4 and maintain that change
through any new incorporations of tcl.m4. 

Also, tcl.m4 does not look in $HOME for tclConfig.sh . If we modify the exit
code, we may also want to consider adding a check for $HOME by default as
well, or modifying the exit code to explicitly tell the user that TCL was
not found (the tcl.m4 code could be construed as a but ambiguous to an
unfamiliar user).

---

First test case- no options specified at all:

$ ./configure
---snip---
checking for Tcl configuration... found /usr/local/lib/tclConfig.sh
checking for existence of /usr/local/lib/tclConfig.sh... loading
checking for Tcl library... using /usr/local/lib/libtcl8.5.so from
tclConfig.sh
checking for Tcl header... using /usr/local/include/tcl.h from tclConfig.sh
---snip---

Good basic test. 
Next, we test where we specify an incorrect path for tclConfig.sh

$ ./configure --with-tcl-config=/foo
---snip---
checking for Tcl configuration... configure: error: /foo directory doesn't
contain tclConfig.sh
<exits>
---snip---

The path we specify doesn't exist, but clearly we're trying to tell it where
to look, so lets exit and try again 

Now, with only tcllib (or tclinc):

$ ./configure -with-tcllib=/foo
---snip---
configure: WARNING:

  You must specify both --with-tcllib and --with-tclinc for either to work.

  configure will now attempt to autodetect both the Tcl library and header.

checking for Tcl configuration... found /usr/local/lib/tclConfig.sh
checking for existence of /usr/local/lib/tclConfig.sh... loading
checking for Tcl library... using /usr/local/lib/libtcl8.5.so from
tclConfig.sh
checking for Tcl header... using /usr/local/include/tcl.h from tclConfig.sh
---snip---

Still works the same as it has in the past. now lets specify both settings
with incorrect paths

$  ./configure --with-tcllib=/foo --with-tclinc=/bar
---snip---
checking for Tcl configuration... found /usr/local/lib/tclConfig.sh
checking for existence of /usr/local/lib/tclConfig.sh... loading
configure: WARNING:

  The file '/foo' given to option --with-tcllib is not valid.
  Specify the full path including the file name (e.g. /usr/lib/libtcl8.5.so)

  configure will now attempt to autodetect both the Tcl library and header.

checking for Tcl library... using /usr/local/lib/libtcl8.5.so from
tclConfig.sh
checking for Tcl header... using /usr/local/include/tcl.h from tclConfig.sh
--snip---

This also continues to work as it has in the past, where we try to find it
for the user anyway. 

Now, what if we want to use a different tcl file than what tclConfig wants
us to (perhaps in the case of an older version being installed by the admin,
and the user installs a recent version) the --with-tcllib/tclinc settings
override what is in tclConfig.sh

$ ./configure --with-tcllib=/home/user/libtcl8.5.so
--with-tclinc=/usr/local/include/tcl.h
---snip---
checking for Tcl configuration... found /usr/local/lib/tclConfig.sh
checking for existence of /usr/local/lib/tclConfig.sh... loading
checking for Tcl library... using /home/user/libtcl8.5.so
checking for Tcl header... using /usr/local/include/tcl.h
---snip---

The --with-tcl[lib,inc] settings were given precedence over the tclConfig.sh
input.
Finally, what if tclConfig.sh isn't found at all?

$ ./configure
---snip---
checking for Tcl configuration... configure: error: Can't find Tcl
configuration definitions. Use --with-tcl to specify a directory containing
tclConfig.sh
(exit)
---snip---

Those are all the test cases I can think of.             

---

Also, I updated the recommended TCL version to 8.6 from 8.5


thommey:
revise routine to search tclConfig.sh (/usr/lib/tcl8.6/tclConfig.sh wasn't found automatically), check http://wiki.tcl.tk/12968

vanosg:
Can you expand on the error you encountered? Or what you're proposing? Your comment is a little abbreviated, thanks

vanosg:
Fixed.

PR #121, opened 2015-10-22
various compile warning fixes

vanosg:
Hi there!

I've been using eggdrop 1.8.0 (CVS) for the past year and finally got  
the time to fix a few compilation warnings that I've seen with recent  
versions of GCC. The patches are generated using a checkout as of  
today, 2015-05-16.
Patch 1, 2 and 3 are more serious than number 4. Patch 4 isn't really  
a fix either, it's just a way to hide the warnings, the other patches  
are fixes however.

If you have any questions or comments, please tell them and I'll see  
if I can rework the patches to match your requirements.

Best regards,
Tobbe

### DIFF
--- a/doc/Changes1.8
+++ b/doc/Changes1.8
@@ -6,6 +6,9 @@ Eggdrop Changes (since version 1.8.0)
 
 1.8.0 (CVS):
 
+  - Fix some compiler warnings
+    Patch by: Tobbe
+
   - Throw error when writing to read-only variables in server module.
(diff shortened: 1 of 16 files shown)


thommey:
Could you make a pull request for the patches from Tobbe excluding the return values one? ("Patch 4")
I'm not a fan of this return value error suppression by hacking the ignore in, it should at least be a separate PR/commit. Thanks.

Issue #123, opened 2015-10-22
GCC5 breaks compile


vanosg:
With GCC 5, it now uses GNU11 as the default rules for compilation. As a result, after several errors during compike, existing eggdrop code now exits with this error:

```
main.o: In function `garbage_collect':
/home/user/eggdrop1.8/src/./main.c:737: undefined reference to `garbage_collect_tclhash'
modules.o:(.data+0x2a0): undefined reference to `open_listen'
tcldcc.o: In function `tcl_listen':
/home/user/eggdrop1.8/src/tcldcc.c:957: undefined reference to `open_listen'
collect2: error: ld returned 1 exit status
Makefile:37: recipe for target '../eggdrop' failed
make[1]: *** [../eggdrop] Error 1
make[1]: Leaving directory '/home/user/eggdrop1.8/src'
Makefile:240: recipe for target 'debegg' failed
make: *** [debegg] Error 2
```

thommey:
Proposed patch probably breaks gnu89 compilation, I'll see about this and possibly remove inline completely.

PR #124, opened 2015-10-22
Fix gcc 5 compile errors

vanosg:
Regarding issue #123 

Referencing https://gcc.gnu.org/gcc-5/porting_to.html, this patch removes the errors. It has been tested on gcc 4.8.4 (ubuntu 14.04) and gcc 5.2.1 (Debian sid/testing).


thommey:
Will fix this another way, this risks breaking gnu89 compilation.

vanosg:
I compiled it under gnu89 and it worked... what are you worried about in particular?

PR #127, opened 2015-10-24
Clarify SSL documentation and errors

vanosg:
This patch does three things: 
-          Updates error messages and documentation to identify issues with
  sslcert not installed to the proper locations if DEST= was used in the make
  install process,
-           clarify the eggdrop.conf file SSL documentation
-          Add debug error messages better identifying issues with ssl certs
-          Generally update SSL docs to make them clearer to folks who
  didn't write the code ;)

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -38,6 +38,7 @@ extern int tls_vfydcc;
 extern struct dcc_t *dcc;
 
 int tls_maxdepth = 9;         /* Max certificate chain verification depth     */
+int ssl_files_loaded = 1;     /* Check for loaded SSL key/cert files          */
 SSL_CTX *ssl_ctx = NULL;      /* SSL context object                           */
 char *tls_randfile = NULL;    /* Random seed file for SSL                     */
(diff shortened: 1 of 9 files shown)


src/tls.c
```diff
@@ -140,9 +141,11 @@ int ssl_init()
   }
   /* Load our own certificate and private key. Mandatory for acting as
      server, because we don't support anonymous ciphers by default. */
-  if (SSL_CTX_use_certificate_chain_file(ssl_ctx, tls_certfile) != 1)
+  if (SSL_CTX_use_certificate_chain_file(ssl_ctx, tls_certfile) != 1) {
+    ssl_files_loaded = 0;
```
Robby-, line 145: Shouldn't this also have some logic to change it back to 1 after a .rehash? Currently this looks like it would require a full restart.

vanosg:
Loading these files is part of ssl_init() , which is only called on restart (or startup). Thus, changing the settings in the config file and rehashing would not load them, and the 0 will remain true.

PR #187, opened 2016-02-19
Fix SSL DNS name matching for telnet

### LINKED ISSUE #98: SSL: Certificate DNS name matching doesn't work in telnet sessions
| Trac | Data |
| --: | :-- |
| Ticket | 98 |
| Reported by | manuel |
| Status | accepted |
| Component | Core |
| Priority | minor |
| Milestone | 1.8.0 |
| Keywords | ssl |
| Version | 1.8.0 CVS |

ssl_handshake() get's called with host=NULL thus ssl_verifycn() will never work. This is because the ssl_handshake() gets called before the dns lookup completes. In order to fix this we have to move the ssl_handshake() into the dcc_telnet_hostresolved()-callback. Additional we need to make sure to not read any data from the socket as this will remove the SSL client-hello from the socket buffer.

Patch attached.

vanosg:
Patch by: Manuel
Fixes #98 

User-submitted patch; not sure how to test!

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -1238,13 +1238,22 @@ static void dcc_telnet(int idx, char *buf, int i)
   dcc[i].sock = sock;
   dcc[i].port = port;
 #ifdef TLS
-  if (dcc[idx].ssl && ssl_handshake(sock, TLS_LISTEN, tls_vfyclients,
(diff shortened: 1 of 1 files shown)


thommey:
Merged

Issue #100, opened 2012-09-19
SSL: verify flags don't work at all


thommey:
| Trac | Data |
| --: | :-- |
| Ticket | 100 |
| Reported by | manuel |
| Status | new |
| Component | Core |
| Priority | major |
| Milestone | 1.8.0 |
| Keywords | ssl |
| Version | 1.8.0 CVS |

This is due to the following problems:
- verify-parameter isn't stored in data->verify inside ssl_handshake()
- data->flags is used where data->verify would be correct
- TLS_VERIFYISSUER and TLS_VERIFYCN + TLS_VERIFYTO and TLS_VERIFYFROM are mixed.

I'll attach/paste a patch in the next comment.

thommey:
manuel:
The patch inverts the exception-flags in data->verify to better reflect the names of the TLS_VERIFY*-macros. This results in the following meaning:
- verify & TLS_VERIFYPEER ...verify the peer
- verify & TLS_VERIFYISSUER ...verify the issuer
- verify & TLS_VERIFYCN ...verify the common name/alt subj names
- etc...
  this wasn't the case before

NOTE: patch only tested with ssl telnet. but I see no obvious reason why this should break bots and dcc.

Patch:

 --- eggdrop1.8.orig/src/eggdrop.h  2010-10-19 15:27:29.000000000 +0200
 +++ eggdrop1.8/src/eggdrop.h   2012-09-19 13:09:18.000000000 +0200
 @@ -708,10 +708,10 @@

  /\* TLS verification flags */
  #  define TLS_VERIFYPEER        0x00000001
 -#  define TLS_VERIFYCN            0x00000002
 -#  define TLS_VERIFYISSUER      0x00000004
 -#  define TLS_VERIFYFROM        0x00000008
 -#  define TLS_VERIFYTO          0x00000010
 +#  define TLS_VERIFYISSUER      0x00000002
 +#  define TLS_VERIFYCN          0x00000004
 +#  define TLS_VERIFYTO          0x00000008
 +#  define TLS_VERIFYFROM        0x00000010
  #  define TLS_VERIFYREV         0x00000020

  /\* Context information to attach to SSL sockets */
 --- eggdrop1.8.orig/src/tls.c  2011-02-04 11:15:12.000000000 +0100
 +++ eggdrop1.8/src/tls.c   2012-09-19 13:49:03.000000000 +0200
 @@ -544,8 +544,8 @@
[...]

PR #206, opened 2016-04-09
Fix DNS Certificate Matching

### LINKED ISSUE #98: SSL: Certificate DNS name matching doesn't work in telnet sessions

vanosg:
Found by: manuel
Patch by: manuel
Fixes: #98 

This is a patch submitted by a user - I have not been able to test anything more than the fact that it compiles. 

@thommey , "NOTE: TLS_VERIFYCN configurations value is inverted"  ... can you confirm that? If so, and it needs fixing, can you please submit a separate issue for it and assign it to me? Thanks!

EDIT: Never mind - this comment is fixed via #100 ... we'll need to make sure the logic matches between the two patches

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -1238,13 +1238,22 @@ static void dcc_telnet(int idx, char *buf, int i)
   dcc[i].sock = sock;
   dcc[i].port = port;
 #ifdef TLS
-  if (dcc[idx].ssl && ssl_handshake(sock, TLS_LISTEN, tls_vfyclients,
-      LOG_MISC, NULL, NULL)) {
-    killsock(sock);
-    lostdcc(i);
-    return;
-  }
   dcc[i].ssl = dcc[idx].ssl;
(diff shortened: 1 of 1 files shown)


vanosg:
build

vanosg:
build

thommey:
Merged

PR #275, opened 2016-10-13
Completely reset channels upon server disconnect, fixes incorrect botonchan results.

Cizzle:
Found by: Cizzle
Patch by: Cizzle
Fixes: any command relying on channel data like "botonchan".

One-line summary: Fix channel data not being reset after certain disconnects and thus any command relying on this data, like "botonchan".

Additional description (if needed): This is especially a problem for invite-only channels where the bot normally asks for an invite to rejoin. Because "botonchan" still returns 1, it thinks it's on the channel and thus doesn't ask an invite. Moreover it keeps trying to message the channel, tries to invite other users and bots etc...

1. Enable tcl on partyline
2. Have your bot join an invite-only channel.
3. Disable any network activity so it pings out, the following works best: ifconfig IFACE down
4. Wait til the bot says the server got stoned and tries another server, then re-enable the network: service networking restart (worked better than just "ifconfig IFACE up")
5. When the bot eventually is connected to an IRC server again, execute on partyline: .tcl botonchan #CHANNEL 
6. Issue ".tcl resetchan #CHANNEL" to have it figure out it's not actually on the channel

### DIFF
--- a/src/mod/server.mod/servmsg.c
+++ b/src/mod/server.mod/servmsg.c
@@ -359,7 +359,7 @@ static int got442(char *from, char *msg)
 
     putlog(LOG_MISC, chname, IRC_SERVNOTONCHAN, chname);
     if (me && me->funcs)
-      (me->funcs[CHANNEL_CLEAR]) (chan, 1);
+      (me->funcs[CHANNEL_CLEAR]) (chan, CHAN_RESETALL);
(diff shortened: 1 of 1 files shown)


thommey:
Hey, I merged this (not closed, that is a github error) as  321c219. Thank you very much for your contribution :)

PR #281, opened 2016-10-20
Add CBC mode to blowfish and use it by default.

Cizzle:
Found by: Cizzle
Patch by: Cizzle
Fixes: ECB mode

One-line summary: makes the default encryption module, blowfish, use CBC mode by default instead of the more insecure ECB mode.

Additional description (if needed): When using "encrypt" or "decrypt", the key can have the "ecb:" or "cbc:" prefix to explicitly set a mode to use. When encrypting, CBC mode is used by default unless the key has "ecb:" (case insensitive). When decrypting however, CBC mode is only used by default if the given string starts with "*", even when the key has "cbc:" prefixed.

### DIFF
--- a/src/mod/blowfish.mod/blowfish.c
+++ b/src/mod/blowfish.mod/blowfish.c
@@ -280,11 +280,8 @@ static int base64dec(char c)
 
 static int cbcbase64dec(char c)
 {
-  int i = 64;
-
-  while (i >= 0 && cbcbase64[i] != c)
-    --i;
-  return i;
+  char *i = strchr(cbcbase64, c);
+  return i ? (int)(i - cbcbase64) : -1;
 }
 
 static void blowfish_encrypt_pass(char *text, char *new)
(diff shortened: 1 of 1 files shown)


Cizzle:
I wasn't sure where to document the ECB / CBC key usage; in doc/tcl-commands.doc (and it's derivatives) or doc/settings/mod.blowfish; so I didn't add this to this PR.

vanosg:
Hi Cizzle, thanks for another PR! If you're ever bored, come say hi in #eggdrop on Freenode sometime. We're currently feature-frozen for 1.8.0, but we'll evaluate this for 1.8.1 as soon as we go stable on the current RC (we're soooo close, I promise!). So, please don't take inactivity on this PR as lack of interest, we'll get to it shortly!

.... and I can't wait for you to submit an AES module to replace this altogether :) Thanks again!

Cizzle:
Well, ECB mode could be considered a security flaw which is fixed by CBC mode? :)

As for AES, this already exists? ftp://ftp.eggheads.org/pub/eggdrop/modules/1.8/rijndael1.0-pseudo.tar.gz
Though I haven't tested it, I already see the use of "ncrypt" in rijndael.c while "encrypt" is still used in rijndael_cipher.c.

src/mod/blowfish.mod/blowfish.c
```diff
@@ -276,6 +278,15 @@ static int base64dec(char c)
   return 0;
 }
 
+static int cbcbase64dec(char c)
```
thommey, line 281: Suggesting
```c
 static int cbcbase64dec(char c)
 {
  char *i = strchr(cbcbase64, c);
  return i ? (int)(i - cbcbase64) : 0;
 }
```
instead.

PR #312, opened 2016-12-25
Fix few more issues with handlen, {} and pidfile.

Cizzle:
Found by: thommey, Geo, Cizzle
Patch by: Cizzle
Fixes: 

One-line summary: Fixes vars using {}, \ as 31th char in nick, incorrect handlen parsing and reverts pidfile fix which caused literal varnames in the generated botchk file.

 - Use a handlen < 10.
 - Use a nick or other var with {} instead of "", with or without [] enclosed.
 - Use $nick as botnet-nick and check "botname" in generated botchk file.
 - Use nick with \ as char 31 (and any other char as 32nd) and check "botname" in generated botchk file (and try running it).


vanosg:
LGTM! :+1:

Issue #301, opened 2016-11-17
"Invalid Argument" error


vanosg:
If  my-ip, my-hostname, vhost4, or vhost6 are improperly set, the incredibly helpful "Invalid Argument" error is returned. Recommend catching the error and replacing with something slightly more useful.

Cizzle:
Could you tell when you see those errors? I've set vhost4 to "999.99.99.0" which is in my eyes "improperly set", but I see no error when starting the bot or on partyline. Aren't my-ip and my-hostname deprecated?

vanosg:
From thommey-

Failed connect to chat.freenode.net (Cannot assign requested address) 
  * valid IP that is not bound to an interface (random IP)

Failed connect to chat.freenode.net (Invalid argument) 
  * assigning 127.0.0.1 as vhost but trying to connect to an IP outside of that netmask (not 127/8, e.g. chat.freenode.net)

PR #316, opened 2016-12-28
Return more user-friendly message with incorrect vhost4/6 setting in config.

Cizzle:
Found by: Geo, thommey
Patch by: Cizzle
Fixes: #301 

One-line summary: Returns more user-friendly messages when vhost4, vhost6 or the deprecated my-ip or my-hostname are incorrectly set in config.

Instead of "Invalid argument" it will now reply "Vhost set in config can not be used to connect to this address".
Instead of "Cannot assign requested address" it will now reply "Vhost set in config is not available on this machine".
The danish, finnish, french and german language files will need an update.

 - Set "vhost4" in config to "127.0.0.1" and try to connect to any public IRC server for the "Invalid argument" reply.
 - Set "vhost4" in config to "1.2.3.4" (or another address if this one would be valid on your machine) and try to connect to any public IRC server for the "Cannot assign requested address" reply.

### DIFF
--- a/language/core.danish.lang
+++ b/language/core.danish.lang
@@ -205,8 +205,8 @@ Telnet botten og skriv 'NEW' som dit nick.\n
 0x647,Modekø er på
 0x648,Serverkø er på
 0x649,Hjælpekø er på
-# 0x64a - unused
-# 0x64b - unused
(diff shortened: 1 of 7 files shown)


vanosg:
LGTM :+1:

vanosg:
build

PR #317, opened 2016-12-29
adding portugueselang

mai0li:


mai0li:
adding portuguese support

Issue #303, opened 2016-12-05
compilation warning when patch() is not used (in stable versions)


thommey:
in full versions we don't call patch() which gives a compile warning, might be obsolete if we replace how we do patchlevels so that they're stable (don't require a separate commit, don't change the git hash)

PR #319, opened 2017-01-02
Bug/patchwarn

vanosg:
Found by: thommey
Patch by: Geo
Fixes: #303 

One-line summary: Change releaseprep patch version string to "" for stable release.

No compile error, and correct version string produced
```
$ ./eggdrop -v
Eggdrop v1.8.0 (C) 1997 Robey Pointer (C) 2010-2016 Eggheads
```

### DIFF
--- a/misc/releaseprep
+++ b/misc/releaseprep
@@ -86,6 +86,8 @@ fix_patch_h() {
 		cat src/patch.h | sed 's/^patch.*Git.*Git version \*\//patch("PRE-RELEASE");           \/* RC version *\//g' > src/patch.h_
 	else
 		cat src/patch.h | sed 's/^patch.*/\/* PATCH GOES HERE *\//g' > src/patch.h_
+		mv src/patch.h_ src/patch.h
+		cat src/patch.h | sed -e '0,/^\/\* PATCH GOES HERE \*\//s//patch("");/' > src/patch.h_
 	fi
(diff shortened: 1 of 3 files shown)


misc/releaseprep
```diff
@@ -86,6 +86,8 @@ fix_patch_h() {
 		cat src/patch.h | sed 's/^patch.*Git.*Git version \*\//patch("PRE-RELEASE");           \/* RC version *\//g' > src/patch.h_
 	else
 		cat src/patch.h | sed 's/^patch.*/\/* PATCH GOES HERE *\//g' > src/patch.h_
+		mv src/patch.h_ src/patch.h
+		cat src/patch.h | sed -e '0,/^\/\* PATCH GOES HERE \*\//s//patch("");/' > src/patch.h_
```
Cizzle, line 90: Holy cannoli, please kill the cats. I also don't know the machine where the releases are prepared on, but I'd be surprised if the sed-binary on it wouldn't have the -i functionality. To top it off, those two sed-invocations can be combined with a mere ';'.
This results in the above 3 lines becoming:
`sed -i 's/^patch.*/\/* PATCH GOES HERE *\//g;0,/^\/\* PATCH GOES HERE \*\//s//patch("");/' src/patch.h`
and obsoleting the line below (under condition that the "cat | sed > *_" scheme is changed in the first if-clause).

The above comments apply for this whole script.

Issue #337, opened 2017-02-21
Versions ($::numversion)


ghost:
When is this updated, and what's the logic (because how I understood it doesn't seem to apply here)?

1.8.0 RC3:
[2017-02-20 20:27:35] <Ami-Laptop> $$ $::numversion
[2017-02-20 20:27:35] <Scorpion> [TCL (1ms)] invalid command name "1080003"

1.8.0 final release [2016-12-04]:
[2017-02-20 20:37:24] <Ami-Laptop> $$ $::numversion
[2017-02-20 20:37:24] <Scorpion> [TCL (1ms)] invalid command name "1080000"

By pure sorting, the FR would come before RC3, which seems wrong.

thommey:
Re-opening because of additional comments/input:

From @maimizuno
> The purpose was: to be able to pull data via a script.
> ".status" only displays to the party line / DCC output, which isn't really scriptable.

> Example: If I want to write a bind for bots based on version or compile flags
> !1800RC4! say I'm an RC4 bot! Upgrade me!
> [bunch of bots reply with the message]

thommey:
Can you check if the new versioning we did for 1.8.1 (rc2 is released) assists you to be able to do this with the global $::version variable? It should contain "rc2" in $::version.

ghost:
'fraid not. :(

[2017-03-03 20:16:20] <Ami-Laptop> $$ $::version
[2017-03-03 20:16:24] <Scorpion> [TCL (1ms)] invalid command name "1.8.1+alpha"
[2017-03-03 20:16:33] <Domino> $$ return $::numversion
[2017-03-03 20:16:34] <Scorpion> [TCL (0ms)] 1080100

Also, if we're on RC2, then, shouldn't the patch level be "01" instead of "00" (unless you didn't release an RC1).

maimizuno:
Also, please see the new comment on #245; What I'm requesting be reviewed isn't being understood. See me on my network / channel (Pixelz knows where I am) for any clarification.

ghost:
Just-grabbed zip from github, freshly compiled just now:

[2017-03-05 04:05:14] <Domino> .tcl set ::version
[2017-03-05 04:05:14] <Dorito> Tcl: 1.8.1+alpha 1080100
[2017-03-05 04:05:23] <Domino> .tcl set ::numversion
[2017-03-05 04:05:23] <Dorito> Tcl: 1080100

Why closed unpatched?

Pixelz:
[...]

PR #338, opened 2017-02-25
Task/setpatch

vanosg:
Found by: maimizuno
Patch by: Geo
Fixes: #337 

One-line summary: Revise eggdrop internal versioning system

Use version.h to set version/patchlevel via misc/setpatch. Each time setpatch is run, the patchlevel digits of egg_numver is incremented by 1, not to exceed 99. misc/newversion was also updated to incorporate this. Under this system, misc/newversion should be run just prior to releasing a Release Candidate. Using 1.8.1 as an example, newversion sets numver to 1080100.  Using setpatch to update for RC (as part of releaseprep), numver will be updated to 1080101. Each new RC will increment by 1, and the stable will become RCX+1 (So, assuming 2 RCs, 1080103). Patches in the develop branch beyond stable will increment the patchlevel digits (last two), until newversion is called again to prepare the next sub-release and the cycle repeats. 


misc/newversion
```diff
@@ -43,16 +42,16 @@ umask 022
 
 OLD_EGGVERSION=$(grep AC_INIT configure.ac | sed -e 's/AC_INIT(\[Eggdrop\],\[//g' -e 's/\],\[bugs@eggheads.org\])//g')
 NEW_EGGVERSION=$(echo $OLD_EGGVERSION | cut -d. -f1-2).$(($(echo $OLD_EGGVERSION | cut -d. -f3) + 1))
-NEW_NUMVERSION=$(($(grep 'egg_numver =' src/main.c | sed -e 's/.*= //' -e 's/;$//') + 100))
+NEW_NUMVERSION=$(($(grep -E '^#define EGG_NUMVER.*' src/version.h |cut -d " " -f 3|sed -e 's/..$//' -e 's/$/00/') + 100))
```
Cizzle: Why not use "sed -e 's/..$/00/'"? And I don't see any ERE's in that regex needing "grep -E"?

misc/setpatch
```diff
@@ -0,0 +1,110 @@
+#! /bin/bash
+#
+# addpatch - modifies src/patch.h
```
Cizzle: Needs update.

misc/setpatch
```diff
@@ -0,0 +1,110 @@
...
+#
+# As a special exception to the GNU General Public License, if you
+# distribute this file as part of a program that contains a
+# configuration script generated by Autoconf, you may include it under
+# the same distribution terms that you use for the rest of that program.
+
+show_usage() {
+	echo "Usage: `basename $0` [-hs] <patchname>"
+	echo ""
+	echo "  patchname  - Name of patch to add."
```
Cizzle, line 30: Needs update too I guess?

misc/setpatch
```diff
@@ -0,0 +1,110 @@
...
+else
+	patch_name=$1
+fi
+
+if test ! -f src/main.c; then
+	echo "You are not in the Eggdrop root directory."
+	exit 1
+fi
+
+oldpatch_name=$(cat src/version.h |grep -E '^#define EGG_PATCH.*'|cut -d "\"" -f 2)
```
Cizzle: Save a process, kill the cat.

misc/setpatch
vanosg: Updated

misc/setpatch
vanosg: This is accurate for usage

misc/setpatch
vanosg: Updated

misc/setpatch
```diff
@@ -0,0 +1,111 @@
+#! /bin/bash
+#
+# addpatch - generates src/version.h
```
Cizzle, line 3: "setpatch"

Cizzle:
src/Makefile.in still has patch.h ref.

PR #356, opened 2017-03-18
(WIP) New password hashing algorithm

thommey:
Just for tracking, not usable yet.


michaelortmann:
I updated this patch to current git, see:

https://github.com/michaelortmann/eggdrop/tree/pass2

- Merge remote-tracking branch 'upstream/feature/pass2' into pass2 (based on current develop 08-Sep-2019 2caaa82b7c87b72a40e4840767b870eaa47744af)
- Remove mbedtls (we talked about this in IRC)
- Cleanup Makefile and strings like 1.9.0 and 2019
- misc/makedepend
- Update ax_check_openssl.m4
- Do not load mod if necessary openssl stuff not found
- Remove gap in modvals.h
- Honor pbkdf2 module when main.c checks for an encryption hook
- Add "#loadmodule pbkdf2" to eggdrop.conf, **we will probably also want config options like cycles**
- Use b64_ntop() and b64_pton() instead of openssl glue

Current state seems to be, that hooks for pbkdf2 are set, but not called, some piece here is missing, probably design work.

thommey:
Thanks for updating this. Removing mbedtls is fine. The transition plan was the following:

- Everytime a user enters/changes/sets their password, store the PASS2 PBKDF2 hash in their useraccount (leave PASS around though), and prefer PASS2 over PASS if it exists
- In a future version, stop setting PASS and remove it once PASS2 is known
- Then, in some future version, possibly remove PASS support entirely

Open for other suggestions of course :)

And yes, I faintly remember that a bit of the hook stuff was missing to achieve it

vanosg:
Closing, this has been used and updated by #903.

Issue #324, opened 2017-01-28
ASN1_STRING deprecated in OpenSSL 1.1


vanosg:
tls.c:285:3: warning: 'ASN1_STRING_data' is deprecated [-Wdeprecated-declarations]
tls.c:359:9: warning: 'ASN1_STRING_data' is deprecated [-Wdeprecated-declarations]

PR #372, opened 2017-04-18
Use new ASN1_STRING_get0_data fucntion when available.

Cizzle:
Found by: @vanosg
Patch by: Cizzle
Fixes: #324 

One-line summary: Fixes the warnings about the deprecated ASN1_STRING_data() function when using OpenSSL 1.1.0+

Adding a (configure-generated) define in config.h (which is included via main.h) generates a "conflicting type" error when compiling and is thus not usable.

Might be preferable to do the define the other way around and use the newer name in the code for future readings?

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -31,6 +31,9 @@
 #include <openssl/err.h>
 #include <openssl/rand.h>
 #include <openssl/x509v3.h>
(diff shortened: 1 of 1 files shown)


thommey:
Can you elaborate on the conflicting type? I would prefer feature tests over version checks.

thommey:
superseded by #373

Issue #366, opened 2017-04-05
Rehashing with ? in alt-nick changes nick


thommey:
Rehashing seems to generate a new preferred nickname instead of checking if the current nickname matches the existing alt-nick pattern (with ? being a digit). This leads to unnecessary nick changes.

vanosg:
thommey, can you show a use case of this happening, so I can make sure I'm following your train of thought correctly?

thommey:
```tcl
set nick NickServ; # something in use
set altnick Eggdrop?????
set keep-nick 1
```
let the bot connect and grab its altnick (Eggdrop19538), then .rehash the bot. It re-generates the alt-nick and sets it to Eggdrop38464 and immediately changes to it, instead of recognizing that its current nick matches the altnick pattern (? being 0-9). This should check if the current nick matches the pattern instead of blindly setting altnick to a constant new random nick.

PR #389, opened 2017-05-03
Check if current nick matches altnick mask before switching, after a rehash

vanosg:
Found by: thommey
Patch by: Geo
Fixes: #366 

One-line summary:


Cizzle:
LGTM

PR #390, opened 2017-05-05
Fix bug where the foreground user lost its owner perms upon rehash/reload

Cizzle:
Found by: @vanosg 
Patch by: Cizzle
Fixes:

### DIFF
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -41,11 +41,12 @@
 
 #include "modules.h"
 
+extern struct dcc_t *dcc;
 extern struct userrec *userlist;
 extern log_t *logs;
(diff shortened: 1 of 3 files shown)


vanosg:
Fixes the bug, thanks!

What about 
```
[14:41:00] #-HQ# restart
You cannot .restart a bot when running -n (due to Tcl).
```

Has that always been there, or is that new? Is there a way around it?

Cizzle:
Hasn't been changed by me and haven't looked into it either.

vanosg:
LGTM :+1: 

Just wanted to make sure this didn't introduce that message. Thanks!

Issue #381, opened 2017-04-26
.help chaninfo


ghost:
.help chaninfo doesn't reflect that you can set flood-* to "0" to deactivate.

vanosg:
I'm on the fence if this is a 'shortcut' or a 'bug' ... it could be fixed one way or the other, but are there second order effects from using '0' as the value?

Cizzle:
As it's listed as a valid option in eggdrop.conf and even in the chaninfo help itself a few lines below, I'd mark this as shortcut. Looking at the code, a "0" for either "count" or "time" disables it. Side effect of using just "0" is that the "time" is set to "1", which should actually be "0"?

PR #392, opened 2017-05-08
Update flood-* docs to reflect actual possibilities.

Cizzle:
Found by: maimizuno
Patch by: Cizzle
Fixes: #381 

Updated docs listing X and X:X as valid options and what happens when X is 0 and when only X is used.


vanosg:
Here's how my little brain reads this:
```
Setting this to X sets the seconds to 1.
```
```
.chanset #geo flood-kick 4:X
Successfully set modes { flood-deop { 4:X } } on #geo.
```
```
.chaninfo #geo
<snip>
flood settings: chan ctcp join kick deop nick
number:           10    3    0    3    4    0
time  :           20   30    0    0    0    0
```
Seconds is 0 - so either that's not what you meant, or it's not set to 1?

Could we add a check in code to that, so that only integers are accepted for the time value?

Cizzle:
Eh, is this when reading my very brief summary of this PR, or when reading the actual docs?
```
Setting this to 0, 0:X or X:0 disables kick flood protection
for the channel. Setting this to X sets the seconds to 1. The "X" is
an integer >= 0.
```
The last sentence is important.

EDIT: Is your example incosistent or is this a bug? You .chanset flood-kick, but it replies with flood-deop?

vanosg:
That copy/paste is from the docs. From your answer, if you don't mind, I'll clarify the language a bit. That aside- we probably should test that value to be an integer.

re: your edit, that's a copy/paste error.

vanosg:
What are your thoughts on this?

```
.chanset #bot flood-deop 4:3
Successfully set modes { flood-deop { 4:3 } } on #bot.

.chanset #bot flood-deop 0:0
Successfully set modes { flood-deop { 0:0 } } on #bot.

.chanset #bot flood-deop 0:4
Successfully set modes { flood-deop { 0:4 } } on #bot.

.chanset #bot flood-deop 4:0
Successfully set modes { flood-deop { 4:0 } } on #bot.

.chanset #bot flood-deop 4:D
Error trying to set flood-deop for #bot, invalid option

.chanset #bot flood-deop D:4
Error trying to set flood-deop for #bot, invalid option

.chanset #bot flood-deop 3
Error trying to set flood-deop for #bot, invalid option

.chanset #bot flood-deop 0
Successfully set modes { flood-deop { 0 } } on #bot.

.chaninfo #bot
<snip>
flood settings: chan ctcp join kick deop nick
number:           15    3    5    3    0    5
time  :           60   60   60   10    0   60
```

src/mod/channels.mod/help/chaninfo.help
```diff
@@ -162,27 +162,21 @@
                        having ops.
         %bstatic%b         Allow only permanent owners to remove the channel?
      The following are flood settings, they are set via ".chanset <channel>
-     flood-type number:seconds", where number and second are integers
-     indicating the number of times in how many seconds the flood will be
-     triggered. Setting one of these to 0:0, 0:1, or 1:0 will deactivate the
-     respective flood setting.
+     <flood-type> <number>[:<seconds>]", where <number> and <seconds> are
```
Cizzle: The `[:<seconds>]` here might be misleading as only 0 is valid without any seconds, which is not clear from the rest of the doc.

src/mod/channels.mod/tclchan.c
```diff
@@ -1247,6 +1247,7 @@ static int tcl_channel_modify(Tcl_Interp *irp, struct chanset_t *chan,
       old_mode_pls_prot = chan->mode_pls_prot;
   struct udef_struct *ul = udef;
   char s[121];
+  char *endptr1, *endptr2;
```
Cizzle: I concur with @thommey: only 1 `*endptr` needed.

src/mod/channels.mod/tclchan.c
```diff
@@ -1512,12 +1513,26 @@ static int tcl_channel_modify(Tcl_Interp *irp, struct chanset_t *chan,
       p = strchr(item[i], ':');
       if (p) {
         *p++ = 0;
-        *pthr = atoi(item[i]);
-        *ptime = atoi(p);
-        *--p = ':';
+        if ((!strtol(item[i], &endptr1, 10) && (strcmp(endptr1, "\0") != 0)) \
```
Cizzle: The `\` at the end is unnecessary.
Could ~~directly assign values to the `strtol` invocation and~~ get rid of the `strcmp`: ~~`if ((!(*pthr = (int) strtol(...)) && (*endptr)) || (!(*ptime = (int) strtol(...)) && (*endptr))) {`~~ `... && (*endptr)`.
~~The else-clause would be redundant then as long as you move the `*--p = ':';` out of it.~~
Edit: NVM the assignments, we don't want to update the setting when invalid values are given.

src/mod/channels.mod/tclchan.c
```diff
@@ -1512,12 +1513,26 @@ static int tcl_channel_modify(Tcl_Interp *irp, struct chanset_t *chan,
...
+          return TCL_ERROR;
+        } else {
+          *pthr = atoi(item[i]);
+          *ptime = atoi(p);
+          *--p = ':';
+        }
       } else {
-        *pthr = atoi(item[i]);
-        *ptime = 1;
+        if (!strtol(item[i], &endptr1, 10) && (endptr1[0] != "\0")) {
```
Cizzle: Comparing a `char` to a `char[]`? As above (but the inverse, `endptr` must point to the end of the string indicating a valid 0), use `!(*endptr)`.

src/mod/channels.mod/help/chaninfo.help
vanosg: Maybe 
```<flood-type> <number>:<seconds>, where.... . Omitting seconds assumes '0'```

Something like that?

Comment (Cizzle) [reply src/mod/channels.mod/help/chaninfo.help]: I'd leave the `[:<seconds>]` but would clarify further that only 0 can be used on it's own?

File: src/mod/channels.mod/tclchan.c [hunk 5]
```diff
@@ -1512,12 +1513,26 @@ static int tcl_channel_modify(Tcl_Interp *irp, struct chanset_t *chan,
       p = strchr(item[i], ':');
       if (p) {
         *p++ = 0;
-        *pthr = atoi(item[i]);
-        *ptime = atoi(p);
-        *--p = ':';
+        if ((!strtol(item[i], &endptr1, 10) && (strcmp(endptr1, "\0") != 0)) \
+           || (!strtol(p, &endptr2, 10) && (strcmp(endptr2, "\0") != 0))) {
```
Cizzle: From the docs I gather `X:`and `:Y` aren't valid, we'd want a check for `!p || !item[i]` then as well?

PR #394, opened 2017-05-10
Add configure arguments to Tcl API. Fix string expansion of it for -v.

### LINKED ISSUE #342: Feature: compile information available via TCL code
From the shell, "eggdrop -v" provides compile-time options. Would be helpful to get that via script (without relying on "exec eggdrop -v").

### LINKED ISSUE #385: Getuser response missing some params
Getuser missing "fstat" and other data.

thommey:
Add configure arguments to Tcl API. Fix string expansion of it for -v.

Fixes #342.

Found by: maimizuno


thommey:
Expansion fix was necessary because otherwise,
`./configure --enable-foo='a'"b"\\c --'enable-bar=a b c'`
caused Eggdrop to not compile (because the AC_DEFINE_UNQUOTED is not a valid C string literal)

doc/sphinx_source/mainDocs/tcl-commands.rst
```diff
@@ -2604,6 +2604,13 @@ config
 
   Module: core
 
+^^^^^^
+config
```
Cizzle: `configure-args`?

Issue #391, opened 2017-05-08
'./eggdrop -c' should not work without -n


vanosg:
Per help, 
```
-nc  Don't background; display channel stats every 10 seconds.
```
There is no '-c' option listed as a valid option in the help. The -n flag keeps eggdrop in the foreground. Running './eggdrop -c' without a -n appears to start the bot into background normally, without the -c causing any different functionality. Using -c by itself should probably error and not allow eggdrop to run. Thoughts?

vanosg:
Same goes for the -t option by itself
```
-nt  Don't background; use terminal to simulate DCC chat.
```
-t launches without any noticeable change or error.

Cizzle:
How about we change the help to show:
```
Usage: eggdrop [options] [config-file]

Options:
-n Don't background; send all log entries to console.
-c  Don't background; display channel stats every 10 seconds.
-t  Don't background; use terminal to simulate DCC chat.
-m   Create userfile.
-h   Show this help.
-v   Show version info, then quit.
The -n, -c and -t cannot be used together.
```
In the code we remain backward compatible of course, only a few minor changes needed (`backgrd = 0` at -c and -t and bitflags of 24 and 20).

vanosg:
I really like that suggestion- I can't think of any reason the n flag is still needed, except as a carryover from some long-ago design decision we could get around, especially with the new code (as you identified). 

As a philosophical point, I think 1.9.0 might be a better opportunity to formally implement a command line change in the documentation while maintaining backwards compatibility (and an associated deprecation notice), and 2.0 to remove the 'old' way altogether, similar to our discussion on the CBC encryption.

Cizzle:
Having checked the code, providing just -t or -c does not do anything without the -n, so it's just as if no flag was supplied.
I'm OK with the roadmap, the code in 2.0 will need some extra checks though to have -t/c/n be mutually exclusive.

vanosg:
TODO: Update return value to 1 (or other value) when these errors occur (currently 0 is returned).

PR #399, opened 2017-05-13
Do not start if -n is missing from -c or -t cli flags

vanosg:
Found by: Geo
Patch by: Geo
Fixes: #391 

One-line summary:
Do not start if -n is missing from -c or -t cli flags

Check for -c or -t, if -n isn't there, exit and return 1.

```
$ ./eggdrop -t

Eggdrop v1.8.1+altnick (C) 1997 Robey Pointer (C) 1999-2017 Eggheads
ERROR: The -n flag is required when using the -c or -t flags. Exiting...
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -589,12 +589,10 @@ static void do_arg()
   } else if (cliflags & 1) {
     show_ver();
     exit(0);
-  } else if ((cliflags & 8) || (cliflags & 4)) {
-    if (!(cliflags & 16)) {
-      printf("\n%s\n", version);
-      printf("ERROR: The -n flag is required when using the -c or -t flags. Exiting...\n\n");
-      exit(1);
-    }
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -589,6 +589,12 @@ static void do_arg()
   } else if (cliflags & 1) {
     show_ver();
     exit(0);
+  } else if ((cliflags & 8) || (cliflags & 4)) {
+    if (!(cliflags & 16)) {
```
Cizzle: Combine these to `} else if (!(cliflags & 16) && (cliflags & 8 || cliflags & 4)) {`, otherwise the WARNING of multiple confs won't be shown when using `./eggdrop -nt bot.conf etc.conf` (and `-nc`).

Cizzle:
LGTM

Issue #406, opened 2017-05-22
Wrong error for chanset flood setting


vanosg:
.chanset #channel flood-deop :4 should return "flood value must be in X:Y format" not "values must be integers"  ... or is my OCD raging too much?

PR #407, opened 2017-05-22
Fix some more invalid options and use correct response.

Cizzle:
Found by: @vanosg 
Patch by: Cizzle
Fixes: #406 

One-line summary: Fixes incorrect output like "4a:5b", "3:4:5" etc being allowed. Also uses the correct error message.


vanosg:
```
.chanset #geo flood-deop 43:45:56
Error trying to set flood-deop for #geo, values must be integers: 43:45:56
```
Values are integers, just too many of them. This bug brought to you by Jenkins!

Cizzle:
Technically speaking the error message is correct, as `43:45`(or `45:56`) is not an integer, but just to please you...

Issue #428, opened 2017-07-20
Adjust default console flags for HQ (-nt)


thommey:
console should be at least with +o because that's where all our errors go

PR #429, opened 2017-07-21
Always have +o console flag for new -nt runs.

Cizzle:
Found by: @thommey 
Patch by: Cizzle
Fixes: #428 

One-line summary: Appends +o to the console flags of the -HQ user when starting eggdrop with -nt.

Additional description: This mode can simply be removed during the run.


vanosg:
LGTM - added to tests.

Issue #433, opened 2017-07-27
IRC server connections cannot be made to both IPv4 and IPv6


nomis:
With a single server hostname that has one A and one AAAA resource record, eggdrop 1.8.2rc2 will only ever attempt to connect to one type of address. If this times out (no response to SYN packets) then it does not retry over the other address family.

The DNS resolution code only prefers IPv4 or IPv6. It also looks like it behaves differently depending on the order of the A and AAAA responses.

vanosg:
Thanks so much for reporting this, and the associated PRs! We'll look at this shortly

vanosg:
Is there currently a server with DNS entries for both records where you know that one is timing out on, so I can replicate this error for additional testing? You can stop in on freenode and let us know there if you prefer. Thank you

nomis:
You can use `irctest4.uuid.uk` and `irctest6.uuid.uk`. These will connect to a freenode server (that is currently working) and Google DNS (which doesn't respond to TCP port 6667 for me).

vanosg:
prefer-ipv6 0:
```
[01:37:11] Trying server [irctest4.uuid.uk]:6667
[01:37:11] DNS Resolver: Creating new record
[01:37:11] DNS Resolver: Sent domain lookup request for "irctest4.uuid.uk".
[01:37:11] DNS Resolver: Received nameserver reply. (qd:1 an:1 ns:4 ar:3)
[01:37:11] DNS Resolver: answered domain query: "irctest4.uuid.uk"
[01:37:11] DNS Resolver: TTL: 5s
[01:37:11] DNS Resolver: TYPE: AAAA: IP6 Address
[01:37:11] DNS Resolver: answered domain query: "uuid.uk"
[01:37:11] DNS Resolver: answered domain query: "uuid.uk"
[01:37:11] DNS Resolver: answered domain query: "uuid.uk"
[01:37:11] DNS Resolver: answered domain query: "uuid.uk"
[01:37:11] DNS Resolver: answered domain query: "d.ns.uuid.uk"
[01:37:11] DNS Resolver: answered domain query: "c.ns.uuid.uk"
[01:37:11] DNS Resolver: answered domain query: "p.ns.uuid.uk"
[01:37:11] DNS Resolver: Received nameserver reply. (qd:1 an:1 ns:4 ar:4)
[01:37:11] DNS Resolver: answered domain query: "irctest4.uuid.uk"
[01:37:11] DNS Resolver: TTL: 5s
[...]

PR #434, opened 2017-07-27
Make T_A handler mirror the T_AAAA behaviour. Partial fix for #433.

nomis:
It doesn't make sense to set `rp->sockname.family` to `AF_INET` and then immediately check if it equals `AF_INET6`. The condition will always be false.

This changes the behaviour of T_A to mirror the behaviour of T_AAAA.


Cizzle:
LGTM

thommey:
LGTM

PR #439, opened 2017-08-03
Sanity check for dcc ban syntax. Fixes #358

### LINKED ISSUE #358: .+ban syntax sanity check
channel after ban is super confusing and unexpected, contrary to most other commands on irc and eggdrop (except maybe /invite). some sanity checks would be nice (ban starting with # etc.)

thommey:


Cizzle:
LGTM, though I think a warning should be shown as well so the user learns the correct syntax?

vanosg:
Agree - I think I'd prefer to see an error message presented, vice just fixing it automatically for the user- Maybe a middleground is something like this?

```
.+ban #chan *!*@dsf.com
Usage: +ban <hostmask> [channel] [%<XdXhXm>] [reason]
Did you mean .+ban *!*@dsf.com #chan ?
```

Cizzle:
LGTM, except that space before the '?'.

thommey:
@vanosg I thought about it some more and you are right. The automatic fixing might be something I personally want because I can never remember the syntax and hate that it's the opposite way from almost all other and IRC commands, but automatic fixing is not the solution for everyone.
Simply warning and applying the ban despite the sanity check, or blocking it, seem like a sane thing to do.
Blocking it for now, unless someone objects then :)

vanosg:
The space before the ? is intentional, to aid in the ease of copy/pasting it. Any concern with that?

thommey:
Can be merged, but keep the branch for now.

Waiting for RC for complaints about legitimate bans starting with CHANPREFIX characters (unlikely but possible).

PR #440, opened 2017-08-03
Reset console draft. Fixes #311.

### LINKED ISSUE #311: Add ability to reset console to defaults
If you turn on a lot of console modes via the .console dcc command, for instance all of them via .console +*, it is later annoying having to reset the console to the default state by individually turning off each console flag.

I propose a new subcommand be added, for instance ".console reset" which will reset the console to the defaults set in eggdrop.conf

thommey:
Found by: Pixelz
Fixes:  #311

### DIFF
--- a/src/cmds.c
+++ b/src/cmds.c
@@ -641,6 +641,7 @@ static void cmd_boot(struct userrec *u, int idx, char *par)
     dprintf(idx, "Who?  No such person on the party line.\n");
 }
 
+/* Make changes to user console settings */
 static void do_console(struct userrec *u, int idx, char *par, int reset)
 {
   char *nick, s[2], s1[512];
@@ -651,16 +652,18 @@ static void do_console(struct userrec *u, int idx, char *par, int reset)
   get_user_flagrec(u, &fr, dcc[idx].u.chat->con_chan);
   strncpyz(s1, par, sizeof s1);
   nick = newsplit(&par);
-  /* Don't remove '+' as someone couldn't have '+' in CHANMETA cause
+  /* Check if the parameter is a handle.
+   * Don't remove '+' as someone couldn't have '+' in CHANMETA cause
    * he doesn't use IRCnet ++rtc.
    */
   if (nick[0] && !strchr(CHANMETA "+-*", nick[0]) && glob_master(fr)) {
-    for (i = 0; i < dcc_total; i++)
+    for (i = 0; i < dcc_total; i++) {
       if (!egg_strcasecmp(nick, dcc[i].nick) &&
           (dcc[i].type == &DCC_CHAT) && (!ok)) {
         ok = 1;
         dest = i;
       }
+    }
     if (!ok) {
       dprintf(idx, "No such user on the party line!\n");
       return;
@@ -670,8 +673,10 @@ static void do_console(struct userrec *u, int idx, char *par, int reset)
     dest = idx;
   if (!nick[0])
     nick = newsplit(&par);
-  /* Consider modeless channels, starting with '+' */
-  if (nick[0] && ((nick[0] == '+' && findchan_by_dname(nick)) ||
+  /* Check if the parameter is a channel.
+   * Consider modeless channels, starting with '+'
+   */
+  if (nick[0] && !reset && ((nick[0] == '+' && findchan_by_dname(nick)) ||
       (nick[0] != '+' && strchr(CHANMETA "*", nick[0])))) {
     if (strcmp(nick, "*") && !findchan_by_dname(nick)) {
       dprintf(idx, "Invalid console channel: %s.\n", nick);
@@ -683,7 +688,8 @@ static void do_console(struct userrec *u, int idx, char *par, int reset)
               nick);
       return;
     }
-    strncpyz(dcc[dest].u.chat->con_chan, nick, 81);
+    strncpyz(dcc[dest].u.chat->con_chan, nick,
+        sizeof dcc[dest].u.chat->con_chan);
     nick[0] = 0;
     if (dest != idx)
       get_user_flagrec(dcc[dest].user, &fr, dcc[dest].u.chat->con_chan);
@@ -749,6 +755,7 @@ static void cmd_console(struct userrec *u, int idx, char *par)
   do_console(u, idx, par, 0);
 }
 
+/* Reset console flags to config defaults */
 static void cmd_resetconsole(struct userrec *u, int idx, char *par)
 {
   do_console(u, idx, par, 1);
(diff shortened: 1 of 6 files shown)


Cizzle:
LGTM

vanosg:
```
.console
Your console is #chan: mkcobrxsdvth (msgs, kicks/modes, cmds, misc, bots, raw, files, server, debug, server output, botnet traffic, share traffic).

.resetconsole
Invalid console channel: .
```

thommey:
That should be fixed now. I also set the default console in eggdrop-basic.conf and in C (it used to be empty).

vanosg:
Should ".resetconsole" by itself only reset the current console channel? It appears to reset all console channels to the default. Perhaps we want to limit the command, without a "*" specified as target console, to only reset the current console channel? (The * target, iirc, is a bit of an undocumented trick to apply the console changes to all channels)

```
.console +h
Set your console to #foo: mkcobxsh (msgs, kicks/modes, cmds, misc, bots, files, server, share traffic)
.console #bar +r
Set your console to #bar: mkcobrxsh (msgs, kicks/modes, cmds, misc, bots, raw, files, server, share traffic).

.console
Your console is #bar: mkcobrxsh (msgs, kicks/modes, cmds, misc, bots, raw, files, server, share traffic).
.console #foo
Set your console to #foo: mkcobrxsh (msgs, kicks/modes, cmds, misc, bots, raw, files, server, share traffic).
.resetconsole
Set your console to #foo: mkcobxs (msgs, kicks/modes, cmds, misc, bots, files, server).
.console #bar
Set your console to #bar: mkcobxs (msgs, kicks/modes, cmds, misc, bots, files, server).
```
^^^ I would expect #bar to retain the 'h' console flag in this scenario, but it was reset. We may want to further clarify the partyline help file as well in either case.

vanosg:
... and now we'll get even further into opinion territory :) With all the discussion we had over eggdrop-basic back in the day, and trying to keep it slim and trim... is this really a setting we want to move back in there? Are most people going to change this setting? Eggdrop will fully function with minimal need to adjust that setting, which I think was one of the criteria for moving stuff in- which settings needed to be changed by the user. 

I have also decided I'm going to start adding random emojis to all my comments. Enjoy :icecream: :haircut_man: :taco:

thommey:
Fixed `<thommey> .console thommey <lamestbot> Invalid console channel: .` in this PR now.

.console -o and .console * -o remove 'o' from all console channels
.console #chan -o removes it only from that channel and changes to it

That makes me think .resetconsole without a channel should also reset them all, and .console doesn't have any option to change "current channel only" other than explicitely specifying it.
Comments?

thommey:
Actually, console flags are global, as in your example. .console #bar +m adds it to all channels, there are no separate entries so .resetconsole has to reset all

vanosg:
Yeah... I'm dumb. Sorry! :duck: :man_dancing:

Cizzle:
Remains the question whether `.resetconsole` without args should reset the channel to it's default (* if no channels are there, or the `force-channel` number in the channel list) too. Right now it doesn't touch the channel.

vanosg:
As I'm writing tests for this... is [channel] even used? As I silly-ly showed above, console flags aren't specific to a channel, so what does specifying a channel actually do?

Pixelz:
Think of channel more like a setting like the rest. You can only specify one channel at a time. For example .console #eggdrop +p would let you see public chat from #eggdrop. If you then did .cosole #eggheads it would show public chat from #eggheads.

Apart from that, whatever channel you specify in console is also used by some other commands, such as .bans (which by the way I think is highly confusing)

thommey:
That parameter mostly exists to .resetconsole because .console has it too and it shares the code. In resetconsole (and console) it sets the current channel to the one specified (for the specified user or yourself).

vanosg:
@Pixelz, thats what I thought too (see my earlier comment in this thread about channel records) but it turns out it doesn't work that way:

```
.console
Your console is #testegg: mkcobxs (msgs, kicks/modes, cmds, misc, bots, files, server).
.+chan #testegg2
[15:22:12] #-HQ# +chan #testegg2
.console
Your console is #testegg: mkcobxs (msgs, kicks/modes, cmds, misc, bots, files, server).
.console #testegg2 +p
[15:22:27] #-HQ# console #testegg2 +p
Set your console to #testegg2: mpkcobxs (msgs, public, kicks/modes, cmds, misc, bots, files, server).
.console #testegg
[15:22:32] #-HQ# console #testegg
Set your console to #testegg: mpkcobxs (msgs, public, kicks/modes, cmds, misc, bots, files, server).
```
See there - I set a console setting for #testegg2, it changes my console to that channel and adds the flag. If I go back to #testegg, the flag I added for #testegg2 remains.

Pixelz:
You just added the public flag, you didn't add it to any particular channel. You can only have a single console channel set at any particular time, and changing the console channel doesn't change any of the flags.

`.console #foo +p`
Eggdrop will show public chat from #foo

`.console #bar`
Eggdrop will show public chat from #bar, and no longer from #foo

`.console *`
Eggdrop will show public chat from all channels

vanosg:
That's exactly what I did above, it doesn't work that way.

vanosg:
LGTM. Tests added. @Cizzle, can you review again?

src/cmds.c
```diff
@@ -715,10 +714,12 @@ static void cmd_console(struct userrec *u, int idx, char *par)
           dcc[dest].u.chat->con_flags &= ~logmodes(s);
       }
     }
+  } else if (reset) {
+    dcc[dest].u.chat->con_flags = (u->flags & USER_MASTER) ? conmask : 0;
```
Cizzle, line 718: Why reset to 0 (no flags) when not a master?
Same at tcldcc.c.
Don't we also want `strncpyz(dcc[dest].u.chat->con_chan, chanset ? chanset->dname : "*", sizeof dcc[dest].u.chat->con_chan);` when resetting?

thommey: That seems to be the overall intention of the code I gathered, even ignoring the bit that was already there in .console:
https://github.com/eggheads/eggdrop/blob/0d6d8d/src/cmds.c#L1388-L1389
https://github.com/eggheads/eggdrop/blob/0d6d8d/src/cmds.c#L2238-L2239
https://github.com/eggheads/eggdrop/blob/0d6d8d/src/dcc.c#L685
https://github.com/eggheads/eggdrop/blob/0d6d8d/src/userrec.c#L918-L936

thommey: I guess unifying that into using check_conflags() is in order, but the result stays pretty much the same I think

Cizzle: As a new non-master user joining partyline has no flags as well, this seems to be OK after all.

thommey:
NOTE: Merge this after all the other flag changing patches

Issue #8, opened 2010-02-03
No console mode to disable seeing links/unlinks.


thommey:
| Trac | Data |
| --: | :-- |
| Ticket | 8 |
| Reported by | MC_8 |
| Status | new |
| Component | Core |
| Priority | minor |
| Milestone | 1.8.0 |
| Keywords | log loglevel |
| Version | 1.8.0 CVS |

[10:20:42] <MC_8> .console -mcobs
[10:20:42] <chipset> Set your console to #abc123: - (none)
[10:20:45] <chipset> **\* (Limbo) Disconnected from: warez. No reason (lost 1 
bot and 1 user)
[10:20:46] <chipset> **\* (Limbo) Linked to warez.

Version:  1.6.7.3  --  v1.6.7+argv  (CVS)

On a 200+ botnet, I'm getting tired of seeing all these links and unlinks.  
Tried to disable the console to show it, and found that I cannot.  I'm stuck 
with the extra scrole.
Comment # 1 From BarkerJr 2002-10-02 12:02:22

**\* Bug 268 has been marked as a duplicate of this bug. ***

Comment # 2 From BarkerJr 2002-10-02 13:54:24

It seems that older bots send 'ct' commands along with unlinks.  But, I also
noticed that the bot displays unlinks from newer bots, too.  If that makes any
sense.

In other words, as long as older bots are on the net, we may never be able to
stop the link msgs.  On the other hand, we may be able to reduce them.

Comment # 3 From BarkerJr 2002-10-02 15:08:31

Created an attachment (id=30) [details]
Patch to fix console +b

I think this patch may fix it.  It seems that the text was being sent to all
dcc chats instead of using putlog, cause the msgs didn't even show up in a log,
which seems odd to me.

This will not hide links/unlinks from bots directly linked to the bot you're
on.  Only remote bots.  I figure local bot msgs are important enough to be put
through -b.

Comment # 4 From Psychopath 2002-11-18 10:57:39

**\* Bug 284 has been marked as a duplicate of this bug. ***

Comment # 5 From Will Buckner 2003-01-17 22:28:00

Maybe a better option would be a flag to not show broadcasts.

Comment # 6 From Carl M. Gregory 2003-07-25 09:28:31

BarkerJr:  This will not hide links/unlinks from bots directly linked to the 
bot you're on.  Only remote bots.  I figure local bot msgs are important enough 
to be put through -b.

Who's to say what is important though?  I think it should be pipped through the 
[...]

PR #441, opened 2017-08-04
Move certain global partyline output to existing and new console log modes

Cizzle:
Found by: MC_8
Patch by: Cizzle
Fixes: #8 

One-line summary: Output mainly dealing with botlinking is now put under the 'b' console log mode and output dealing with messages coming from linked bots are now put under the new 'l' console mode.

Additional description: Many of these messages were only meant for (bot)masters but could be seen by anyone.

1. `.console * +l` on a leafbot B
2. `.unlink C` on the hubbot
3. leafbot B will show a message that C was unlinked
4. `.link C` on the hubbot
5. leafbot B will show a message that C is linked
6. `.console * -l` on a leafbot B
7. `.unlink C` on the hubbot
8. leafbot B will not show any message


doc/sphinx_source/coreDocs/core.rst
```diff
@@ -225,7 +227,7 @@ logfile <logflags> <channel> "logs/logfile"
 Console Settings
 ----------------
 
-  set console "mkcobxs"
+  set console "mkcobixs"
```
vanosg: 'l' not 'i'

src/botcmd.c
```diff
@@ -697,7 +696,7 @@ static void bot_nlinked(int idx, char *par)
     dprintf(idx, "error %s\n", BOT_YOUREALEAF);
   }
   if (s[0]) {
-    chatout("*** %s\n", s);
+    putlog(LOG_BOTS, "*", "%s.", s);
```
vanosg, line 699: For standardization purposes, can we leave the proceeding '***'s of the message in place here (and elsewhere)? It would be good to retain as much of the previous appearance/functionality as possible.

vanosg:
Should this be considered backward compatible? 
1) Does this mess with messages triggering the BCST bind (see in-code question regarding ***s, for instance)
2) Less likely... BOT/JOIN/DISC binds?
3) Is there any other situation (I can't think of one, but again, I'm not 'smart'...) where a user is counting on some information being on the b console flag that is now moved to l? 

If the answer is yes to any of those, I think this might be something to save for 1.9. I have ideas to discuss if the answer is 'yes' but you really want it in here, so don't hold back... thanks!

src/botcmd.c
Cizzle: We can, but I don't see any standard in keeping '***' in logmessages, especially with the time prefixed to it now.

Cizzle:
1. No. Only `chatout` calls are removed which only prints text on the partyline of connected users.
2. Still idem ditto.
3. Actually no info that was on the 'b' mode was moved to the 'l' mode. Three previous 'b'-mode messages are slightly altered (all still pertaining their info and even more added).

Cizzle:
Note: when merging PR #440 into develop, file `src/chanprog.c` at line 454 should have `LOG_BOTMSG` appended to `conmask`.

vanosg:
1) ok
2) I give up, what is idem ditto?
3) by 'pertaining their info,' you mean no info was removed, only info was added?

And I'm still a fan of leaving the ***s in the messages...

Cizzle:
Latin for "same", here used as "same as number 1". Yes, no info is removed, only added and might be in a different wording.

vanosg:
I'm still a fan of leaving messages (***) add they were, but if your have a reason for the change... LGTM :+1:

thommey:
I would prefer a unified prefix for botnet traffic, such as [b<-]/[b->] or maybe show the handle in there, [HubBot<-] [HubBot->] or anything similar to that really.

Cizzle:
For me the "***" is just more clutter, it adds nothing to the message. As thommey pointed out before on IRC, the "***" is more used for fatal/internal messages. The LOG_BOTMSG already has some sort of prefix as "(SendingBotName)". The LOG_BOTS and LOG_BOTMSG should not be confused with botnet traffic (LOG_BOTNET and LOG_BOTSHARE)!

vanosg:
@Cizzle , can you resolve these new conflicts resulting from merging the other PRs? I think we might have re-used some of the values in eggdrop.h (and the levels needs to match now as well). Thanks

PR #445, opened 2017-08-16
Swap incorrect 'ssl-privatekey/certificate not loaded' error messages

vanosg:
Found by: Geo
Patch by: Geo
Fixes: #444 

One-line summary: Swap incorrect 'ssl-privatekey/certificate not loaded' error messages

Additional description (if needed):

Test cases demonstrating functionality (if applicable):

with #set ssl-certificate "eggdrop.crt"
```
[03:01:07] ERROR: TLS: ssl-certificate not set, ignoring ssl-privatekey.
```

with #set ssl-privatekey "eggdrop.key"
```
[03:00:46] ERROR: TLS: ssl-privatekey not set, ignoring ssl-certificate.
```

PR #449, opened 2017-08-18
Fail early on TLS certificate loading errors.

### LINKED ISSUE #447: OpenSSL: temporary CTX with testing cert/key or invalidate cert after
right now, we load the certificate into the global ssl_ctx, and then load the key.
1) we don't verify the key with SSL_CTX_check_private_key
2) we can't "unset" the certificate if they key is invalid or doesn't exist, so we should use a temporary ssl_ctx to do that

thommey:
Fatal on cert loading errors and report unsupported STARTTLS to LOG_MISC.

PR based on #445 bug/sslmsg

Fixes #447

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -138,14 +138,14 @@ int ssl_init()
     return -1;
   }
   ssl_files_loaded = 0;
-  if (!tls_certfile[0]) {
-    /* Intentional fall-through for !tls_certfile[0] && !tls_keyfile[0] */
-    if (tls_keyfile[0])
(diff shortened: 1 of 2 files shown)


Cizzle:
LGTM

vanosg:
Tests added. 

Do we want to kill eggdrop if privatekey is set, but certificate is commented out? Currently we just warn, but we kill if one of those settings points at a non-existent file... this would feel a little more like a standard behavior to me. If yes, I (or you) can add; otherwise, LGTM :+1:

vanosg:
Per thommey, kill the bastard.

Cizzle:
Still LGTM.

PR #450, opened 2017-08-18
Allow parallel make for building eggdrop

thommey:
Allows "make -j 9" instead of "make" (cannot use -j with config or install) to make use of multiple processors. Only tested on GNU Make so far, needs testing on other make systems.


thommey:
I think we can merge this (but keep the branch for revert), and wait for complaints. I tested it with gnu make and bmake, works fine, and the only change is some streamlining of rules (&&) and 2-3 more sub-make invocations to enforce dependency order with -j

Issue #21, opened 2010-02-03
userfile sharing fails during certain NAT situations


thommey:
| Trac | Data |
| --: | :-- |
| Ticket | 21 |
| Reported by | Darkreap1, simple |
| Status | accepted |
| Component | Module: share.mod |
| Priority | minor |
| Milestone | 1.8.0 |
| Version | 1.8.0 CVS |

Created an attachment (id=94) [details]
userfile sharing patch

Bots should be connecting to the bot address just like they do for regular
linking. Instead they are exchanging IP addresses which can get messy if you
use nat-ip, this problem has existed since I started using 1.3. Basically if
you have a bot on your lan, sharing with other bots on your lan, and also
sharing with bots on the internet, its going to fail unless you have some
firewall rules to work around it.

crappy patch included

PR #456, opened 2017-09-02
Use the currently used address for userfile sharing.

Cizzle:
Found by: Darkreap1, simple
Patch by: Cizzle
Fixes: #21 

One-line summary: A slave will now use the address of the hub (to which it's succesfully connected) instead of an IP given by the hub which was wrong in certain cases (especially in NAT situations).

Additional description: backward compatibility is kept, older bots still get the IP given by the hub.


thommey:
LGTM

PR #460, opened 2017-09-14
Bug/tlsaddrcheck

vanosg:
Found by: Geo
Patch by: Geo
Fixes: NA

One-line summary:
Various error-checks against +bot/chaddr functions

Prevents IPv6 addresses from being added when no IPv6 support
Prevents + ports from being added when no TLS support
Prevents out of range ports
Checks for missing address field

### DIFF
--- a/src/cmds.c
+++ b/src/cmds.c
@@ -745,7 +745,7 @@ static void cmd_console(struct userrec *u, int idx, char *par)
 static void cmd_pls_bot(struct userrec *u, int idx, char *par)
 {
   char *handle, *addr, *port, *relay, *relay2, *host;
-  char saddr[sizeof(struct in_addr)];
+  struct in_addr saddr;
   struct userrec *u1;
   struct bot_addr *bi;
   int i, found = 0;
@@ -1092,7 +1092,7 @@ static void cmd_chaddr(struct userrec *u, int idx, char *par)
 #endif
   int i, found = 0, telnet_port = 3333, relay_port = 3333;
   char *handle, *addr, *port, *relay, *relay2;
-  char saddr[sizeof(struct in_addr)];
+  struct in_addr saddr;
   struct bot_addr *bi;
   struct userrec *u1;
 
@@ -1132,7 +1132,7 @@ static void cmd_chaddr(struct userrec *u, int idx, char *par)
   }
   if (relay) {
     relay2 = relay;
-    // Convert to just the port number string for error checking
+    /* Convert to just the port number string for error checking */
     relay2++;
     if (*relay2 == '+') {
       relay2++;
@@ -1148,7 +1148,7 @@ static void cmd_chaddr(struct userrec *u, int idx, char *par)
     return;
   }
 
-// Check if user forgot address field
+/* Check if user forgot address field */
   for (i=0; i < strlen(addr); i++) {
     if (!isdigit((unsigned char) addr[i]) && (addr[i] != '/')) {
       found=1;
--- a/src/cmds.c
+++ b/src/cmds.c
@@ -742,10 +742,25 @@ static void cmd_console(struct userrec *u, int idx, char *par)
   }
 }
 
+/* Check if a string is a valid integer and lies between
+ * two given integers. Returns 1 if true, 0 if not.
+ */
+char check_int_range(char *value, int min, int max) {
+  char **endptr = NULL;
+  int intvalue;
+
+  if (value) {
+    intvalue = strtol(value, endptr, 10);
+    if ((intvalue < max) && (intvalue > min) && (!endptr && (strcmp(value,"")))) {
+      return 1;
+    }
+  }
+  return 0;
+}
+
 static void cmd_pls_bot(struct userrec *u, int idx, char *par)
 {
-  char *handle, *addr, *port, *relay, *relay2, *host;
-  struct in_addr saddr;
+  char *handle, *addr, *port, *port2, *relay, *relay2=NULL, *host;
   struct userrec *u1;
   struct bot_addr *bi;
   int i, found = 0;
@@ -758,9 +773,10 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
 
   handle = newsplit(&par);
   addr = newsplit(&par);
-  port = newsplit(&par);
+  port2 = newsplit(&par);
+  port = strtok(port2, "/");
+  relay = strtok(NULL, "/");
   host = newsplit(&par);
-  relay = strchr(port, '/');
 
   if (strlen(handle) > HANDLEN)
     handle[HANDLEN] = 0;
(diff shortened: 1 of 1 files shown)


src/cmds.c
```diff
@@ -744,9 +744,11 @@ static void cmd_console(struct userrec *u, int idx, char *par)
 
 static void cmd_pls_bot(struct userrec *u, int idx, char *par)
 {
-  char *handle, *addr, *port, *relay, *host;
+  char *handle, *addr, *port, *relay, *relay2, *host;
+  char saddr[sizeof(struct in_addr)];
```
thommey: this should simply be: struct in6_addr saddr or struct in_addr saddr, depending on #ifdef IPV6

src/cmds.c
```diff
@@ -772,6 +775,61 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
+    return;
+  }
+#endif
+  if ((atoi(port) < 1) || (atoi(port) > 65535)) {
+    dprintf(idx, "Ports must be integers between 1 and 65535.\n");
+    return;
+  }
+  if (relay) {
+    relay2 = relay;
+    // Convert to just the port number string for error checking
```
thommey: C89 doesn't allow // comments, use /* */ instead

src/cmds.c
```diff
@@ -772,6 +775,61 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
+      return;
+    }
+  }
+
+  if (*addr == '+') {
+    dprintf(idx, "Bot address may not start with a +.\n");
+    return;
+  }
+
+// Check if user forgot address field
```
thommey: C99 comment, /* */ instead

src/cmds.c
```diff
@@ -772,6 +775,61 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
+  for (i=0; i < strlen(addr); i++) {
+    if (!isdigit(addr[i]) && (addr[i] != '/')) {
```
thommey: the isdigit function needs to be called with the character value (char) as an unsigned integer, like so:
isdigit((unsigned char)addr[i])
otherwise (on signed char platforms), negative values could be passed which is undefined behavior.
The reason isdigit() takes an int is that character constants ('a') are of type int.

Why is it making sure "addr" is only digits and / instead of the port field? addr would be the IP address, and port is already split into port and relayport and therefore checked.

src/cmds.c
```diff
@@ -1033,8 +1090,9 @@ static void cmd_chaddr(struct userrec *u, int idx, char *par)
 #ifdef TLS
   int use_ssl = 0;
 #endif
-  int telnet_port = 3333, relay_port = 3333;
-  char *handle, *addr, *port, *relay;
+  int i, found = 0, telnet_port = 3333, relay_port = 3333;
```
thommey, line 1125: please apply all comments from above to this one too.
maybe it makes sense to have a function to check for 1<=x<=65535 now that it is used 4 times

thommey (dismissed):

src/cmds.c
vanosg: done

src/cmds.c
vanosg: done

src/cmds.c
vanosg: I just need a random buffer to hold the reponse, I don't ever use the struct format nor reference it... isn't it 'cheaper' to do it this way?

src/cmds.c
vanosg: This checks to see if the address field was skipped by the user- if addr is all numbers (and /), it is not a valid address (neither IPv4, IPv6, or a hostname), and most likely the port field.

src/cmds.c
```diff
@@ -772,6 +775,61 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
     return;
   }
 
+#ifndef IPV6
+  if (!inet_pton(AF_INET, addr, saddr)) {
```
Cizzle: A single `if (strchr(addr, ':'))` is probably less resourcedemanding?

src/cmds.c
```diff
@@ -772,6 +775,61 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
+#endif
+
+#ifndef TLS
+  if ((*port == '+') || (relay && (relay[1] == '+'))) {
+    dprintf(idx, "Ports prefixed with '+' are not enabled "
+      "(this Eggdrop was compiled without TLS support).\n");
+    return;
+  }
+#endif
+  if ((atoi(port) < 1) || (atoi(port) > 65535)) {
```
Cizzle: This still validates `7of9`? Same for relay.

src/cmds.c
```diff
@@ -772,6 +775,61 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
+    dprintf(idx, "Ports must be integers between 1 and 65535.\n");
+    return;
+  }
+  if (relay) {
+    relay2 = relay;
+    /* Convert to just the port number string for error checking */
+    relay2++;
+    if (*relay2 == '+') {
```
Cizzle: Unnecessary, '+' is valid for atoi.

src/cmds.c
```diff
@@ -772,6 +775,61 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
+      "(this Eggdrop was compiled without TLS support).\n");
+    return;
+  }
+#endif
+  if ((atoi(port) < 1) || (atoi(port) > 65535)) {
+    dprintf(idx, "Ports must be integers between 1 and 65535.\n");
+    return;
+  }
+  if (relay) {
+    relay2 = relay;
```
Cizzle: No need for a relay2 if you take the next comment into account and change the 1 to 0 at line 858 and remove the 1 at line 861.

src/cmds.c
```diff
@@ -772,6 +775,61 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
+      return;
+    }
+  }
+
+  if (*addr == '+') {
+    dprintf(idx, "Bot address may not start with a +.\n");
+    return;
+  }
+
+ /* Check if user forgot address field */
```
Cizzle: This comment should be above line 814?

src/cmds.c
```diff
@@ -772,6 +775,61 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
+  for (i=0; i < strlen(addr); i++) {
```
Cizzle: `addr[i]`

src/cmds.c
```diff
@@ -772,6 +775,61 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
+    if (*relay2 == '+') {
+      relay2++;
+    }
+    if ((atoi(relay2) < 1) || (atoi(relay2) > 65535)) {
+      dprintf(idx, "Ports must be integers between 1 and 65535.\n");
+      return;
+    }
+  }
+
+  if (*addr == '+') {
```
Cizzle: Move this to the for-loop below to catch `.+bot Bot 1234/+4321`

src/cmds.c
Cizzle, line 1125: Same for my comments.

src/cmds.c
vanosg: Fixed

src/cmds.c
vanosg: Fixed

src/cmds.c
vanosg: Fixed

src/cmds.c
vanosg: OBE, but fixed.

src/cmds.c
vanosg: Fixed

src/cmds.c
vanosg: Fixed, I think...

src/cmds.c
vanosg: done

src/cmds.c
vanosg: Done

src/cmds.c
vanosg: Done

PR #471, opened 2017-10-30
Bug/tlsaddrcheck

vanosg:
Found by:
Patch by:
Fixes: 

One-line summary:

### DIFF
--- a/src/cmds.c
+++ b/src/cmds.c
@@ -745,7 +745,7 @@ static void cmd_console(struct userrec *u, int idx, char *par)
 static void cmd_pls_bot(struct userrec *u, int idx, char *par)
 {
   char *handle, *addr, *port, *relay, *relay2, *host;
-  char saddr[sizeof(struct in_addr)];
+  struct in_addr saddr;
   struct userrec *u1;
   struct bot_addr *bi;
   int i, found = 0;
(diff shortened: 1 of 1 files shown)


src/cmds.c
```diff
@@ -810,25 +810,25 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
-    }
-    for (i=0; i < addr[i]; i++) {
+ /* Check if user forgot address field by checking if argument is completely
+  * numerical, implying a port was provided as the next argument instead.
+  */
+    for (i=0; addr[i]; i++) {
+      if (addr[i] == '+') {
+        dprintf(idx, "Bot address may not contain a +.\n");
+        return;
+      }
```
Cizzle: If I may suggest as replacement for the above if-block (and it's copy at line 1155):
```
#define PORT_ARG_CHARS "+/"
if (strchr(PORT_ARG_CHARS, addr[i])) {
  dprintf(idx, "Bot address may not contain a %c.\n", addr[i]);
  break;
}
```
And then remove the ` && (addr[i] != '/')` below (and at line 1159).
The `#define` should go elsewhere of course (and named perhaps different too).

Reasoning: more correct errorreport, sooner breaking of the forloop for '/' and using  a break instead of return to have the usage reporting too.

vanosg: That is some sexy code right there.

PR #481, opened 2017-11-11
Control sharing changes to the userfile from passive bots (slaves).

### LINKED ISSUE #1: option to control userfile changes with sharing flags (Bug #59)
| Trac | Data |
| --: | :-- |
| Ticket | 1 |
| Reported by | Ben Dover |
| Status | new |
| Component | Core |
| Priority | minor |
| Milestone | 1.8.1 |
| Keywords | botnet |
| Version | 1.8.0 CVS |

add option to control changes to user files
a bot would reject any userfile changes unless the bot requesting the change is 
set botattr +p

this would prevent a remote slave bot from altering the master bot's user files.

global-\* are not the answer.  they can only be used on the top master.

with multi-level linking, allowing the +p flag to control changes would allow 
intermediate bots to reject changes from slave bots linked to them but allow 
changes from the upstream master bot.

provides more protection from a less secure slave bot from being hacked and used 
to make mass changes on the secure master bot's records.

### LINKED ISSUE #7: Security flaws when more bots share ...
| Trac | Data |
| --: | :-- |
| Ticket | 7 |
| Reported by | TaKeDa |
| Status | new |
| Component | Core |
| Priority | minor |
| Milestone | 1.8.1 |
| Keywords | botnet sharing owner |
| Version | 1.8.0 CVS |

normaly, when you try remove or change host, pass etc. of
permanent owner it isn't possible, but when is sharing you can do it on other 
egg ...
I think, that maybe when you change pass or hosts etc when this information 
will get bot which have this user set to permanent owner the bot shouldnm't 
change anything but eventualy force his setting ...
but when is removing user... maybe bot before do this should ask other bots 
that they agree to do this and then remove ...
actualy the removing and changing password is dangerous ...
Every who is in botnet can change somebody password or remove him from botnet
even when he doesn't have access to bot which this person owns ...

Comment # 1 From Ben Dover 2000-10-26 10:14:11 [reply]

<rant>
I think this is a good example of why things should be discussed on the maillist 
before being added here for action.  Getting user input and discussion I think 
is a critical part of the process and should not be depreciated.  
</rant>

That said, this needs thought and may have other concerns.  [that is why 
involving more in active discussion might help clarify a path to proceed on]

Ultimately permowners are permowners of bots, not permowners of the botnet.  
What if a shell of a slave bot is hacked and someone adds new permowners?  
Should someone on the master bot be able to deflag that illegal user over the 
botnet?  or should the hacked permowner status protected?

I think much of the concerns you express would be solved by the sharing changes 
I proposed a few times before.  See bug # 59.  with the option proposed, the +p 
flag would control this, controlling not only userfile shares, but changes to 
the userfile.  iow, a slave could not change ANY records on an upstream bot.  
changing owner addys and passes are not the only thing that needs protection.

Comment # 2 From TaKeDa 2001-01-15 02:59:43

Yes but this will allow changes only from hub, I need to have possibility to 
change userfile from leaf (because I don't want to give +n access everyone bot 
owner on mine hub) of course hub filters all global flags.

Comment # 3 From TaKeDa 2002-10-02 23:37:26

I already wrote mine own solution to that problem, it is part of
multilevel_sharing (I'm talking about additional flag +s (+r in patches for
eggdrop 1.6.10 and older) which is protecting selected users from changing them
on other bot than main hub)

Comment # 4 From TaKeDa 2002-10-02 23:40:58

this bug is still in current eggdrop

Comment # 5 From Will Buckner 2004-06-10 15:52:15

Moving this to Eggdrop 1.7 product.

Cizzle:
Found by: Ben Dover
Patch by: Cizzle
Fixes: #1 

One-line summary: Introduces new bot flags ```bcejnu``` to allow a passive bot (slave) of sharing changes to bans, channels, exempts, invites, ignores and users respectively with an aggressive bot (hub). Also introduces bot flag ```d``` for when you don't want to share anything, as bot flag ```s``` will still allow anything to be shared.

Additional description: This partially fixes #7 as well, or even completely depending on the stance if permowners should be extra protected from bots you've allowed to make changes or not.

---Edit--- Had to change botflag v to j as it turned out botflags v to z aren't usable.
---Edit2--- Changed botflag s to still allow anything to be shared for backward compat and added botflag d to disable sharing anything.

Test cases demonstrating functionality:
On HubBot:
```
.+bot SlaveBot localhost 3344
.botattr SlaveBot +s
```
On SlaveBot:
```
.+bot HubBot localhost 3333
.botattr HubBot +hp
.+user TestUser
```
The user TestUser will be created on the HubBot. Now we will disable this by issuing on HubBot:
```
.unlink SlaveBot
.botattr SlaveBot -s+d
```
And on SlaveBot:
```
.+user TestUser2
```
The user TestUser2 will not be created on the HubBot. Now we will allow this by issuing on HubBot:
```
.unlink SlaveBot
.botattr SlaveBot +u
```
Now when SlaveBot tries to add a user, change a nick etc, it will be propagated to HubBot.
Be aware that botflag d can be kept in conjunction with any of the bcejnu flags, those override botlfag d's behavior.


vanosg:
Build finished.

vanosg:
Build finished.

src/mod/share.mod/share.c
```diff
@@ -1375,9 +1378,17 @@ static void sharein_mod(int idx, char *msg)
...
       /* Found a match */
-      (C_share[i].func) (idx, msg);
+      struct flag_record fr = { FR_BOT , 0, 0, 0, 0, 0 };
+      struct flag_record req = { FR_BOT | FR_OR, 0, 0, 0, 0, 0 };
+
+      break_down_flags(C_share[i].flags, &req, NULL);
+      get_user_flagrec(dcc[idx].user, &fr, NULL);
+      if (flagrec_eq(&req, &fr)) {
+        (C_share[i].func) (idx, msg);
+      }
```
thommey, line 1394: Should we issue some kind of log message for this?

vanosg:
So the piece that I hadn't realized yet- this allows linked bots, sharing userfiles, to become desynchronized. If I add the leafbot on the hub with a bottattr of +n and then try and do a .+user user1 on the leafbot, the leafbot still adds that user to the local userfile. I guess linked bots that don't share userfiles are already going to be in the exact situation so it's not crazy to see this state, but that user would be over-written on the next userfile download from the hub, which doesn't happen on non-sharing linked bots.

I'm not advocating for or against any specific change here, but I would like to spotlight this 'feature' for discussion. I think in an ideal world, if this flag was set then the leaf wouldn't take action locally either, but I don't know that that is possible within our existing structure.

Issue #14, opened 2010-02-03
chon binds triggered when a control proc releases an idx


thommey:
| Trac | Data |
| --: | :-- |
| Ticket | 14 |
| Reported by | Tothwolf |
| Status | new |
| Component | Core |
| Priority | blocker |
| Milestone | 1.8.0 |
| Version | 1.8.0 CVS |

chon binds are being trigged when relesing an idx from a control loop. chon binds are not triggered when exiting the filesys, etc, but for some reason they are when the control proc returns 1 to release control of the idx back to the partyline.

thommey:
@Pixelz:
Priority change to blocker for 1.8.0 because of a possible change in the Tcl API behavior, although undocumented.

PR #484, opened 2017-11-23
Do not trigger CHON binds when returning from a control script.

Cizzle:
Found by: Tothwolf
Patch by: Cizzle
Fixes: #14 

One-line summary: When returning from a control proc (.tcl control <idx> <procname>) we no longer trigger CHON binds.

Additional description: It's also more clear now that we're returning from the script and the text has been made translatable. Other already existing usages of the text were replaced as well.

Test cases demonstrating functionality:
Source the following Tcl script:
```
bind chon -|- * chonproc
proc chonproc {hand idx} { putlog "We got: $hand and $idx" }
proc choncontrol {idx text} {
        putlog "Getting text from $idx: $text"
        if {$text eq "quit"} {
                putdcc $idx "Ending this nonsense!"
                return 1
        }
        return 0
}
```
Then have two partyline connections open of which you know which has what IDX (see .dccstat). Then on either one run (with IDX their idx):
```
.tcl control IDX choncontrol
test
quit
```

### DIFF
--- a/language/core.danish.lang
+++ b/language/core.danish.lang
@@ -401,7 +401,7 @@ Telnet botten og skriv 'NEW' som dit nick.\n
 0xe16,DCC forbindelse lukket (%s!%s)
 0xe17,Fejlede indgående TELNET (%s)
 0xe18,Afviste %s/%d (ugyldig kilde port)
-# 0xe19 - unused / ubrugt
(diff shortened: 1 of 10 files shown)


thommey:
LGTM after indention change in #define DCC_BADIP

PR #486, opened 2017-11-27
Prevent loop when writing to stdout in foreground mode causing segfault.

Cizzle:
When running the bot in foreground mode and console mode +ob and with the bot linked via TLS to another bot, a segfault will occur upon killing the bot (either .die or CTRL+C).


vanosg:
Replicated with segfault w/o patch, no segfault with patch.

LGTM

PR #489, opened 2017-11-28
Trim/update tcl.m4

vanosg:
Found by: NA
Patch by: Geo
Fixes: NA

One-line summary:
Replace TEA full tcl.m4 with Eggdrop-specific portions + updated search paths


Cizzle:
If configuration and compilation still works on the supported systems, this LGTM. Can't more stuff be left out regarding TK?

vanosg:
I don't see much left in the way of TK references... I'm going to merge this in so I can hit it with the compile farm, but please continue to hack away at it if you or thommey can further improve.

Issue #490, opened 2017-12-02
copy-to-tmp 1 breaks sharing


thommey:
with the new copy-to-tmp setting/code, it seems like copy-to-tmp being 1 breaks sharing under some circumstances (only observed with ssl so far).
there seems to be garbage memory before the actual userfile (45,000 bytes in one instance).
client stores the userfile but fails on loading it (no sanity check apparently), suggesting scripts/weed conversion and dies.

Found by: OUTsider (helped track down the issue)

PR #492, opened 2017-12-02
Use tmpfile for receiving userfile when copy-to-tmp is enabled.

Cizzle:
Found by: OUTsider, simple, Geo, ...
Patch by: Cizzle
Fixes: #490 

One-line summary: The userfile being received is now properly being sent to a tmpfile first if copy-to-tmp is set.


thommey:
LGTM, after doc update, needs config/documentation rewording ("for sending")

PR #497, opened 2017-12-18
Remove false SSL noise.

Cizzle:
Ignores the close notify warnings which are expected to end SSL connections and ignores false errors that appear when not enough data is present yet for non-blocking sockets.

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -648,9 +648,23 @@ void ssl_info(SSL *ssl, int where, int ret)
            (where & SSL_CB_READ) ? "read" : "write",
            SSL_alert_type_string_long(ret),
            SSL_alert_desc_string_long(ret));
-  } else if (ret <= 0 && where & SSL_CB_EXIT) {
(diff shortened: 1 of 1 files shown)


vanosg:
On a leaf bot with mkcobxs enabled on console, I got none of the 'usual' noise. 

Adding the +d flag gives me stuff like
```
[03:35:00] TLS: state change: unknown state
[03:35:00] TLS: state change: unknown state
[03:35:00] TLS: state change: unknown state
[03:35:00] TLS: state change: unknown state
[03:35:00] TLS: awaiting more reads
[03:35:00] TLS: awaiting more reads
[03:35:00] sockread EAGAIN: 10 11 (Resource temporarily unavailable)
[03:35:00] TLS: awaiting more reads
[03:35:00] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 10
--[03:35:00] TLS: state change: unknown state
[03:35:00] TLS: state change: unknown state
```
which seems like a totally appropriate place to stash those messages. 

LGTM

Issue #498, opened 2017-12-18
Percent symbol magically dropped (~ 1/2 of them)


VertigoRN:
What wrong with "%" in botnet? It dropped 1/2 of them on version 1.8.2+addbot:
![default](https://user-images.githubusercontent.com/34381201/34118575-4424904a-e430-11e7-95ab-7710069ce813.png)

I returned to released version (1.8.2) and problem is gone:
![default](https://user-images.githubusercontent.com/34381201/34118587-4cac389e-e430-11e7-91ff-32afd9f085dd.png)

PR #499, opened 2017-12-19
Fix percentsign being interpreted by format printers.

Cizzle:
Found by: VertigoRN
Patch by: Cizzle
Fixes: #498 

One-line summary: Fixes the percentsign being dropped half the time for certain botnet commands.

### DIFF
--- a/src/botmsg.c
+++ b/src/botmsg.c
@@ -42,7 +42,7 @@ static char OBUF[1024];
  */
 void tandout_but EGG_VARARGS_DEF(int, arg1)
 {
-  int i, x;
+  int i, x, len;
   char *format;
   char s[601];
   va_list va;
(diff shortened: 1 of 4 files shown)


src/botmsg.c
```diff
@@ -42,21 +42,24 @@ static char OBUF[1024];
...
   va_list va;
 
   x = EGG_VARARGS_START(int, arg1, va);
   format = va_arg(va, char *);
 
-  egg_vsnprintf(s, 511, format, va);
+  len = egg_vsnprintf(s, 511, format, va);
   va_end(va);
-  s[sizeof(s) - 1] = 0;
+  if (len >= sizeof(s)) {
```
thommey: we only write 511, so to detect truncation (\0 skipping), we need to check for 511

src/botmsg.c
```diff
@@ -196,13 +199,17 @@ void send_tand_but(int x, char *buf, int len)
   int i, iso = 0;
 
   if (len < 0) {
+    /* -INT_MIN is INT_MAX+1 which is... INT_MIN */
```
thommey: I would suggest removing the check, breaks with one's complement (rare)

thommey (changes requested):

PR #532, opened 2018-07-26
Update config.sub/guess

vanosg:
Found by: odidev, lag-linaro

One-line summary:
Among other benefits of updating the 10 year old file, adds support for arm64v8 architecture


Cizzle:
LEBTM (Even Better).

PR #533, opened 2018-07-27
Isolate global isban/is[*] checks from channel checks

### LINKED ISSUE #531: ISBAN: overlapping functions
Documentation says:

isban <ban> [channel]

  Returns: 1 if the specified ban is in the global ban list; 0
  otherwise. If a channel is specified, that channel's ban list is
  checked as well.

Question: how can you tell if a ban was added BOTH globally and channel-specific?

Coding a check for { [isban * #] && ![isban *] } doesn't work because, while it will isolate bans that are NOT global, it will also hide those that are BOTH global and local.

This is important if a bot master is trying to clean up the ban list, they can't "SEE" why a ban may be re-triggering (especially is STICKY) if it can't be seen / listed.

Solution: ISBAN should either show global (if no chan is listed), or local [channel]-only bans (if a channel is specified). This current "overlap" of returning globally-matching bans when a channel is specified is causing hell!

vanosg:
Found by: maimizuno
Patch by: Geo
Fixes: #531 

One-line summary:
Stops isban (and other is* commands) from returning 1 on a channel query when only a global ban/whatever exists. 

```
.bans *!*@test.com
Global bans:
! [  1] *!foo@test.com (perm)
        vanosg: requested
        Created 03:34
Channel bans for #foober:  (* = not placed by bot)

.tcl isban *!foo@test.com
Tcl: 1
.tcl isban *!foo@test.com #foober 
Tcl: 0

### Now add a channel ban, so both global and channel exist:
.+ban *!foo@test.com #foober
New #foober ban: *!foo@test.com (requested)
.tcl isban *!foo@test.com #foober 
Tcl: 1
.tcl isban *!foo@test.com
Tcl: 1

### Now remove global, so only channel ban exists:
.-ban *!foo@test.com
Removed ban: *!foo@test.com
.tcl isban *!foo@test.com         
Tcl: 0
.tcl isban *!foo@test.com #foober 
Tcl: 1

### DIFF
--- a/doc/sphinx_source/mainDocs/tcl-commands.rst
+++ b/doc/sphinx_source/mainDocs/tcl-commands.rst
@@ -659,75 +659,75 @@ ischanjuped <channel>
 
   Module: channels
 
-^^^^^^^^^^^^^^^^^^^^^
-isban <ban> [channel]
-^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+isban <ban> [channel [-channel]]
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-  Returns: 1 if the specified ban is in the global ban list; 0 otherwise. If a channel is specified, that channel's ban list is checked as well.
-
-  Module: channels
-
-^^^^^^^^^^^^^^^^^^^^^^^^^
-ispermban <ban> [channel]
-^^^^^^^^^^^^^^^^^^^^^^^^^
-
-  Returns: 1 if the specified ban is in the global ban list AND is marked as permanent; 0 otherwise. If a channel is specified, that channel's ban list is checked as well.
+  Returns: 1 if the specified ban is in the global ban list; 0 otherwise. If a channel is specified, that channel's ban list is checked as well. If the -channel flag is used at the end of the command, \*only\* the channel bans are checked.
 
   Module: channels
 
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
-isexempt <exempt> [channel]
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+ispermban <ban> [channel [-channel]]
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by michaelortmann** [comment]
the function specification reads "as well" not "instead":

isban <ban> [channel]

  Returns: 1 if the specified ban is in the global ban list; 0
  otherwise. If a channel is specified, that channel's ban list is
  checked as well.

  Module: channels

not returning 1 if a ban is in the global ban list would break this specification. existing tcls may rely on it. instead of api change i propose to change the function in a backward compatible way. something like:

isban <ban> [channel] [flag]

if flag not set: old behaviour

if flag set: new behaviour

or find another/ a better solution for the problem this issue is based on.

**Comment by maimizuno** [comment]
Bingo.

**Comment by Cizzle** [comment]
Docs should be changed too and the `!ok` is obsolete in the first 3 ifchanges.
As this changes the behavior, the presented solution would be for 1.9.0? Otherwise you'd have to add an option to the command(s) to indicate you only want to search the #channel list (like `isban * #eggdrop 1`) which then again would make scripts not work on older bots.

**Comment by vanosg** [comment]
Yeah... I really wanted this to not wait for 1.9.0, but if I'm being honest, it does constitute API change, as much as I disagree with how it was originally written (ie feature, not bug). Meh!

**Comment by thommey** [comment]
I don't like breaking scripts that way in 1.9.0 or any version. Can't we introduce "ischannelban" to check a specific channel banlist only? The bad thing about this one is that it's a behavior change, not an API change, and it will affect existing scripts that won't get touched.

**Comment by vanosg** [comment]
OK, reworked this one. I still say this feature was incorrectly written, but its been like this for 30 years, so... ok. Added a '-channel' flag to restrict the check to only a channel. Test cases as such:
```
Global bans:   (! = not active on #eggdroptest)
! [  2] *!*@soo.com (perm)
        -HQ: $-1y
        Created 508 days ago
Channel bans for #eggdroptest:   (! = not active, * = not placed by bot)
! [  6] *!*@foo.com (perm)
        -HQ: requested
        Created 02:58

### Existing functionality, check that *!*@soo.com is on either global or channel list (here, the global list)
.tcl isban *!*@soo.com #eggdroptest
Tcl: 1
### New functionality, restrict check to only the channel provided
.tcl isban *!*@soo.com #eggdroptest -channel
Tcl: 0
### Old functionality, check that *!*@foo.com is on channel or global list (here, the channel list)
.tcl isban *!*@foo.com #eggdroptest
Tcl: 1
### Old functionality, check that *!*@foo.com is on the global list (here, the channel list)
.tcl isban *!*@foo.com
Tcl: 0

PR #536, opened 2018-07-30
enhance seed for random

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: seed / srandom

One-line summary:
enhance seed / srandom

srandom is currently seeded like this:
  srandom((unsigned int) (now % (getpid() + getppid())));
the part:
  % (getpid() + getppid())"
truncates the range to maxpid+maxppid, which is 128k
the patch enhances the seed in 3 ways
1st: the truncation is fixed by replacing the "mod (pid + ppid)" by "xor pid"
2nd: "now" is in seconds and is replaced with tv_sec * tv_usec (milliseconds)
3rd: if the modern getrandom() libc function is available, it is used instead
think of getrandom() as /dev/urandom, but without any file descriptor
its available under freebsd12+, linux glibc 2.25+, solaris 11.3+
this patch enhances (only) the seed for srandom
and doesnt change any other random functions

i testet the new code with and without HAVE_GETRANDOM


configure
```diff
@@ -729,7 +729,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
```
Cizzle: Please use an updated autoconf to not encounter this undesirable side effect.

Cizzle:
Rest looks good but not sure where this stands in regards to keeping in line with ~~C99~~C89 and POSIX 2001 @thommey ?

michaelortmann:
this patch is c89 and posix2001

tested with:

gcc -march=native -O2 -pipe -fstack-protector-strong -fno-plt -fomit-frame-pointer -pipe -Wall -I.. -I..  -DHAVE_CONFIG_H -I/usr/include -g3 -DDEBUG -DDEBUG_ASSERT -DDEBUG_MEM -DDEBUG_DNS  '-DCCFLAGS="gcc -march=native -O2 -pipe -fstack-protector-strong -fno-plt -fomit-frame-pointer -pipe -Wall -I.. -I..  -DHAVE_CONFIG_H -I/usr/include -g3 -DDEBUG -DDEBUG_ASSERT -DDEBUG_MEM -DDEBUG_DNS "' '-DLDFLAGS="gcc"' '-DSTRIPFLAGS="touch"' -Wno-format-truncation -std=c89 -D_POSIX_C_SOURCE=200112L -c ./main.c

(pls ignore the -Wno-format-truncation, i dont want this here)

results in 2 issues, both not related to that patch here.

issue 1 is "inline" keyword, which would need c99 but i would prefer it being removed, because compilers often know better what to inline and what not:
static inline void garbage_collect(void)

issue 2 is comments in new getopt() code. those are new style // instead of old style /* */

we could easily make main.c c89+posix2001, but not with this patch this time.

src/main.c
```diff
@@ -1016,6 +1020,19 @@ int mainloop(int toplevel)
   return (eggbusy || tclbusy);
 }
 
+void init_random() {
+  unsigned int seed;
+#ifdef HAVE_GETRANDOM
+  if (getrandom(&seed, sizeof(seed), 0) != sizeof(seed))
+    fatal("ERROR: getrandom()\n", 0);
+#else
+  struct timeval tp;
```
thommey, line 1029: I think we can make the function static, as it is only used here, if you agree. functions with no arguments should have "void" as argument specification, init_random() means "unspecified arguments". `static void init_random(void)` would be my proposal.

thommey (commented):
looks good to me, just recommend changing the init_random definition, then it's fine, thank you :)

src/main.c
michaelortmann: made it static now. thank you. but init_random() {...} is function definition, not function declaration. so it does not mean unspecified arguments in this case.

vanosg:
Looks to be done- any other comments @thommey @Cizzle ?

Cizzle:
This PR includes a commit from another PR, not sure if this could mess up a merge.

michaelortmann:
merging upstream/develop fixed it. ty

Cizzle:
LGTM

Issue #527, opened 2018-06-26
Failures and warnings when running make config and make install with --jobs > 1


sbraz:
When running `make -j13 DEST=/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt/eggdrop install`, we get:
```
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.              
make[1]: Entering directory '/var/tmp/portage/net-irc/eggdrop-1.8.3/work/eggdrop-1.8.3/scripts'                                                 
/usr/bin/install: cannot create regular file '/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt/eggdrop/eggdrop-1.8.3': No such file or directory
/usr/bin/install: target '/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt/eggdrop/' is not a directory: No such file or directory
Creating 'scripts' subdirectory.                            
make: *** [Makefile:352: install-bin] Error 1                                                                                                                                                                      
make: *** Waiting for unfinished jobs....                                                                                                                                                                          
make: *** [Makefile:390: install-data] Error 1                                                                                                                                                                    
/usr/bin/install: cannot create regular file '/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt/eggdrop': No such file or directory
mkdir: cannot create directory ‘/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt’: File exists
make: *** [Makefile:466: install-doc] Error 1                                                 
mkdir: cannot create directory ‘/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt’: File exists
mkdir: cannot create directory ‘/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt’: File exists
mkdir: cannot create directory ‘/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt’: File exists  
mkdir: cannot create directory ‘/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt’: File exists         
mkdir: cannot create directory ‘/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt/eggdrop’: File exists
mkdir: cannot create directory ‘/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt/eggdrop’: File exists
mkdir: cannot create directory ‘/var/tmp/portage/net-irc/eggdrop-1.8.3/image/opt/eggdrop’: File exists
[...]

PR #537, opened 2018-07-31
Fix main Makefile for parallellisation related to issue #527

Cizzle:
Found by: sbraz
Patch by: Cizzle
Fixes: #527 

One-line summary: Fixes parallellisation for several `make` targets, most important being `config` and `install`.


thommey (approved):
LGTM. We can skip the BSD checks, it looks fine

PR #539, opened 2018-08-07
Include '+' port prefix in $server. Fixes #538

### LINKED ISSUE #538: TLS $::server / SSL
$::server does not show "+" to indicate an SSL connection. It looks the same as any regular non-SSL connection.

vanosg:
Found by: maimizuno
Patch by: Geo
Fixes: #538 

One-line summary:

```
.tcl putlog $server
[15:58:58] barjavel.freenode.net:+7000
```
and
```
.tcl putlog $server
[16:00:21] barjavel.freenode.net:6667
```

### DIFF
--- a/src/mod/server.mod/server.c
+++ b/src/mod/server.mod/server.c
@@ -1267,7 +1267,12 @@ static char *traced_server(ClientData cdata, Tcl_Interp *irp,
     int servidx = findanyidx(serv);
 
     /* return real server name */
+#ifdef TLS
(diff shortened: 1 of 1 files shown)


vanosg:
Altering output *could* break a script... does this need to wait until 1.9?

Cizzle:
LGTM.
Looking at the git history/blame I don't think this was intentionally left out but was forgotten and is as such a bug which need fixing ASAP.

Issue #226, opened 2016-07-17
Update non-needed compat code


vanosg:
Lots of code seems unnecessary now, particularly in the compat section. Trim the fat!

ghost:
?

thommey:
This is speaking about the src/compat/ directory that has a lot of compatiblity functions we can remove, because we require POSIX 2001.

thommey:
And check the licenses of the code that remains

Cizzle:
Reopening as not all have been checked yet.

Cizzle:
Reopening as we still have (at least) strftime (#635).
Others that are subject to removal: inet_pton, inet_ntop, strcasecmp, snprintf.
Other inet related functions might too if we change more to use getaddrinfo and getnameinfo.

PR #540, opened 2018-08-09
Remove compat memset (eggdrop requires posix 2001)

michaelortmann:
Found by: vanosg
Patch by: michaelortmann
Fixes: #226 

One-line summary:
memset as part of issue 226

eggdrop depends on posix 2001. posix 2001 defines memset(). see http://www.unix.org/version3/inttables.pdf. this patch gets rid of src/compat/memset*. egg_bzero() and egg_memset() is left as is, but redefined to directly use memset() with a one-liner define in eggdrop.h. we only keep egg_memset() for backward compatibility to older modules, and could remove it from later eggdrop versions when api breakage becomes possible. while at it, i replaced some forgotten memset(dest, 0, n) with egg_bzero() and removed one register keyword in compress.c. maybe we should open an issue calling for removal of most of register keywords within eggdrop. modern compilers are good at register allocation.

i added compress mod to eggdrop.conf and started eggdrop as minimal test. so far so good.


src/modules.c
```diff
@@ -543,7 +543,7 @@ Function global_table[] = {
   /* 252 - 255 */
   (Function) egg_snprintf,
   (Function) egg_vsnprintf,
-  (Function) egg_memset,
+  (Function) 0,                   /* was egg_memset -- use memset() instead */
```
Cizzle: Comment should rather point to egg_bzero?

Cizzle:
LGTM apart from the nitpick comment. Merger should be cautious not to automatically close #226.

src/modules.c
michaelortmann: its done

Cizzle:
LGTM²

michaelortmann:
removing a function from table in src/modules.c looks like module api change? maybe delay for 1.9.0 instead of 1.8.x?

vanosg:
Agree - however small the chance is, some other module could be calling egg_memset. Pending any objections, I'll move this to 1.9.0.

Aside from that, nice modernization to codebase. Always fun to see the code size actually shrink after a PR!

src/modules.c
```diff
@@ -543,7 +543,7 @@ Function global_table[] = {
   /* 252 - 255 */
   (Function) egg_snprintf,
   (Function) egg_vsnprintf,
-  (Function) egg_memset,
+  (Function) 0,                   /* was egg_memset -- use memset() or egg_bzero() instead */
```
thommey, line 546: could we export memset if it has the same signature at this point to avoid breaking modules?

thommey:
Did you manually remove memset.h from Makefile.in or did you run "misc/makedepend"? If not, misc/makedepend would be preferred so it becomes a no-op when we run it in the future (line breaks etc.)

src/modules.c
michaelortmann: yes. done.
src/eggdrop.h
#define egg_memset(dest, c, len) memset(dest, c, len)
i did a quick test, and egg_memset was once again useable in module.

michaelortmann:
now i also did run misc/makedepend.

michaelortmann:
because egg_memset is once again useable by modules, i hereby request milestone 1.8.4 instead of 1.9.0.

thommey (approved):
LGTM

vanosg:
builf

vanosg:
build

PR #543, opened 2018-08-10
(WIP) fix stringop-truncation warning

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: stringop-truncation warning

One-line summary:
fixes the only stringop-truncation warning

modern gcc throws exactly 1 direct stringop-truncation warning on eggdrop. this patch fixes this. i guess the original author found no other solution of setting bf_mode. i tried #define which crashed and *bf_mode = "ecb" which doesnt work. the solution is bfmode[] = "ecb".

there are another 13 indirect warnings, but those disappear, if you would set -Wno-format-truncation.

./eggdrop -ntm
[...]
.set blowfish-use-mode
[03:51:47] #-HQ# set blowfish-use-mode
Currently: ecb

### DIFF
--- a/src/mod/blowfish.mod/blowfish.c
+++ b/src/mod/blowfish.mod/blowfish.c
@@ -34,7 +34,8 @@
 #undef global
 static Function *global = NULL;
 
-static char bf_mode[4];
+/* ECB by default for now, change at v1.9.0! */
+static char bf_mode[] = "ecb";
 
 /* Each box takes up 4k so be very careful here */
 #define BOXES 3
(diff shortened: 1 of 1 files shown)


michaelortmann:
i need more time thinking about this compiler warning, there are more ways (do nothing is one of those) to handle this warning. yes i know, it is a false alarm. my updated guess is, after we replace the strncpyz() macro with strlcpy() (extern/intern) as i want to do with code, which is currently part of #549 this and some more warnings could disappear for free.

michaelortmann:
#592 will fix this stringop-truncation warning

PR #546, opened 2018-08-12
Fix fread checks.

Cizzle:
Found by: Cizzle
Patch by: Cizzle
Fixes: 

One-line summary: Several fread checks weren't properly checking. This caused amongst others issues transfering (user)files.

### DIFF
--- a/src/mod/filesys.mod/dbcompat.c
+++ b/src/mod/filesys.mod/dbcompat.c
@@ -162,7 +162,7 @@ static void convert_version1(FILE *fdb_s, FILE *fdb_t)
       fdbe->gots = fdb1.gots;
       if (fdb1.sharelink[0])
         malloc_strcpy(fdbe->sharelink, fdb1.sharelink);
-      filedb_addfile(fdb_s, fdbe);
+      filedb_addfile(fdb_t, fdbe);
       free_fdbe(&fdbe);
     }
   }
--- a/src/mod/filesys.mod/dbcompat.c
+++ b/src/mod/filesys.mod/dbcompat.c
@@ -144,28 +144,26 @@ static void convert_version1(FILE *fdb_s, FILE *fdb_t)
   filedb1 fdb1;
 
   fseek(fdb_s, 0L, SEEK_SET);
-  while (!feof(fdb_s)) {
-    if (fread(&fdb1, 1, sizeof fdb1, fdb_s) && !feof(fdb_s)) {
-      if (!(fdb1.stat & FILE_UNUSED)) {
-        filedb_entry *fdbe = malloc_fdbe();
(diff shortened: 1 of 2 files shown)


src/mod/filesys.mod/dbcompat.c
```diff
@@ -145,8 +145,7 @@ static void convert_version1(FILE *fdb_s, FILE *fdb_t)
 
   fseek(fdb_s, 0L, SEEK_SET);
   while (!feof(fdb_s)) {
-    fread(&fdb1, sizeof(filedb1), 1, fdb_s);
-    if (!feof(fdb_s)) {
+    if (fread(&fdb1, 1, sizeof fdb1, fdb_s) && !feof(fdb_s)) {
```
michaelortmann: lets examine all results any fread() may have:
(1) return > 0, this automatically also means !feof ; the case return > 0, feof is never possible
  [ so: "if fread && !feof" is the same as "if fread"; this new redundant code doesnt look right somehow ]
(2) return 0, feof ; this is the normal eof case
(3) return 0, !feof ; this means an error happened

for case (1) and (2) the new code doesnt change any branch logic. the while and the inner if is branched like the old code and looks fine.

for case (3) there is a difference
the inner if is not executed anymore with the new code
that looks fine, because in this case of fread-error, we probably dont want to do stuff with the result of fread()
BUT: the while logic is still the same, feof is not set in case of error i guess(if i guess wrong, then the old code would work fine also!)
i guess we may (still) loop endlessly in case of error-fread()
[ this is a classic fread() programmers trap ]
has this part of the code triggered any fread error handling problem anyway?
i would not touch it if it hasnt

TLDR-Version: it was a good idea to check fread() > 0, but new code has new problems and doest fix endless loop problem in case of fread error.

i propose code like this:

while ((bytes = fread(fdb_s)) > 0) {
}
if eof(fdb_s) { // if we want to fclose(fdb_s) now
}
else if ferror(fdb_s) { // we should want to handle the error
}

Cizzle: I guess I should've been even more thorough with the `fread`'s, like you have been; my main focus was on checking their return values and the order of arguments.
I'm not aware of past problems with this part of the code but as you've pointed out improvements can be done. Your proposition looks like a good idea.

michaelortmann: i was wrong with (1) so my comment can be confusing. thx Crizzle

michaelortmann:
i think there is another(?) error in (no longer used?) function convert_version1():
filedb_addfile(fdb_s) should be filedb_addfile(fdb_t)

michaelortmann:
compiling with gcc and CFLAG -D_FORTIFY_SOURCE in context of #549 now independently also spotted the unused fread() result. good you patch it here already.

michaelortmann:
LGTM

PR #547, opened 2018-08-12
(WIP) compress mod requires share mod

### LINKED ISSUE #401: Compress module tethering with share module
Sharing module isn't loaded; compress is (I want to use the COMPRESSFILE/UNCOMPRESSFILE commands. but not userfile share!):

#### SHARE MODULE ####
#loadmodule share
#set allow-resync 0
#set resync-time 900
#set private-global 1
#set private-globals nmtolv
#set private-user 1
#set override-bots 0

#### COMPRESS MODULE ####
loadmodule compress

Apparently, COMPRESS has a dependency ([MODULES] compress 1.2 {share 2.3} {eggdrop 108.0}). This means that SHARE is subsequently loaded (which I don't want the memory footprint / overhead, thus, not loaded!!).

This means: to use commands like COMPRESSFILE / UNCOMPRESSFILE, I have to load the SHARE module? I know it's auto-loading (I don't want it) because:

\<Dorito> Module unloaded: compress
\<Dorito> Sending 'share end' to all sharebots...
\<Dorito> Unloaded sharing module, flushing tbuf's...
\<Dorito> Module unloaded: share
\<Dorito> Unloading transfer module, killing all transfer connections...

Is the SHARE module really necessary to use COMPRESS for the COMPRESSFILE/UNCOMPRESSFILE options? If so, shouldn't the following also be true:

(1) TCL-COMMANDS.DOC (which is a TEXT file, not a DOC file) should show the module dependencies: "YOU MUST LOAD THE SHARE MODULE TO USE THESE COMPRESS MODULE COMMANDS!!"

[or]

(2) the modules should COMPLAIN via PUTLOG when a dependancy is missing, let you decide what to you about it (sacrifice the option of using COMPRESSFILE/UNCOMPRESSFILE)

[or]

(3) CONFIG variable setting: $::autoload-depencies (if 0, fatal-error the bot; if 1, auto-load dependency modules as needed)

michaelortmann:
…share mod - quick hack but many open questions

Found by: maimizuno
Patch by: michaelortmann
Fixes: #401 

One-line summary:
compress mod does not require share mod (quick hack, many open questions) and typo fix (provides provides -> provides

compress mod still does the "depend mod" test on share mod which loads share.mod in case it exists/is loadable. but if it fails, it no longer depends on it, but only reduces functionality. this is a quick hack. it solves the issue, one wants tcl compress commands but one doesnt want share mod. it requires the user to make share mod unloadable. config unsetting is not sufficient, but removing share mod from filesystem would be. you guess right, this quick hack leaves open and new questions and if applied as is, would certainly raise new issues like "do we need a setting controlling autoload of depend mods" or "what is the order of module loading and should it matter". and what, if the user loads compress without share, and then later decides to also load share. wouldn't she then expect compression support for share?

i did make install, removed the share mod, started eggdrop, loaded the compress mod and tested .tcl compressfile.

### DIFF
--- a/src/dns.c
+++ b/src/dns.c
@@ -503,6 +503,7 @@ void *thread_dns_hostbyip(void *arg)
   i = getnameinfo((const struct sockaddr *) &addr->addr.sa, addr->addrlen,
                   dtn->host, sizeof dtn->host, NULL, 0, 0);
   if (i) {
+    debug1("dns: thread_dns_hostbyip(): getnameinfo(): error = %s", gai_strerror(i));
 #ifdef IPV6
     if (addr->family == AF_INET6)
       inet_ntop(AF_INET6, &addr->addr.s6.sin6_addr, dtn->host, sizeof dtn->host);
@@ -520,13 +521,13 @@ void *thread_dns_hostbyip(void *arg)
 void *thread_dns_ipbyhost(void *arg)
 {
   struct dns_thread_node *dtn = (struct dns_thread_node *) arg;
-  struct addrinfo *res0, *res;
-  int i;
+  struct addrinfo *res0 = NULL, *res;
+  int error;
   sockname_t *addr = &dtn->addr;
 
-  i = getaddrinfo(dtn->host, NULL, NULL, &res0);
+  error = getaddrinfo(dtn->host, NULL, NULL, &res0);
   memset(addr, 0, sizeof *addr);
-  if (!i) {
+  if (!error) {
 #ifdef IPV6
     for (res = res0; res; res = res->ai_next) {
       if (res == res0 || res->ai_family == (pref_af ? AF_INET6 : AF_INET)) {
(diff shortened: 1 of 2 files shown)


Cizzle:
I'm afraid I can't support quick hacks to be added to the codebase. The typofix is fine of course but should be singled out.

michaelortmann:
besides singling out the minor typo fix, what to do next here?

thommey:
Yeah this needs a broader fix, which allows checking for a loaded module without depending on it. And a hook the "other way". My suggestion would be this:

When compress gets loaded, it can check whether share.mod exists or not, without loading it.
When share.mod gets loaded, it can check whether compress.mod is loaded, and then call a hook in compress.mod to tell it that it is now available, and compress can "re-initialize" or whatever that changes.

The hooks can be hard-coded between those 2 modules, that's fine (exported function in compress.mod that share.mod calls to let it know it's available).

Would need to figure out if those are unloadable, and if so, maybe allow unloading share.mod with the same mechanic.

PR #549, opened 2018-08-14
(WIP) be portable and c89 and posix 2001

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: be portable and c89 and posix 2001

One-line summary:
be portable and c89 and posix 2001

be portable means, operating system dependency in net.c in open_telnet_raw() was fixed.
readability for getopt cli flags was fixed
strncpyz macro was replaced with strlcpy (this fixes -Wno-stringop-truncation and/or -Wno-stringop-overflow and using strncpy is mostly misuse, it doesnt equal strlcpy, also zero fills a buffer and was never intended as a safer version of strcpy. sometimes memcpy is a better solution)

Configured with: 'CFLAGS=-std=c89 -D_POSIX_C_SOURCE=200112L -D_XOPEN_SOURCE=500 -D_DEFAULT_SOURCE -Wno-format-overflow'

with those CFLAGS make doesn't throw any error or warning

which means, after this patch, eggdrop is pretty c89 and posix 2001

there could be more c86 issues not uncovered by now, maybe even more variable definitions outside function header, and i shall track them down also.

with additional -O2 CFLAG 7 stringop-truncation warnings occured and were fixed with commits:
  https://github.com/eggheads/eggdrop/pull/549/commits/2e534849c68d00c80383afebc644d3033b9f0cc1
  https://github.com/eggheads/eggdrop/pull/549/commits/c9f655bbefec2d29e352fe2d123359dad8cf3388
  https://github.com/eggheads/eggdrop/pull/549/commits/c7430b6b66467af881ad416315915a452e425517
  https://github.com/eggheads/eggdrop/pull/549/commits/dc70597a398baa78b9a4ba1f9d8a3adf8b977b06

eggdrop was tested and connected and joined some channel
https://github.com/eggheads/eggdrop/pull/549/commits/c7430b6b66467af881ad416315915a452e425517 was tested for bans and .tcl need-op:
  .bans must yield a string like Created 22:34, last used 22:45 correctly
  .chanset #c89 need-op test
  Successfully set modes { need-op { test } } on #c89.
strlcpy was testet w/o on linux and w/ on freebsd

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -291,6 +291,9 @@
 /* Define to 1 if you have the <string.h> header file. */
 #undef HAVE_STRING_H
 
+/* Define to 1 if you have the `strlcpy' function. */
+#undef HAVE_STRLCPY
+
 /* Define to 1 if you have the `strncasecmp' function. */
 #undef HAVE_STRNCASECMP
 
--- a/configure
+++ b/configure
@@ -6919,7 +6919,7 @@ ac_config_commands="$ac_config_commands $ac_stdint_h"
 
 
 # Checks for functions and their arguments.
-for ac_func in clock dprintf fsync getdtablesize getrusage inet_aton isascii mbrlen memcpy memset random rand lrand48 rename setpgid sigaction sigemptyset snprintf strcasecmp strncasecmp uname vsnprintf inet_ntop
+for ac_func in clock dprintf fsync getdtablesize getrusage inet_aton isascii mbrlen memcpy memset random rand lrand48 rename setpgid sigaction sigemptyset snprintf strcasecmp strlcpy strncasecmp uname vsnprintf inet_ntop
 do :
   as_ac_var=`$as_echo "ac_cv_func_$ac_func" | $as_tr_sh`
 ac_fn_c_check_func "$LINENO" "$ac_func" "$as_ac_var"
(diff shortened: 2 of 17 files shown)


michaelortmann:
i will come back here (and maybe split this pull request into smaller parts) after #568 has been committed.

thommey:
strlcpy isn't portable, what's the benefit over strncpyz?

michaelortmann:
strncpyz() uses strncpy(). strncpy is not what many assume. its not a safer strcpy() replacement, but a fixed size string copy. the big/practical difference is, that strncpy() will fill the rest of the destination buffer with \0. given a small src string: the larger the dest buffer the more the difference. you may argue: so only some cpu cycles wasted. and it would be true. but remember, we also get compiler warnings now about string truncation / format / underflows / etc. and some of those new warnings are caused not by wrong code, but by macro/function nesting magick happening regarding strncpyz(). we could also get rid of this, if we use a clean strlcpy() implementation instead.

sure, strlcpy() isnt portable if you view it from a certain angle. but let us view it from another one:
what we really want to do, when we use strncpyz() is what strlcpy() does, not some zero filling.
so: we replace our portable strncpyz() with  our portable strlcpy() code.
the function may even keep its name.
eggdrop would be c89/posix2001 here.
but. now. if the operating system has got strlcpy(), we should use it instead.
its optional. we don't give up any portability.
and i really think its the right way to do. ship portable function, but use operating system / libc if possible instead.

michaelortmann:
i will resume here, after #592

michaelortmann:
Meanwhile every single part of this PR is already merged into develop.

Issue #535, opened 2018-07-29
SSL


ghost:
I'm told that this has been discussed before, but I'm not going to search through over 500 entries to find it, so let's start fresh. Thanks to Pixelz for some of the background information.

According to everything I've read, the following are true:

(1) Eggdrop, if SSL is compiled in, can do SSL connections for botnet (if you listen +port)

(2) Eggdrop, if SSL is compiled in, will attempt to do STARTTLS if the bots, on both ends, are SSL-enabled but a non-SSL port is being used for the connection attempt

(3) Eggdrop, if SSL is compiled in, will *NOT* link bots if no certificate is available when STARTTLS is thrown (as per the config settings)

There is a specific problem as a result: for those on "admin [group-policy] limited windows systems" with WinDrop (which is based on Eggdrop, thus this being reported here, before anyone yells at me -- I'll yell right back), the use of OPENSSL may not be available to a given user in a multi-user environment (or a simple, single-user computer due to whatever) where certificate generation isn't wanted, desired, or possible. Between group policies and administrative lock-outs in various configurations make this a real issue.

Currently, I have to put up an off-line (NO-IRC, remember those?) hub bot just to hub a local cluster of automatically-SSL-compiled-in bots because I can't link them together through LOCALHOST. I don't have the option of getting the "non-SSL" port of WinDrop because the botnet they hook into *requires* SSL connections there (I can't make those consistently connect / hold). In addition there are several 1.6.13, 1.6.19, and 1.6.21 bots in there that can't handle the SSL connections anyway. Duct tape, spit, bailing wire, and magic is holding that part of the botnet together.

Now, why shouldn't I have the right (and control) to link my own LOCALHOST bots without being ***FORCED*** to set up certificates for each one??? Shouldn't I have the right to make that decision for myself?

I've been told that there was a suggestion "floating around" (not found in these 500+ issues) of using a port # of "-3333" where "-" says: DO-NOT-USE-SSL/STARTTLS-NO-MATTER-WHAT (in other words, "stay in plaintext"). It seems that, that suggestion either died or just wasn't supported. In either case: why?

Shouldn't we, as bot owners *of any kind and of any version and of any OS (given the various restrictions possible)* have control of that?

Revive the "-port" discussion.

skralg:
[...]

PR #552, opened 2018-08-16
Force no TLS on plaintext botlinks with -<port>

Cizzle:
Found by: simple, maimizuno, ...
Patch by: Cizzle
Fixes: #535 

One-line summary: Adds functionality to add "-" before a port in a botaddr, which forces keeping the link plaintext instead of trying STARTTLS.

 - Add a plain listen port first on the hub.
 - Enable console modes +gh on slave and hub
 - Add a bot or change a bot's botaddr to use the plain port, once with and once without "-".
 - The logmessages should show STARTTLS being sent by the hub on both occasions, but being ignored when using "-".

### DIFF
--- a/doc/sphinx_source/mainDocs/tcl-commands.rst
+++ b/doc/sphinx_source/mainDocs/tcl-commands.rst
@@ -1605,10 +1605,10 @@ setdccaway <idx> <message>
   Module: core
 
 ^^^^^^^^^^^^^^^^^^^^^^^^
-connect <host> <[+]port>
+connect <host> <[+|-]port>
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-  Description: makes an outgoing connection attempt and creates a dcc entry for it. A 'control' command should be used immediately after a successful 'connect' so no input is lost. If the port is prefixed with a plus sign, SSL encrypted connection will be attempted.
+  Description: makes an outgoing connection attempt and creates a dcc entry for it. A 'control' command should be used immediately after a successful 'connect' so no input is lost. If the port is prefixed with a plus sign, SSL encrypted connection will be attempted, when prefixed with a minus sign, secure connections are disallowed.
 
   Returns: idx of the new connection
 
@@ -1618,7 +1618,7 @@ connect <host> <[+]port>
 listen <port> <type> [options] [flag]
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-  Description: opens a listening port to accept incoming telnets; type must be one of "bots", "all", "users", "script", or "off". Prefixing the port with a plus sign will make eggdrop accept SSL connections on it.
+  Description: opens a listening port to accept incoming telnets; type must be one of "bots", "all", "users", "script", or "off". Prefixing the port with a plus sign will make eggdrop accept SSL connections on it, prefixing with a minus sign will disallow secure connections.
 
     listen <port> bots [mask]
 
--- a/src/cmds.c
+++ b/src/cmds.c
@@ -766,17 +766,18 @@ static void cmd_resetconsole(struct userrec *u, int idx, char *par)
   do_console(u, idx, par, 1);
 }
 
-/* Check if a string is a valid integer and lies non-inclusive
- * between two given integers. Returns 1 if true, 0 if not.
+/* Check if a string is a valid port between 0 and 65536 non-inclusive.
+ * Negative portnumbers are also valid.
+ * Returns 0 if not valid or the absolute port value if valid.
  */
-int check_int_range(char *value, int min, int max) {
+int strtoport(char *value) {
   char *endptr = NULL;
-  long intvalue;
+  int intvalue;
 
   if (value && value[0]) {
-    intvalue = strtol(value, &endptr, 10);
-    if ((intvalue < max) && (intvalue > min) && (*endptr == '\0')) {
-      return 1;
+    intvalue = abs((int) strtol(value, &endptr, 10));
+    if ((intvalue < 65536) && (intvalue > 0) && (*endptr == '\0')) {
+      return intvalue;
     }
   }
   return 0;
@@ -787,7 +788,7 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
   char *handle, *addr, *port, *port2, *relay, *host;
   struct userrec *u1;
   struct bot_addr *bi;
-  int i, found = 0;
+  int i, tport = 3333, rport = 3333, found = 0;
 
   if (!par[0]) {
     dprintf(idx, "Usage: +bot <handle> [address [telnet-port[/relay-port]]] "
@@ -862,13 +863,13 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
   }
 #endif
   if (port) {
-    if (!check_int_range(port, 0, 65536) && !check_int_range(port, -65536, 0)) {
+    if (!(tport = strtoport(port))) {
       dprintf(idx, "Ports must be integers between 1 and 65535.\n");
       return;
     }
   }
   if (relay) {
-    if (!check_int_range(relay, 0, 65536) && !check_int_range(relay, -65536, 0)) {
+    if (!(rport = strtoport(relay))) {
       dprintf(idx, "Ports must be integers between 1 and 65535.\n");
       return;
     }
@@ -892,18 +893,16 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
   strcpy(bi->address, addr);
 
   if (!port) {
-    bi->telnet_port = 3333;
-    bi->relay_port = 3333;
+    bi->telnet_port = tport;
+    bi->relay_port = rport;
   } else {
 #ifdef TLS
     if (*port == '+')
       bi->ssl |= TLS_BOT;
     else if (*port == '-')
       bi->ssl |= TLS_BOT_REJ;
-    bi->telnet_port = abs(atoi(port));
-#else
-    bi->telnet_port = atoi(port);
 #endif
+    bi->telnet_port = tport;
     if (!relay) {
       bi->relay_port = bi->telnet_port;
 #ifdef TLS
@@ -915,10 +914,8 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
         bi->ssl |= TLS_RELAY;
       else if (*relay == '-')
         bi->ssl |= TLS_RELAY_REJ;
-      bi->relay_port = abs(atoi(relay));
-#else
-      bi->relay_port = atoi(relay);
 #endif
+      bi->relay_port = rport;
     }
   }
 
@@ -1211,13 +1208,13 @@ static void cmd_chaddr(struct userrec *u, int idx, char *par)
   }
 #endif
   if (port && port[0]) {
-    if (!check_int_range(port, 0, 65536) && !check_int_range(port, -65536, 0)) {
+    if (!(telnet_port = strtoport(port))) {
       dprintf(idx, "Ports must be integers between 1 and 65535.\n");
       return;
     }
   }
   if (relay) {
-    if (!check_int_range(relay, 0, 65536) && !check_int_range(relay, -65536, 0)) {
+    if (!(relay_port = strtoport(relay))) {
       dprintf(idx, "Ports must be integers between 1 and 65535.\n");
       return;
     }
@@ -1264,7 +1261,7 @@ static void cmd_chaddr(struct userrec *u, int idx, char *par)
       bi->ssl |= TLS_BOT;
     else if (*port == '-')
       bi->ssl |= TLS_BOT_REJ;
-    bi->telnet_port = abs(atoi(port));
+    bi->telnet_port = telnet_port;
     if (!relay) {
       bi->relay_port = bi->telnet_port;
       bi->ssl |= (*port == '+') ? TLS_RELAY : ((*port == '-') ? TLS_RELAY_REJ : 0);
@@ -1274,14 +1271,14 @@ static void cmd_chaddr(struct userrec *u, int idx, char *par)
       } else if (*relay == '-') {
         bi->ssl |= TLS_RELAY_REJ;
       }
-      bi->relay_port = abs(atoi(relay));
+      bi->relay_port = relay_port;
 #else
   } else {
-    bi->telnet_port = atoi(port);
+    bi->telnet_port = telnet_port;
     if (!relay) {
       bi->relay_port = bi->telnet_port;
     } else {
-      bi->relay_port = atoi(relay);
+      bi->relay_port = relay_port;
 #endif
     }
   }
--- a/src/cmds.c
+++ b/src/cmds.c
@@ -854,20 +854,21 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
   }
 
 #ifndef TLS
-  if ((port && *port == '+') || (relay && relay[0] == '+')) {
-    dprintf(idx, "Ports prefixed with '+' are not enabled "
+  if ((port && (*port == '+' || *port == '-')) ||
+      (relay && (*relay == '+' || *relay == '-'))) {
+    dprintf(idx, "Ports prefixed with '+' or '-' are not enabled "
       "(this Eggdrop was compiled without TLS support).\n");
     return;
   }
 #endif
   if (port) {
-    if (!check_int_range(port, 0, 65536)) {
+    if (!check_int_range(port, 0, 65536) && !check_int_range(port, -65536, 0)) {
       dprintf(idx, "Ports must be integers between 1 and 65535.\n");
       return;
     }
   }
   if (relay) {
-    if (!check_int_range(relay, 0, 65536)) {
+    if (!check_int_range(relay, 0, 65536) && !check_int_range(relay, -65536, 0)) {
       dprintf(idx, "Ports must be integers between 1 and 65535.\n");
       return;
     }
@@ -897,19 +898,27 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
 #ifdef TLS
     if (*port == '+')
       bi->ssl |= TLS_BOT;
-#endif
+    else if (*port == '-')
+      bi->ssl |= TLS_BOT_REJ;
+    bi->telnet_port = abs(atoi(port));
+#else
     bi->telnet_port = atoi(port);
+#endif
     if (!relay) {
       bi->relay_port = bi->telnet_port;
 #ifdef TLS
-      bi->ssl *= TLS_BOT + TLS_RELAY;
+      bi->ssl |= (*port == '+') ? TLS_RELAY : ((*port == '-') ? TLS_RELAY_REJ : 0);
 #endif
     } else  {
 #ifdef TLS
-      if (relay[0] == '+')
+      if (*relay == '+')
         bi->ssl |= TLS_RELAY;
-#endif
+      else if (*relay == '-')
+        bi->ssl |= TLS_RELAY_REJ;
+      bi->relay_port = abs(atoi(relay));
+#else
       bi->relay_port = atoi(relay);
+#endif
     }
   }
 
@@ -920,9 +929,9 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
            relay ? relay : "", host[0] ? " " : "", host);
 #ifdef TLS
     dprintf(idx, "Added bot '%s' with address [%s]:%s%d/%s%d and %s%s%s.\n",
-            handle, addr, (bi->ssl & TLS_BOT) ? "+" : "", bi->telnet_port,
-            (bi->ssl & TLS_RELAY) ? "+" : "", bi->relay_port, host[0] ?
-            "hostmask '" : "no hostmask", host[0] ? host : "",
+            handle, addr, (bi->ssl & TLS_BOT) ? "+" : ((bi->ssl & TLS_BOT_REJ) ? "-" : ""),
+            bi->telnet_port, (bi->ssl & TLS_RELAY) ? "+" : ((bi->ssl & TLS_RELAY_REJ) ? "-" : ""),
+            bi->relay_port, host[0] ? "hostmask '" : "no hostmask", host[0] ? host : "",
             host[0] ? "'" : "");
 #else
     dprintf(idx, "Added bot '%s' with address [%s]:%d/%d and %s%s%s.\n", handle,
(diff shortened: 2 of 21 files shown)


doc/sphinx_source/mainDocs/tcl-commands.rst
```diff
@@ -1616,7 +1616,7 @@ connect <host> <[+]port>
 listen <port> <type> [options] [flag]
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-  Description: opens a listening port to accept incoming telnets; type must be one of "bots", "all", "users", "script", or "off". Prefixing the port with a plus sign will make eggdrop accept SSL connections on it.
+  Description: opens a listening port to accept incoming telnets; type must be one of "bots", "all", "users", "script", or "off". Prefixing the port with a plus sign will make eggdrop accept SSL connections on it, prefixing with a minus sign will disallow secure connections.
```
vanosg, line 1621: Is there ever a case where we want users to be able to disallow using secure connection? I appreciate the thoroughness of the change, but is modifying the listen command with the same syntax perhaps too far?

key2peace: uhm, yes, imagine localhost connections on your own dedicated box where secure localhost connections are just a waste of cpu cycles..
Also I would be in favour of explicitly using +&lt;port&gt; which would emply enforcing it (non-tls rejected), -&lt;port&gt; when explicitly left off like the patch intends and plain &lt;port&gt; for any variation allowed

vanosg: Hi key2peace, thanks for the comment. This was not directed at the general concept of disallowing TLS, rather the specific case of with listen. Separate from the listen, botllinks can disable TLS using a - prefix in the .botaddr entry, which was the original intent of this PR, and would still allow you to disable a secure link on a per-bot basis.

Cizzle: When you enfore no TLS on the listening port of a hub, there's no starttls initiated by the hubbot and thus a little bit less noise (and cpu cycles both on the hub and the slaves).

src/cmds.c
```diff
@@ -853,20 +853,21 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
+  if ((port && (*port == '+' || *port == '-')) ||
+      (relay && (*relay == '+' || *relay == '-'))) {
+    dprintf(idx, "Ports prefixed with '+' or '-' are not enabled "
       "(this Eggdrop was compiled without TLS support).\n");
     return;
   }
 #endif
   if (port) {
-    if (!check_int_range(port, 0, 65536)) {
+    if (!check_int_range(port, 0, 65536) && !check_int_range(port, -65536, 0)) {
```
thommey: maybe just check_int_range(port, -65536, 65536)?

src/cmds.c
```diff
@@ -853,20 +853,21 @@ static void cmd_pls_bot(struct userrec *u, int idx, char *par)
...
       dprintf(idx, "Ports must be integers between 1 and 65535.\n");
       return;
     }
   }
   if (relay) {
-    if (!check_int_range(relay, 0, 65536)) {
+    if (!check_int_range(relay, 0, 65536) && !check_int_range(relay, -65536, 0)) {
```
thommey: maybe just check_int_range(port, -65536, 65536)?

src/mod/filesys.mod/filesys.c
```diff
@@ -805,8 +805,8 @@ static void filesys_dcc_send_hostresolved(int i)
       open_telnet_raw(dcc[i].sock, &dcc[i].sockname) < 0)
         dcc[i].type->eof(i);
 #ifdef TLS
-      else if (dcc[i].ssl && ssl_handshake(dcc[i].sock, TLS_CONNECT, tls_vfydcc,
-                                           LOG_FILES, dcc[i].host, NULL))
+      else if (dcc[i].ssl & DCC_TLS_USE && ssl_handshake(dcc[i].sock,
```
thommey: paranthesis around (x & y) && z are probably recommended by the compiler as a warning :)

src/mod/transfer.mod/transfer.c
```diff
@@ -875,7 +875,7 @@ static void dcc_get_pending(int idx, char *buf, int len)
   i = answer(dcc[idx].sock, &dcc[idx].sockname, &port, 1);
   killsock(dcc[idx].sock);
 #ifdef TLS
-  if (dcc[idx].ssl && ssl_handshake(i, TLS_LISTEN, tls_vfydcc,
+  if (dcc[idx].ssl & DCC_TLS_USE && ssl_handshake(i, TLS_LISTEN, tls_vfydcc,
```
thommey: (x & y) && z probably recommended

thommey (commented):
Is it correct that DCC_SSL_USE doubles now as indicator of '+' and that ssl is currently being used? the fingerprint change seems to indicate that you can't set a fingerprint on an optimistic ssl connection now. I might be wrong about that, is it true?

src/cmds.c
Cizzle: Still have to check for zero though.

src/mod/filesys.mod/filesys.c
Cizzle: No warnings here but I'll change it for you.

Cizzle:
The DCC_TLS_USE indicates TLS is used on a connection and will cause a '+' to show up where display of a portnumber is present. This is nothing new, I've only given it a name. This is seperate from the '+' being used actively to indicate a TLS enforced listening port (on either slave or hub), for these the TLS_BOT* and TLS_RELAY* defines are used.
Regarding fingerprints I can't answer your question, it's also not clear to me what either "true" or "false" would mean for this feature here.

src/cmds.c
key2peace: if (!check_int_range(abs(port), 0, 65536)) {

might be offtopic to this issue but:
check_int_range should be using min and max values correctly, it uses the values as border (which is syntax technically wrong), so I'd suggest to change https://github.com/eggheads/eggdrop/blob/develop/src/cmds.c#L778 to:
 if ((intvalue <= max) && (intvalue >= min) && (*endptr == '\0')) {

the new version would be (solving both your comments):
if (!check_int_range(abs(port), 1, 65535)) {

while at it, apply this also to:
src/cmds.c:870:    if (!check_int_range(relay, 0, 65536)) {
src/cmds.c:1204:    if (!check_int_range(port, 0, 65536)) {
src/cmds.c:1210:    if (!check_int_range(relay, 0, 65536)) {
src/tcluser.c:386:    if (q && !check_int_range(q, 0, 65536)) {
src/tcluser.c:390:    if (p && !check_int_range(p, 0, 65536)) {

src/cmds.c
key2peace: no, that would include 0 which is invalid

src/cmds.c
Cizzle: Not a bad idea, on both accounts. However `port` is a char* on which you can't call `abs()`.

key2peace: ok, here we go:
https://github.com/eggheads/eggdrop/blob/489b8405b321d63c8351ee641fd52c5cd23fdec7/src/cmds.c#L769-L783

```
    /* Check if a string is a valid integer between
     * two given integers. Returns 1 if true, 0 if not.
     */
    int check_int_range(char *value, int min, int max) {
      char *endptr = NULL;
      long intvalue;

      if (value && value[0]) {
        intvalue = abs(strtol(value, &endptr, 10));
        if ((intvalue =< max) && (intvalue >= min) && (*endptr == '\0')) {
          return 1;
        }
      }
      return 0;
    }
```

https://github.com/eggheads/eggdrop/blob/489b8405b321d63c8351ee641fd52c5cd23fdec7/src/cmds.c#L864,1204

`    if (!check_int_range(port, 1, 65535)) {`

https://github.com/eggheads/eggdrop/blob/489b8405b321d63c8351ee641fd52c5cd23fdec7/src/cmds.c#L870,1209

` if (!check_int_range(relay, 1, 65535)) {`

https://github.com/eggheads/eggdrop/blob/489b8405b321d63c8351ee641fd52c5cd23fdec7/src/tcluser.c#L384-393

```
    /* Verify */
    /* check_int_range returns 0 if q or p is NULL */
    if (q && !check_int_range(q, 1, 65535)) {
      Tcl_AppendResult(irp, "0", NULL);
      return TCL_OK;
    }
    if (p && !check_int_range(p, 1, 65535)) {
      Tcl_AppendResult(irp, "0", NULL);
      return TCL_OK;
    }
```

michaelortmann: i would also change that line. but go one step further. check_int_range() is always called with fixed parameters 2 and 3 within eggdrop. we dont need to keep it for compatibility, its not exportet. i propose, to rename it to something like strtoport() wich can return error if not in range and the port as an int if successful. the func only needs one param, the char*port. the rest of the code can use the returned int, and doesnt have to fight with strtol or atoi again later.

ok, not rly part of the PR, but hey, its out of my head now.

Cizzle: @key2peace Good luck reusing that function to check if -9 lies between -100 and 0.
@michaelortmann I wouldn't disapprove but as you mention: not the point of this PR.

key2peace: @cizzle it is only used to check port range, the usage is the same everywhere
@michaelortmann rename it to isvalidport(port) and rip out min/max as it should be between 1 and 65535 anyways

vanosg:
Bot A = 1.8.3
Bot B = 1.8.3+negssl
Linking Bot A to Bot B

Connecting from a 1.8.3 bot to this patch, using .+bot ip 3184  (no + or -)
```
B:
5   0.0.0.0                                  + 3184 (telnet)  lstn  3184
 
[20:00:51] net: sockread(): SSL_read() SSL_ERROR_SYSCALL
[20:00:51] NET: SSL read failed. Non-SSL connection?
[20:00:51] net: eof!(read) socket 10
[20:00:51] Lost telnet connection to telnet@scoobalooba/50585

A:
[20:00:36] Linking to testbot at 192.168.180.132:3184 ...
[20:00:51] net: attempted socket connection refused: 192.168.180.132
[20:00:51] net: attempted socket connection refused: 192.168.180.132
[20:00:51] net: attempted socket connection refused: 192.168.180.132
*** [goodbot] Couldn't link to testbot.
[20:00:51] Failed link to testbot.
```

```
B:
5   0.0.0.0                                  - 3184 (telnet)  lstn  3184

[20:02:38] Telnet connection: scoobalooba/39297
[20:02:38] net: eof!(read) socket 11
[20:02:38] Timeout/EOF ident connection
[20:02:38] Challenging goodbot...
[20:03:41] net: eof!(read) socket 10
[20:03:41] Lost dcc connection to goodbot (telnet@scoobalooba/39297)
```

Now, connecting to a 1.8.3+negssl bot, from a 1.8.3 bot. (Yes, this is a stupid test, but it still breaks). 
The 1.8.3 bot adds the negssl bot with +3184, -3184, and 3184:
```
A: +3184
[20:10:27] Linking to testbot at 192.168.180.132:3184 ...
[20:10:28] TLS: attempting SSL negotiation...
[20:10:28] TLS: state change: before SSL initialization
[20:10:28] TLS: state change: before SSL initialization
[20:10:28] TLS: state change: SSLv3/TLS write client hello
[20:10:28] TLS: awaiting more writes
[20:10:28] TLS: handshake failed due to the following errors: 
[20:10:28] TLS: error:1408F10B:SSL routines:ssl3_get_record:wrong version number
[20:10:28] net: attempted socket connection refused: 192.168.180.132
[20:10:28] net: attempted socket connection refused: 192.168.180.132
[20:10:28] net: attempted socket connection refused: 192.168.180.132
*** [goodbot] Couldn't link to testbot.
[20:10:28] Failed link to testbot.
```

A: -3184
Can't do it on a 1.8.3

A: 3184
Linked


Now, connecting from negssl to 1.8.3, where negssl uses +/- on +bot:
B: +3183
```
[00:04:17] Linking to goodbot at 192.168.180.132:+3183 ...
[00:04:17] TLS: attempting SSL negotiation...
[00:04:17] TLS: state change: before SSL initialization
[00:04:17] TLS: state change: before SSL initialization
[00:04:17] TLS: state change: SSLv3/TLS write client hello
[00:04:17] TLS: awaiting more writes
[00:04:17] TLS: handshake failed due to the following errors: 
[00:04:17] TLS: error:1408F10B:SSL routines:ssl3_get_record:wrong version number
```

B: 3183 (no +-)
Works

B: -3183
```
[00:10:39] Linking to goodbot at 192.168.180.132:-3183 ...
[00:10:39] Password required for connection to goodbot.
[00:10:39] Linked to goodbot.
```

michaelortmann:
```
Geo> Have bot A listen +3333
Geo> Have bot B add bot A with :3333 (no +)
Geo> have bot B link to bot A, and observe the result
```

**1. eggdrop 1.8.2**

When doing .+bot <host> 3333
eggdrop 1.8.2 differs from 1.8.
eggdrop 1.8.2 adds the bot with +3333
eggdrop 1.8.3 adds the bot with 3333
Thats why the 1.8.2 would link

2. eggdrop 1.8.3

http://docs.eggheads.org/mainDocs/tls.html
says:
By default, eggdrop now automatically attempts to protect botnet links with SSL if it is compiled with TLS support.
this could be read as:
destination port is 3333: try ssl, fallback to non ssl

We observe, when adding a bot with 3333, it wont try ssl.

There is/was no fallback. if +PORT, eggdrop would force SSL only, the desired behaviour, and if PORT (without "+"), eggdrop would only try PLAIN. The bug in eggdrop 1.8.2 covered this fact :/

If you wanna add the "try ssl" also for PORT (without "+") with fallback to PLAIN, then one of the starting points would probably be:

`dcc[i].ssl &&`
in line
https://github.com/eggheads/eggdrop/blob/develop/src/botnet.c#L1098

In a small test of mine, eggdrop 1.8.3 linked ssl to ssl only bot:

BotA:
```
[06:37:54] #-HQ# dccstat
IDX ADDR                                     + PORT NICK      TYPE  INFO
--- ---------------------------------------- ------ --------- ----- ---------
5   0.0.0.0                                  + 3333 (telnet)  lstn  3333
6   192.168.0.4                                   0 (dns)     dns   (ready)
1   0.0.0.0                                       0 -HQ       chat  flags: cptEp/0
10  127.0.0.1                                  6667 (server)  serv  (lag: 0)
11  127.0.0.1                                +41639 BotB      bot   flags: puCosgwliA
```

BotB:
```
.whois BotA
[06:38:37] tcl: builtin dcc call: *dcc:whois -HQ 1 BotA
[06:38:37] #-HQ# whois BotA
HANDLE                           PASS NOTES FLAGS           LAST
BotA                             yes      0 b               06:26 (linked)
  ADDRESS: 127.0.0.1
     users: 3333, bots: 3333
.link BotA
[06:40:37] Linking to BotA at 127.0.0.1:3333 ...
[06:40:38] TLS: attempting SSL negotiation....
[...]
[06:40:38] Linked to BotA.
```

But it couldn't link to a non-ssl bot. I don't have an easy fix. Eggdrop simply has no fallback from ssl to plain yet. That is still open to implement and it will involve some real work.

Maybe after changing the following line as proposed above to starttls in any case:
https://github.com/eggheads/eggdrop/blob/develop/src/botnet.c#L1098
we must modify
dcc.c:failed_link:
https://github.com/eggheads/eggdrop/blob/develop/src/dcc.c#L273
we already fallback to other ports if the main port isn't available

So i propose to try a solution where we check all those ports, but if we check with ssl we also check for nonssl before trying the next port. this doubles the checks

a state less solution:

after line `dcc[idx].timeval = now;` we insert something like:

```
if (dcc[idx].ssl)
  dcc[idx].ssl = 0; /* toggle ssl / plain for next try */
  if (open_telnet(idx, dcc[idx].host, dcc[idx].port) < 0) /* dont increase port here */
    failed_link(idx);
else
  dcc[idx].ssl = 1; /* toggle ssl / plain for next try */
```

then comes the rest of the code in this func, whichn increased the port num and tries again

vanosg:
Since we're hijacking this PR now...

BotA has 'listen +3183' in the config.
BotB adds BotA with ':port' (no +)

Perhaps a clue in this- after linking, BotB is returned an error from BotA:
```[06:06:28] Refused telnet@mymachine (invalid handle: BotB)```

**So to clarify what @michaelortmann said above- it looks like the root problem is that if a bot listens with +, it isn't sending a STARTTLS when receiving a connection.** 


What BotA should be doing is sending a STARTTLS to BotB. It does not. To compare full logs of good vs bad:

Good:
```
[06:11:56] Telnet connection: mymachine/49009
[06:11:56] Timeout/EOF ident connection
[06:11:56] Sent STARTTLS to testbot...
[06:11:56] Challenging testbot...
[06:11:56] TLS: handshake successful. Secure connection established.
[06:11:56] TLS: certificate subject: C=EU, ST=EU, O=SNOOP, OU=Botnet, emailAddress=foo, UID=bot
[06:11:56] TLS: certificate issuer: C=EU, ST=EU, O=SNOOP, OU=Botnet, emailAddress=foo, UID=bot
[06:11:56] TLS: certificate MD5 Fingerprint: 90:A1:F5:7B:86:46:DA:10:8A:8F:3D:BA:ED:20:2D:68
[06:11:56] TLS: certificate SHA1 Fingerprint: DB:D0:24:83:64:AE:03:97:52:21:B7:22:2F:F0:FA:B9:85:4E:D7:02
[06:11:56] TLS: certificate valid from Sep 28 01:03:13 2016 GMT to Sep 28 01:03:13 2017 GMT
[06:11:56] TLS: cipher used: ECDHE-RSA-AES256-GCM-SHA384 TLSv1.2; 256 bits (256 secret)
[06:11:56] Linked to testbot.
```
Bad:
```
[06:11:04] Telnet connection: scoobalooba/42117
[06:11:04] TLS: attempting SSL negotiation...
[06:11:04] TLS: state change: before SSL initialization
[06:11:04] TLS: state change: before SSL initialization
[06:11:04] TLS: awaiting more reads
[06:11:04] TLS: handshake in progress
[06:11:04] TLS: awaiting more reads
[06:11:04] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:04] net: eof!(read) socket 11
[06:11:04] Timeout/EOF ident connection
[06:11:04] TLS: awaiting more reads
[06:11:05] TLS: awaiting more reads
[06:11:05] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:05] DNS Resolver: Cache record for "scoobalooba" (192.168.180.132) has expired. (state: 0) Marked for expire at: 1541225464.
[06:11:05] TLS: awaiting more reads
[06:11:05] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:06] TLS: awaiting more reads
[06:11:06] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:06] TLS: awaiting more reads
[06:11:06] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:07] TLS: awaiting more reads
[06:11:07] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:07] TLS: awaiting more reads
[06:11:07] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:08] TLS: awaiting more reads
[06:11:08] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:08] TLS: awaiting more reads
[06:11:08] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:09] TLS: awaiting more reads
[06:11:09] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:09] TLS: awaiting more reads
[06:11:09] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:10] TLS: awaiting more reads
[06:11:10] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:10] TLS: awaiting more reads
[06:11:10] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:11] TLS: awaiting more reads
[06:11:11] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:11] TLS: awaiting more reads
[06:11:11] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:12] TLS: awaiting more reads
[06:11:12] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:12] TLS: awaiting more reads
[06:11:12] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:13] TLS: awaiting more reads
[06:11:13] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:13] TLS: awaiting more reads
[06:11:13] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:14] TLS: awaiting more reads
[06:11:14] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:14] TLS: awaiting more reads
[06:11:14] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:15] TLS: awaiting more reads
[06:11:15] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:15] TLS: awaiting more reads
[06:11:15] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:16] TLS: awaiting more reads
[06:11:16] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:16] TLS: awaiting more reads
[06:11:16] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:17] TLS: awaiting more reads
[06:11:17] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:17] TLS: awaiting more reads
[06:11:17] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:18] TLS: awaiting more reads
[06:11:18] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:18] TLS: awaiting more reads
[06:11:18] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:19] TLS: awaiting more reads
[06:11:19] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:19] TLS: awaiting more reads
[06:11:19] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:20] TLS: awaiting more reads
[06:11:20] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:20] TLS: awaiting more reads
[06:11:20] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:21] TLS: awaiting more reads
[06:11:21] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:21] TLS: awaiting more reads
[06:11:21] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:22] TLS: awaiting more reads
[06:11:22] sockread EAGAIN: 9 11 (Resource temporarily unavailable)
[06:11:22] TLS: awaiting more reads
[06:11:22] dequeue_sockets(): errno = 11 (Resource temporarily unavailable) on 9
[06:11:23] TLS: awaiting more writes
[06:11:23] net: sockread(): SSL_read() SSL_ERROR_SYSCALL
[06:11:23] NET: SSL read failed. Non-SSL connection?
[06:11:23] net: eof!(read) socket 9
[06:11:23] Lost telnet connection to telnet@scoobalooba/42117
```

You can see these debug message diverge very early on.

vanosg:
So now, my theory as to what is going wrong. 

Doing a listen on +port means accepting user and bot connections. When the bot links, the bot listening doesn't know that it is a bot, and doesn't know it has to send a STARTTLS because, *I think* the handle isn't sent until after the secure connection is established. So, the linking bot knows it should be waiting for a STARTTLS, because its a bot link, but it never comes because the listening bot doesn't know that it is a bot connecting.

...is one theory, until a better one comes along. Will look into it soon

Edit: Yes, confirmed, handle isn't sent until after a secure link is established

michaelortmann:
Are you sure, the listener is the problem or the right place to fix?

See my comment under #754.

You can find the fix and the rest of the story in #755.

Of course #755 fixes #754, not #552.

Because #755 and #552 overlap, i also took a quick look into this #552 patch. It looks compatible. After #755 you would have to apply DCC_TLS_USE not only in https://github.com/eggheads/eggdrop/blob/314d05ada1427a349aab9126441d6fe5156f2330/src/botnet.c#L1106 but also in the then duplicated code in https://github.com/eggheads/eggdrop/blob/6a8b698dd94c6264a60d879bfdd1487e69a267a5/src/dcc.c#L307.

Cizzle:
Closing as this feature will become obsolete in the next 1.9 version. Adding this now for just one version would introduce possible unseen bugs and a need to remain backward compatible.

PR #553, opened 2018-08-16
rewrote replace_spaces(), to be O(n) instead of O(N**2)

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: fat

One-line summary:
replace_spaces() was O(N**2) and memory unfriendly

function replace_spaces() in transfer.c is used only once in transfer.c. so side effects of patching this function are somewhat limited. the function was O(N**2) while using strchr again and again from the start of string instead of the last occurrence. it was allocating, copying, freeing without need. this patch slims eggdrop.

  created file "dccsend test.txt" in eggdrop directory
  bot:
    .tcl dccsend "dccsend test.txt" testuser
  irc client with nick=testuser:
    *** DCC SEND (dccsend_test.txt 12) request received from[...]


Cizzle:
You're wrong about the complexity, the strchr changed the same pointer it searched to the last occurence, which could be improved by a + 1 in the loop.

michaelortmann:
youre right. rest holds.

PR #554, opened 2018-08-17
replace getdtablesize() and FD_SETSIZE with maxfd+1

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: getdtablesize

One-line summary:
unnecessary and unwanted getdtablesize() in mainloop

getdtablesize() is not rly posix 2001 so select() should be used with FD_SETSIZE or (better) a calculated maxfd. we could easily calculate maxfd where we FD_SET() but we don't need to to get the other benefits. less fat in src and less stress in mainloop. getdtablesize() -> prlimit64() was called every second. you can check the prlimit64() call (or a similar call) with strace(). right after that call the result of getdtablesize() was set to FD_SETSIZE  in case FD_SETSIZE was defined and lower.
meanwhile this fix calculates maxfd and uses select(max + 1, ...).

started eggdrop, connected to server

this patch touches eggdrops networking core so i would like to see at least two LGTMs here.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -99,9 +99,6 @@
 /* Define to 1 if you have the `fsync' function. */
 #undef HAVE_FSYNC
 
-/* Define to 1 if you have the `getdtablesize' function. */
-#undef HAVE_GETDTABLESIZE
-
 /* Define to 1 if you have the `gethostbyname2' function. */
 #undef HAVE_GETHOSTBYNAME2
 
--- a/configure
+++ b/configure
@@ -6919,7 +6919,7 @@ ac_config_commands="$ac_config_commands $ac_stdint_h"
 
 
 # Checks for functions and their arguments.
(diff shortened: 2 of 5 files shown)


Cizzle:
LGTM

vanosg:
build

thommey:
The "nfds" passed to select should be the maximum file descriptor we're interested in + 1.
Passing any maximum value of the setsize just means that select() will have to check for all of those if they are in rfds/wfds/efds. Probably negligible performance penalty these days, but still not what's intended (man 2 select on linux describes nfds as maxfd_in_set+1 as well).

So, IMHO, it's unnecessary to figure out any maximum, including FD_SETSIZE, because we're screwed if we get an FD that high anyway, because we cannot monitor it with select() and cannot recover from that. We could just gather the max fd and +1 it for the first select() argument.

michaelortmann:
@thommey: done

thommey (approved):
LGTM

vanosg:
build

Issue #555, opened 2018-08-17
DCCLIST TELNET: SSL port


ghost:
\#\# dcclist telnet
<&Samantha> [TCL (0ms)] {10 (telnet) * TELNET {lstn  1111} 1534144391} {11 (bots) * TELNET {lstn  1112} 1534144391}

Like #538, DCCLIST TELNET doesn't indicate SSL port via "+"

ghost:
I have a patch to offer to fix this.

ghost:
Sorry for all the edit messages in your channel. I have to manually ESCAPE every damned line on github ....

I can't find an awake coder, and, I don't do "pull requests / push request /copy entire damned repositories" because ... I just don't know how and I don't care to learn how for just one file.

So, I'll put it here.

$ diff -EBaNU0 old new

diff -EBaNU0 old/dcc.c new/dcc.c
--- old/dcc.c   2018-08-17 21:25:35.374389214 -0700
+++ new/dcc.c   2018-08-17 21:25:01.070578553 -0700
@@ -1436,2 +1436,10 @@
\-  sprintf(buf, "lstn  %d%s", dcc[idx].port,
\-          (dcc[idx].status & LSTN_PUBLIC) ? " pub" : "");
\+       /* 2018-08-17 21:13:00 -0700: Domino */
\+#ifdef TLS
\+  sprintf(buf, "lstn %s%d%s",
\+    dcc[idx].ssl ? "+" : "",
\+    dcc[idx].port,
\+    (dcc[idx].status & LSTN_PUBLIC) ? " pub" : "");
\+#else
\+  sprintf(buf, "lstn %d%s", dcc[idx].port,
\+    (dcc[idx].status & LSTN_PUBLIC) ? " pub" : "");
\+#endif

First "real" patch (this doesn't count the $::configure-args and seen.c patches, which were cosmetic). Don't get too mad at me.

ghost:
\<\~Domino> pep tcl foreach a [dcclist] { putquick "PRIVMSG #channel :$a " }
\<%Pepsi> 5 (telnet) * TELNET {lstn 16516} 1534567259 
\<%Pepsi> 6 (bots) * TELNET {lstn +16517} 1534567259 
\<%Pepsi> 7 (dns) {} DNS {dns   (ready)} 1534567259 
\<%Pepsi> 9 Hub-Bette \<IP> BOT {bot   flags: pucosgwliA} 1534567261 
[...]

PR #564, opened 2018-08-18
Add + when using TLS for TELNET in Tcl dcclist.

Cizzle:
Found by: maimizuno
Patch by: maimizuno
Fixes: #555 

One-line summary: Adds a + when using a TLS TELNET listen port in the Tcl `dcclist` output.

Additional description (if needed): The `.dccstat` output is slightly altered aswell, the INFO column will now show this + as well.

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -1433,7 +1433,7 @@ static void eof_dcc_telnet(int idx)
 
 static void display_telnet(int idx, char *buf)
 {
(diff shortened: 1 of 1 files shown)


Cizzle:
After merge, #552 should be updated for the affected line as well.

Cizzle:
Changing this to 1.9.0 as the added "+" might be unexpected and break parsers.

thommey:
I'd consider this for 1.8.4 in conjunction with adding '-' from #552

PR #566, opened 2018-08-21
fix #354 and fix #387 (fix order of binds)

### LINKED ISSUE #354: bind invite gets the hostmask on every other triggering
Full explanation already written out here:   http://forum.egghelp.org/viewtopic.php?p=105906#105906
See the short little demo script there.

### LINKED ISSUE #387: depending on call sequence, newsplit() can break bind strings
From #365 
> Further testing reveals two things:
 (1) mode bind for - modes triggers BEFORE raw mode binds (whereas + binds trigger AFTER the raw mode bind)
 (2) there's a SECOND 0x00 in the - mode raw: "nick(0x00)ident@host(0x00)#channel (+/-)mode target"

Found by @maimizuno

Under some circumstances it can happen that bind mode trigger before bind raw - in which case bind mode uses newsplit() on the user and thereby destroys the string for the raw bind call. we should at least rule it out in that case. It seemed to depend on whether it's -o or +o

michaelortmann:
Found by: jack3-again and maimizuno
Patch by: michaelortmann
Fixes: #354 and #387

One-line summary:
fix order of binds, which fixes modification/substitution/truncation of args between binds.

sometimes the raw bind is executed before the builtin invite. sometimes the other way around.
the builtin invite modifies the args in chan.c gotinvite() in the following 2 lines:
newsplit(&msg);
nick = splitnick(&from);
now i can confirm its the same as issue #387, but its not only newsplit() but also splitnick() and any other function modifying the parameters of the builtin functions. btw: we are able to reproduce the problem now again. see test cases below. the sequence "toggles" every time.
first i prepared a fix to copy "msg" and "from" before those modifying function calls.
but there were more builtins like this.
then it turned out its better to fix the order of binds.
the mysterious "toggle" was found. it is probably an optimization, moving last recently used binds to the front of the bind table.
this leads to toggling the order of execution every time the same binds are triggered.
now i pushed a new solution which keeps the builtin functions as is, keeps the optimization, but fixes the optimization, so that its only done, when there is exactly 1 bind for the trigger.
no more toggle, no more truncation/substitution and a happy maimizuno.

testscript (modify #foo as you like):

bind raw - invite fix354
proc fix354 args {
    putquick "PRIVMSG #foo :raw invite triggered: [join $args ", "]" 
}
bind raw - mode fix387
proc fix387 args {
    putquick "PRIVMSG #foo :raw mode triggered: [join $args ", "]"
}

/invite botnick chan
/invite botnick chan
before: the 2nd invite leads to reversed bind order and args truncation, as shown in #354 
after: its fine now

/mode #foo +v botnick
/mode #foo +o botnick
before: the end mode change leads reversed bind order and args truncation, as shown in #387
after: its fine now

### DIFF
--- a/doc/sphinx_source/mainDocs/tcl-commands.rst
+++ b/doc/sphinx_source/mainDocs/tcl-commands.rst
@@ -2683,7 +2683,9 @@ as 'bind msg - stop msg:stop' (which makes a msg-command "stop" call the
 Tcl proc "msg:stop") will overwrite any previous binding you had for the
 msg command "stop". With stackable bindings, like 'msgm' for example,
 you can bind the same command to multiple procs. When the bind is triggered,
-ALL of the Tcl procs that are bound to it will be called.
+ALL of the Tcl procs that are bound to it will be called. Raw binds are
+triggered before builtin binds, as a builtin bind has the potential to
+modify args.
 
 To remove a bind, use the 'unbind' command. For example, to remove the
 bind for the "stop" msg command, use 'unbind msg - stop msg:stop'.
--- a/src/mod/irc.mod/chan.c
+++ b/src/mod/irc.mod/chan.c
@@ -28,7 +28,7 @@
 static time_t last_ctcp = (time_t) 0L;
 static int count_ctcp = 0;
 static time_t last_invtime = (time_t) 0L;
-static char last_invchan[81] = "";
+static char last_invchan[300] = "";
 
 /* ID length for !channels.
  */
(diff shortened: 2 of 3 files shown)


michaelortmann:
Its ready for test/review/approval now.

thommey:
LGTM

ghost:
https://github.com/eggheads/eggdrop/pull/566/commits/2c7d1456ba0d30aaabb19b43c60b5f60c4644182

   * reason 1: order should be raw than builtin

Should be "then" (one then the other)

Issue #201, opened 2016-03-08
Add debug log that shows executed binds and time it took


thommey:
Sometimes bots just ping timeout because a Tcl script is blocking, this feature request would add a details debug log showing all triggered binds (their procs) and how long the execution took.
Found by: saapete

michaelortmann:
pls come up/step forward with a concept answering 3 questions:

1st: where exactly we should log this

2nd: how to enable/disable that logging
debugx() ?
putlog() ?
console flags ?
additional debug log file?
(new) eggdrop.conf setting(s)?

3rd: what to log
only external tcl procs?
also builtin procs?

PR #570, opened 2018-08-22
Add debug log that shows executed binds/procs and time it took

michaelortmann:
Found by: saapete
Patch by: michaelortmann
Fixes: #201

One-line summary:
Add debug log that shows executed binds/procs and cpu time (user and sys) it took

open for discussion, see our comments below

```
$ ./eggdrop -nt
[...]
.console +d
[...]
[22:46:25] triggering bind evnt:init_server
[22:46:25] triggered bind evnt:init_server, user 0.075ms sys 0.038ms
[...]
.gf lilith
[22:48:53] triggering bind dcc_setuserinfo
Now: lilith
[22:48:53] #-HQ# gf lilith
[22:48:53] triggered bind dcc_setuserinfo, user 0.050ms sys 0.046ms
```


thommey (changes requested):
I think that was before we discussed C89. Variable declarations must be moved up (the timevals).
Can we use clock_gettime(CLOCK_MONOTONIC) instead so this isn't affected by clock jumps? Maybe only count CPU time? Not sure, open for suggestions.

michaelortmann:
i moved the following questions a had over from #201:

pls come up/step forward with a concept answering 4 questions:

1st: where exactly we should log this
debugx() ?
putlog() ?
additional debug log file?

2nd: how to enable/disable that logging
console flags ?
(new) eggdrop.conf setting(s) ?
currently there is no option to enable/disable this performance logging, and it is output for console +d

3rd: what to log
only external tcl procs ?
also builtin procs ?

4th: real time or cpu time ?

michaelortmann:
**@thommey:** your requested changes are committed

src/tclhash.c
```diff
@@ -736,8 +738,18 @@ static inline int trigger_bind(const char *proc, const char *param,
    */
   Tcl_SetVar(interp, "lastbind", (char *) mask, TCL_GLOBAL_ONLY);
 
+  if(proc) {
+    debug1("triggered bind %s start", proc);
```
thommey: I would suggest "triggering bind %s"

src/tclhash.c
```diff
@@ -736,8 +738,18 @@ static inline int trigger_bind(const char *proc, const char *param,
...
+    r = clock_gettime(CLOCK_MONOTONIC, &tp1);
+  }
   x = Tcl_VarEval(interp, proc, param, NULL);
   Context;
+  if (proc && !r) {
+    clock_gettime(CLOCK_MONOTONIC, &tp2);
+    debug2("triggered bind %s end time = %.3fms", proc,
```
thommey: I would suggest "triggered bind %s, duration = %.3fms"

src/tclhash.c
```diff
@@ -736,8 +738,18 @@ static inline int trigger_bind(const char *proc, const char *param,
    */
   Tcl_SetVar(interp, "lastbind", (char *) mask, TCL_GLOBAL_ONLY);
 
+  if(proc) {
+    debug1("triggered bind %s start", proc);
+    r = clock_gettime(CLOCK_MONOTONIC, &tp1);
```
thommey: CPU time is better

thommey (commented):
+d debug logging should be ok

michaelortmann:
@thommey:

i changed the wording as requested

but i don't think we can/should use clock_gettime() with CLOCK_*_CPUTIME_ID

because despite clock_gettime() being posix 2001, CLOCK_*_CPUTIME_ID is not really, some operating systems don't have this (minix) and others implemented it first in year 2013 (openbsd).

i don't feel good using it and i don't have a replacement to offer for those systems lacking it.

besides its unclear from the text in issue #201, what the opener of the issue expects here, could expect CPU time, but could also expect other time.

thommey:
The issue was mostly raised about CPU time, because we have no way of accounting for resource usage of particular scripts. How about getrusage()?

michaelortmann:
sure clock() getrusage()... i throw them over board before, but now theire in again :)

are we even sure, we want to log all binds? or only log tcl procs?

thommey:
Interesting point. I think actual Tcl procs (those not starting with '*') would be enough for now

thommey:
The original report was filed, because it is difficult or impossible to track down Eggdrops using a lot of CPU time by script. Tcl retains no association between "script file" and procs, there is no real concept of "script file" (some [info] commands apparently retain it but not on the level we need and not backwards compatible).

We can exclude internal binds (all procs starting with '*') for now.

Logging per call seems fine for now, instead of aggregating and complicated statistics tracking, but that would certainly be an option for the future. getrusage() sounds appropriate to me.

michaelortmann:
@thommey: its done, take a look.

michaelortmann:
ready for review again.

Cizzle:
LGTM

vanosg:
build

thommey (approved):

Issue #521, opened 2018-05-24
Feature request: botnet throughput console


ghost:
.

Cizzle:
I believe the "m" stands for "me", so you have "me" sending to target/receiving from source. I'm fine by changing it to "b" which'd still be logical then as "bot" for the ghuv flags; unless someone opposes...?

ghost:
Issue acknowledged by coders; issue no longer needed.

ghost:
Cizzle commented:
>I believe the "m" stands for "me", so you have "me" sending to target/receiving from source

Makes sense. Problem is: it's ambiguous with IRC mode queue output; the marker is the same ("[m->]"). The botnet data stream header should be (IMnshO) different, in case you're trying to read through your console output to debug a problem, or, using a script to scan through a log file which may have both. Plus, isn't it better to favor the more-intuitive "understood-at-a-glance" method?

PR #574, opened 2018-08-22
Change output for logmodes ghtu to show b instead of m

Cizzle:
Found by: maimizuno
Patch by: Cizzle
Fixes: #521 

One-line summary: Output for logmodes ghtu (botnet in and out and botnet share in and out) will now use a "b" instead of an "m".

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -350,9 +350,9 @@ static void dcc_bot_new(int idx, char *buf, int x)
 
   if (raw_log) {
     if (!strncmp(buf, "s ", 2))
-      putlog(LOG_BOTSHRIN, "*", "{m<-%s} %s", dcc[idx].nick, buf + 2);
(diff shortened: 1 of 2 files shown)


vanosg:
As trivial as it is, should this be 1.9? Particularly as the user referenced scripts doing log parsing... could this break something?  I'm being oversensitive, but want someone to tell me it's not a big deal.

thommey (approved):

thommey:
Intercepting those should be done by tracing/overwriting putbot and similar instead of bind log on our raw log. I would say the rawlog format is not part of our API, that's a bit far fetched. 1.8.4 is fine

vanosg:
build

vanosg:
I was more focused on parsing text file logs, after the fact. Agree its not API, just wondering if it makes sense to wait for 1.9 to mess with stuff like this. Again, I'm being oversensitive.

vanosg:
If it breaks your scripts, blame thommey. :) Apparently change = progress. Merging!

PR #575, opened 2018-08-22
get rid of less useful random() in mainloop

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: FIXME

One-line summary:
fixes a FIXME, which is true, its a waste of cpu, doesnt help much, esp. now that we have a better seed in place and are sure this random() is used for trivia instead of crypto.

./eggdrop -ntm
[...]
.tcl rand 1000
Tcl: 398

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -816,15 +816,6 @@ int mainloop(int toplevel)
    * calls to periodic_timers
    */
   now = time(NULL);
-  /*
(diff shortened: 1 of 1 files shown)


thommey:
LGTM

thommey (approved):
LGTM

PR #581, opened 2018-08-26
Fix small issues found while compiling on sunos

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: small issues found while compiling on sunos

One-line summary:
silence some sunos compiler warnings and clean up code

see commit messages below

small test, start eggdrop, pid is displayed fine here:
./eggdrop
[...]
Launched into the background  (pid: 18138)

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -608,7 +608,7 @@ static int dcc_bot_check_digest(int idx, char *remote_digest)
 
   MD5_Init(&md5context);
 
-  egg_snprintf(digest_string, 33, "<%x%x@", getpid(),
+  egg_snprintf(digest_string, 33, "<%lx%x@", (long) getpid(),
                (unsigned int) dcc[idx].timeval);
   MD5_Update(&md5context, (unsigned char *) digest_string,
              strlen(digest_string));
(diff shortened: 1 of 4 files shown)


src/dcc.c
```diff
@@ -608,7 +608,7 @@ static int dcc_bot_check_digest(int idx, char *remote_digest)
 
   MD5_Init(&md5context);
 
-  egg_snprintf(digest_string, 33, "<%x%x@", getpid(),
+  egg_snprintf(digest_string, 33, "<%lx%x@", (long) getpid(),
```
thommey, line 611: %lx would correspond to unsigned long, right?

thommey (changes requested):

michaelortmann:
there is no such sing as "signed hex", so i use %lx for signed long as well as unsigned long. i guess every compiler will do the right thing. if i would cast (unsigned long) instead of (long) the code could be wrong, if the long is negative. which of course, wont ever happen with getpid(). maybe, it could happen, if microsoft invented their own getpid_s() ;)

so after reviewing my code, i still guess its correct.

but as always, if i error, if i didnt understand the problem, explain again pls.

thommey:
Ah, right. pid_t is signed, forgot about that, all good.

thommey (approved):

Issue #580, opened 2018-08-26
Botnet linking with TLS (STARTTLS)


ghost:
From the leaf's point of view:

<Domino> .link MyHub
<Leaf> [17:34:19] #Domino# link MyHub
<Leaf> [17:34:19] Linking to MyHub at <Hub's IP>:9348 ...
<Leaf> [17:34:19] Got STARTTLS from MyHub. Replying...
<Leaf> [17:34:19] Received challenge from MyHub... sending response ...
... and nothing more happens.

On the target (MyHub)
<MyHub> [02:38:00] Telnet connection: <Leaf's IP>/39848
<MyHub> [02:38:00] Timeout/EOF ident connection
<MyHub> [02:38:00] Sent STARTTLS to Leaf...
<MyHub> [02:38:00] Challenging Leaf...
<MyHub> [02:35:02] TLS: failed in: SSLv2/v3 read client hello A.
<MyHub> [02:35:02] TLS: failed in: SSLv2/v3 read client hello A.
<MyHub> [02:35:02] TLS: failed in: SSLv2/v3 read client hello A.
<MyHub> [02:35:02] TLS: failed in: SSLv2/v3 read client hello A.
... this repeats until the leaf gives-up (timeout), APPROXIMATELY 40 TIMES !!
<MyHub> [02:39:01] Lost dcc connection to Leaf (telnet@<Leaf's IP>/39848)

Once this begins, I'm helpless to STOP it!

.DCCSTAT on the leaf provides:

<Leaf> 12 <MyHub's IP> + 9348 MyHub bot* waited 39s

<Domino> .tcl killdcc 12
<Leaf> Tcl error: invalid idx

Ummmm.... That's useless (and a lie)

Let's try on the MyHub!
<Domino> .dccstat
<Snip>
<MyHub> 25 <Leaf's IP> +38008 Leaf pass waited 25s
<MyHub> [02:41:26] TLS: failed in: SSLv2/v3 read client hello A.
<MyHub> [02:41:27] TLS: failed in: SSLv2/v3 read client hello A.
<MyHub> [02:41:28] TLS: failed in: SSLv2/v3 read client hello A.
<MyHub> [02:41:29] TLS: failed in: SSLv2/v3 read client hello A.
<Domino> .tcl killdcc 25
<MyHub> [02:41:30] TLS: failed in: SSLv2/v3 read client hello A.
<MyHub> Tcl error: invalid idx
[...]

PR #582, opened 2018-08-26
make 2 code pathes/errors look different, _may_ also help debug #580

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: problem figuring out which code path was taken

One-line summary:
without this patch 2 different code pathes produce same error message, so one cant tell from the log which code path was taken.

the code comment says the code was copied from man page example
but the (current) example of ssl_set_info_callback(3) has different error messages

i did only a small compile/start test, i didnt check tls dunctionality at all.
but this patch only changes log text, what could probably go wrong.


Cizzle:
LGTM

Issue #155, opened 2016-01-29
Rework dcclist


thommey:
The current dcclist output is completely unparsable and lacks information (please someone add details about what's missing).
We need to rework it or introduce a new command and deprecate dcclist in the manually.
Missing: at least SSL info
Found by: jack3

ghost:
Was this fixed by #555 ?

Cizzle:
If you can't come up with things that are missing, then this is fixed.

ghost:
He was saying that
(1) it's not parsable (how isn't it?)
(2) other information was missing (no specific items suggested)
(3) I patched the SSL issue already.

Looks like it was a third-party comment anyway to begin with. I'm not sure what can't be parsed out (the type is rather dynamic, yes but ...) but it's a proper list, and the data positions don't change.

Now, I'm just curious.

Cizzle:
Looking at it closer I might say that these things are missing;
 - SERVER-types don't have the port (and TLS) info
 - if an IP doesn't resolve to a hostname, it's an empty element; I feel the IP could be used then
Anyone concur?

ghost:
> SERVER-types don't have the port (and TLS) info

I agree. That would be useful; see 'which' of the server entries connected if there are more than one entry for a given serv-but-different ports (some people use SSL, with a non-SSL fall-back in the list)

> if an IP doesn't resolve to a hostname, it's an empty element; I feel the IP could be used then

Good point.

I like both.

PR #583, opened 2018-08-26
Use IP if no hostname is present in dcclist.

Cizzle:
Found by: Cizzle
Patch by: Cizzle
Fixes: Suggestion 1 in #155 

One-line summary: Use the IP if no hostname is present ofr entries in de tcl `dcclist`.

As this might break parsers, setting this to milestone 1.9.0.

### DIFF
--- a/src/tcldcc.c
+++ b/src/tcldcc.c
@@ -666,7 +666,8 @@ static int tcl_dcclist STDVAR
       }
       list[0] = idxstr;
       list[1] = dcc[i].nick;
-      list[2] = dcc[i].host;
+      list[2] = (dcc[i].host == NULL || dcc[i].host[0] == '\0') ?
+                iptostr(&dcc[i].sockname.addr.sa) : dcc[i].host;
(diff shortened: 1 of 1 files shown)


thommey (approved):

thommey:
How does it break parsers? the IP instead of the hostname should be handled fine

Cizzle:
Because someone might have made a parser that relies on "{}" being there if no hostname could be resolved? And/or a parser that has `*\.[a-z]+$` for the hostname field?

vanosg:
Per the other half of #155 I added in ports/SSL status:

```
.tcl dcclist
Tcl: {5 (telnet) * 5184 TELNET {lstn  5184} 1561740285} {6 (dns) 127.0.0.53 0 DNS {dns   (ready)} 1561740285} {1 -HQ llama@console 0 CHAT {chat  flags: cptEp/0} 1561740294} {10 (server) barjavel.freenode.net +7000 SERVER {serv  (lag: 0)} 1561740286}
```

Does that scratch the itch for you?

PR #584, opened 2018-08-26
Add port and TLS info as a separate list item for tcl dcclist.

Cizzle:
Found by: Cizzle
Patch by: Cizzle
Fixes: Suggestion 2 at #155 

One-line summary: Adds the port and possible TLS info as a separate list item to tcl `dcclist`.


thommey (changes requested):
LGTM but needs to be adjusted to show '-' after #552

vanosg:
#552 closed, OBE.

thommey (approved):

Cizzle:
Code doing the same was aded in commit b29d8b6 so after unconflicting there isn't that much left to do.

PR #585, opened 2018-08-27
Fix c89 comments and arg validation

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: c89 comments, arg validation

One-line summary:
Fixes c89 comments, option alignment, flag constants, exit code for bad option, static functions

**Fix "CFLAGS="-std=c89" ./configure" throws compiler error: " error: C++ style comments are not allowed in ISO C90" found with gcc**
Fix comments to be c89 style
Align help/usage options
Improve command line arg validation code readability
exit(1) instead of exit(0) for bad option
Make show_help() and show_ver() static functions
Don't hardcode program name for show_help() / Usage
Slightly changed wording for show_help() / Usage -h and -v

this patch was part of #549 and is hereby seperated to be easy to merge

first i wanted only to fix the comments to be c89,
but the non c89 comments in main.c showed some code there
which screamed for more improvement
and while reading and testing it became clear
**there was also a tiny bug inside
in case of an unknown bad flag eggdrop would exit(0) as if everything were fine**

tested all command line options and a bad option

### DIFF
--- a/src/lang.h
+++ b/src/lang.h
@@ -241,7 +241,6 @@
 #define IRC_NICK_FLOOD          get_language(0x673)
 
 /* Eggdrop command line usage */
-#define EGG_USAGE               get_language(0x700)
 #define EGG_RUNNING1            get_language(0x701)
 #define EGG_RUNNING2            get_language(0x702)
 #define EGG_NOWRITE             get_language(0x703)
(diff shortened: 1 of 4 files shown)


thommey (approved):

vanosg:
build

PR #587, opened 2018-08-27
Remove all inline keywords

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: c89 and optimization bloat

One-line summary:
inline keyword is not c89. and as thommey put it: eggdrop will suffer 450% performance penalty due to this patch ;). no, rly, the compiler does a great job, if we let her.

just a quick ./eggdrop -nt, worked.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -796,7 +796,7 @@ int init_language(int);
 int ssl_init();
 #endif
 
-static inline void garbage_collect(void)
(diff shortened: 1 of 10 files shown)


thommey (approved):
LGTM

vanosg:
build

PR #588, opened 2018-08-28
update aclocal.m4

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: update aclocal.m4

One-line summary:
update aclocal.m4: use gnu autoconf macro using AC_SEARCH_LIBS instead of AC_CHECK_LIB and update ax_create_stdint_h.m4

**Additional description (if needed):**

use gnu autoconf macro using AC_SEARCH_LIBS instead of AC_CHECK_LIB:

https://src.fedoraproject.org/rpms/eggdrop/blob/master/f/eggdrop-1.6.21-no_libdns.patch
https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Libraries.html
https://www.gnu.org/software/autoconf-archive/
http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_lib_socket_nsl.m4

update ax_create_stdint_h.m4 from version 2010-03-01 to version 2018-07-27:

http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_create_stdint_h.m4

moved both m4 files, and the already included tcl.m4, into new m4 dir

**note:** you need to run /misc/runautotools to fully test this patch, but eggdrop would also compile/run without doing so.

./configure

diff on arch linux:

before:

checking for socket in -lsocket... no
checking for connect in -lnsl... yes
checking for gethostbyname in -ldns... no

after:

checking for library containing gethostbyname... none required
checking for library containing socket... none required

diff on sunos 5.11:

before:

checking for socket in -lsocket... yes
checking for connect in -lnsl... yes
checking for gethostbyname in -ldns... no

after:

checking for library containing gethostbyname... -lnsl
checking for library containing socket... -lsocket

make worked on both systems, only difference, ld flags under sunos are reversed, but it doesnt matter if its "-lnsl -lsocket" or "-lsocket -lnsl"


thommey (approved):
LGTM, the licenses are acceptable now

vanosg:
If I'm reading this correctly, we now have a ./m4 directory, but it only has some of the m4 files in it (tcl.m4, aclocal.m4 are not). Is this intentional?

michaelortmann:
aclocal.m4 is the main one, it will reside where it is.
this aclocal.m4 then includes other m4s and thoseshould be in m4 directory,
so will see that tcl.m4 will be put into m4/ also...

michaelortmann:
aclocal.m4 is the main one, it will reside where it is and include all the other ones, as it already does with tcl.m4. now ive also moved tcl.m4 into m4/ directory. ty.

looking for review/merge again...

PR #590, opened 2018-08-28
Add support links to motd

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: enhancement

One-line summary:
see commit

it was my idea to add support links to eggdrop motd, like
https://github.com/freebsd/freebsd/blob/master/etc/motd
open for discussion, including reject


vanosg:
Good idea- consolidated just a bit.

PR #592, opened 2018-08-30
Use strlcpy() instead of strncpyz()

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes:

One-line summary:
Use strlcpy() instead of strncpyz() and define strncpyz() to use strlcpy() instead of strncpy()

**Additional description (if needed):**
Nice side effect: less gcc stringop-truncation warnings
strncpyz() was a macro using strncpy().
we dont want strncpy due to its zero padding.
strncpyz() was replaced with strlcpy()
if the operating system / libc provides strlcpy() we will use it.
if it doesn't we will use our new src/compat/strlcpy, which is a copy of the bsd implementation, see:
https://raw.githubusercontent.com/freebsd/freebsd/master/lib/libc/string/strlcpy.c
strncpyz() is exportet to modules for backward compatibility

****note:** you need to run /misc/runautotools to fully enjoy/test this patch**, but eggdrop would also compile/run without doing so (in which case it wont detect operating system strlcpy() and will always use compat strlcpy()).

**Test 1: Comparison of output of make:**

branch develop:
  make 2>&1 | tee make.before.txt
branch strlcpy:
  make 2>&1 | tee make.after.txt
diff:
  -rw-r--r-- 1 michael michael 32538 Aug 30 03:53 make.after.txt
  -rw-r--r-- 1 michael michael 36713 Aug 30 03:51 make.before.txt
its smaller / some of the following gcc warnings are no more:
  warning: ‘strncpy’ output may be truncated copying 323 bytes from a string of length 323 [-Wstringop-truncation]
  note: in expansion of macro ‘strncpyz’

So, this patch also is a good start to fix gcc 8 compiler warnings, esp. -Wstringop-truncation

**Test 2: Linux**

strlcpy() was not found by configure, ok

compat strlcpy() is used, ok:

```
$ nm -l eggdrop|grep strlcpy
000000000005bd20 T strlcpy	/home/michael/projects/eggdrop/src/compat/strlcpy.c:51
$ nm -l modules/blowfish.so|grep strlcpy
(empty output is fine, because in this case eggdrop modules use compat strlcpy() exported by eggdrops module api)
```

quick test: make, install, run, some commands, ok

**Test 3: Freebsd**

strlcpy() was found by configure, ok

operating system strlcpy() is used, ok:

```
$ nm -l eggdrop|grep strlcpy
                 U strlcpy@@FBSD_1.0
$ nm -l modules/blowfish.so|grep strlcpy
                 U strlcpy@@FBSD_1.0
```

quick test: make, install, run, some commands, ok


src/mod/module.h
```diff
@@ -501,6 +504,7 @@
 #define tcl_resultint ((int (*)(void))global[300])
 #define tcl_resultstring ((const char *(*)(void))global[301])
 #define getdccfamilyaddr ((int (*) (sockname_t *, char *, socklen_t, int))global[302])
+#define strncpyz (global[303])
```
thommey: shouldn't we export strlcpy() instead and just define strncpyz for modules?

michaelortmann: see my comment below...

michaelortmann:
1. all strncpy() were replaced by strlcpy()

2. it took me some time to export strlcpy (only in case we use compat strlcpy), because other compat versions did some prefix/namespace stuff like egg_strncasecmp. but i finally got it.

3. strncpy() is now only exportet to modules for backward compatibility

michaelortmann:
all strncpy() were replaced by strlcpy():

find ./ -type f -exec sed -i 's/strncpyz(/strlcpy(/g' {} \;

**note: dont do this in git directory. cd src/ first!**

it took me some time to export strlcpy() to modules (only in case we use compat strlcpy), because other compat functions do some prefix/namespace stuff like egg_strncasecmp. but i finally got it done with #ifdef.

strncpy() is now only exported to modules for backward compatibility

ready for review again.

Cizzle:
LGTM

vanosg:
build

michaelortmann:
Note that the -Wstringop-truncation is not necessarily gone because its fixed. strlcpy() is not checked by gcc. Meanwhile i already invested some time into checking and fixing all strlcpy()s. To do that automatically  i temporarely replace all strlcpy() with strncpy() and let gcc help me.

PR #593, opened 2018-08-30
fix invalid memory access

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: invalid memory access

One-line summary:
underrun for s[strlen(s) - 1]

the parser was ugly and so i rewrote it.

with gcc:
CFLAGS=-fsanitize=address,undefined
type \n at the partyline and the sanitizer will report the underflow


thommey (approved):
took me a while but this looks fine

PR #595, opened 2018-08-31
(Fix #504) Fix --with-sslinc --with-sslib for openssl 1.0

### LINKED ISSUE #504: OpenSSL buf2hexstr issue
Found by: cpg

With openssl-1.0.2n installed at /usr/local/openssl/  and eggdrop compiled with ./configure --with-sslinc=/usr/local/openssl/include/ --with-ssllib=/usr/local/openssl/lib/  : 

```
checking for openssl/ssl.h... yes
checking for openssl/x509v3.h... yes
checking for SSL_accept in -lssl... yes
checking for X509_digest in -lcrypto... yes
checking for hex_to_string... no
checking for OPENSSL_hexstr2buf... no
checking for string_to_hex... no
checking for OPENSSL_buf2hexstr... no
checking for OpenSSL... no (make sure you have version 0.9.8 or higher installed)
```

Note that it finds ssl.h, and doesn't complain about "Invalid path to OpenSSL files" which it would if it couldn't find OpenSSL (As in the error ```WARNING: Invalid path to OpenSSL headers. /etc/openssl/ doesn't contain the required files.```) . Instead, because its choking on buf2hextr, it claims it can't find OpenSSL, when it has in fact found that.

a) probably need to amend that error message a bit? a
b) We worked on this as part of #323 , maybe we need to touch it up a bit?
c) Caps for OPENSSL? yuck :)

michaelortmann:
Found by: cpg
Patch by: michaelortmann
Fixes: #504 

One-line summary:
Fix --with-sslinc --with-sslib for openssl 1.0

1.
aclocal.m4 didn't really use --with-ssllib for hex_to_string/string_to_hex detection
2.
some parts of aclocal.m4 save $CC/$CPP, then set them to include $sslinc and then resets them
doing the same here doesn't help fixing the bug, because we need the ld path changed
3.
i have the feeling the rest of aclocal.m4s handling of $CC/$CPP/$sslinc/$SSL_LIB may also be broken
4.
i am no autoconf expert, but i think setting CPPFLAGS/LDFLAGS instead of $CC/$CPP should be preferred
5.
after the configure was fixed, the make failed while linking:
[...]
ssl_sess.c:(.text+0x208c): undefined reference to `ENGINE_get_ssl_client_cert_function'
[...]

reversing the order of -lcrypto -lssl fixed the make

this patch is a minimal solution for #504

./configure
[...]
checking for hex_to_string... no
checking for OPENSSL_hexstr2buf... yes
checking for string_to_hex... no
checking for OPENSSL_buf2hexstr... yes
[...]
SSL/TLS Support: yes (OpenSSL 1.1.0i  14 Aug 2018)
[...]

./configure --with-sslinc=/home/michael/opt/openssl-1.0.2p/include/ --with-ssllib=/home/michael/opt/openssl-1.0.2p/lib/
[...]
checking for hex_to_string... yes
checking for string_to_hex... yes
[...]
SSL/TLS Support: yes (OpenSSL 1.0.2p  14 Aug 2018)
[...]

a tiny test compile and run was successful

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -1616,13 +1616,13 @@ AC_DEFUN(EGG_TLS_WITHSSL,
   [
     if test "$enable_tls" != "no"; then
       if test -d "$withval"; then
-        AC_CHECK_LIB(ssl, SSL_accept, , [havessllib="no"], [-L$withval -lcrypto])
         AC_CHECK_LIB(crypto, X509_digest, , [havessllib="no"], [-L$withval -lssl])
+        AC_CHECK_LIB(ssl, SSL_accept, , [havessllib="no"], [-L$withval -lcrypto])
         if test "$havessllib" = "no"; then
           AC_MSG_WARN([Invalid path to OpenSSL libs. $withval doesn't contain the required files.])
         else
           AC_SUBST(SSL_LIBS, [-L$withval])
-          LDFLAGS="${LDFLAGS} $SSL_LIBS"
+          LDFLAGS="${LDFLAGS} -L$withval"
         fi
       else
         AC_MSG_WARN([You have specified an invalid path to OpenSSL libs. $withval is not a directory.])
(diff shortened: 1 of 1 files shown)


thommey (approved):
LGTM

Issue #176, opened 2016-02-14
Investigate various OS patches


thommey:
Check which of http://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/chat/eggdrop/patches/ are necessary, please also check what other distributions patch into eggdrop (gentoo, ports, macports, etc.).

Pixelz:
Investigate cygwin patch at http://windrop.sourceforge.net/compile18cvs.html

michaelortmann:
http://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/chat/eggdrop/patches/patch-src_eggdrop.h
and
http://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/chat/eggdrop/Makefile
means, we should enhance autoconf/configure support for
Fine tuning of the installation directories
?

michaelortmann:
Investigate freebsd patches at:
https://svnweb.freebsd.org/ports/head/irc/eggdrop-devel/files/
https://svnweb.freebsd.org/ports/head/irc/eggdrop/files/

irc discussion result: remove 99-100% of inline keywords, they are also not c89

investigate opensuse linux patches at:
https://build.opensuse.org/package/show/home%3Azombah%3Abranches%3Ahome%3Aonielsen/eggdrop

investigate trueos patches at:
https://github.com/trueos/trueos-ports/tree/trueos-master/irc/eggdrop/files

investigate gentoo linux patches at:
https://gitweb.gentoo.org/repo/gentoo.git/tree/net-irc/eggdrop/files

investigate dragonfly bsd patches at:
https://github.com/DragonFlyBSD/DPorts/tree/master/irc/eggdrop

michaelortmann:
> Investigate cygwin patch at http://windrop.sourceforge.net/compile18cvs.html

This patch must NOT be applied. cygwin console is fine, but cmd.exe hangs after:
```
eggdrop.exe-1.9.0 eggdrop.conf
exit
```

michaelortmann:
This Issue is done. Only thing remaining was moved into #474. I think you can close this Issue here now.

PR #606, opened 2018-09-03
Update operating system support (DragonFly BSD)

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: Part of #176 

One-line summary:
Enhance "DragonFly BSD" support, switched default build from static to shared

This patch is based on
https://raw.githubusercontent.com/DragonFlyBSD/DPorts/master/irc/eggdrop/dragonfly/patch-configure
and enhances "DragonFly BSD" support and *BSD Support in general
there was a wildcard *BSD in aclocal.m4
which may match too many *BSD and the same time doesnt match DragonFly
this patch replaces this wildcard with proper Names
this patch also enhances doc for dragonfly bsd
this patch also rips out support for BSD/OS (originally called BSD/386 and sometimes known as BSDi) because Wind River discontinued sales of BSD/OS at the end of 2003, with support terminated at the end of 2004."

I also tested eggdrop on TrueOS and it worked out of the box, because its only a minimal modified FreeBSD. If there is any official list of supported/tested operating systems for eggdrop besides "any c89/posix2001 should work", we could add DragonFly and TrueOS now, but i didnt find such a list. But what i found was doc/BUG-REPORT and i updated it.,

**when merging, please run misc/runautotools**

The following test was done on "DragonFly BSD" 5.2.2:

**before:**

./configure
```
[...]
WARNING:

  The operating system you are using has not yet had a great
  deal of testing with Eggdrop. For this reason, this compile
  will default to "make static".

  To enable module support, type "make eggdrop" instead of just
  "make" after you run "make config" (or "make iconfig").

  As we have not done a sufficient amount of testing on this
  OS, your feedback is greatly appreciated. Please let us know
  at bugs@eggheads.org if there are any problems compiling with
  module support, or if you got it to work :)
[...]
```

**after:**

./configure
make config
make
make install
./eggdrop -nt
```
[...]
Hey -HQ!  My name is dragonflybsd and I am running eggdrop v1.8.3+acupdates, on DragonFly 5.2-RELEASE.
[...]
```


thommey (approved):
Should be fine, until someone complains :)

PR #613, opened 2018-09-05
(Fix #185) Enhance TLS

### LINKED ISSUE #185: TLS: add a few extra settings
In addition to the `ssl-ciphers` setting I'd like to request that an `ssl-protocols` setting be added, so that for example protocols such as SSLv2 and SSLv3 can be disabled or only TLSv1.2 be allowed.

Maybe also a setting to disable compression? Or even hardcode it (probably the better option).

Also, since eggdrop acts as a server for its listening ports, a setting like `ssl-dhparam` for specifying a custom Diffie-Hellman parameters file is probably in order as well. https://weakdh.org/

michaelortmann:
Found by: Robby-
Patch by: michaelortmann
Fixes: #185

One-line summary:
Enhance TLS

1. Add new config variable ssl-protocols
2. Add new config variable ssl-dhparam
3. Drop insecure ssl compression
4. Add comment to eggdrop.conf that setting ssl-ciphers may not impact TLSv1.3 ciphersuites

i tested the new code and variables with temp debug printf and with .link sslbot.
looks good.


Cizzle:
Don't see why this couldn't go in 1.8.4; still lacking some docs though.

src/tls.c
```diff
@@ -168,6 +170,94 @@ int ssl_init()
            ERR_error_string(ERR_get_error(), NULL));
     ERR_free_strings();
   }
+  /* Let advanced users specify the list of allowed ssl protocols */
+  #define EGG_SSLv2   1 << 0
```
thommey: paranthesis around definition please (1 << 0), in case it's moved and used in odd places

thommey (dismissed):

src/tls.c
michaelortmann: done

michaelortmann:
@thommey: Requested changes done.

@Cizzle: Im no native speaker, but if you can tell me, **where** i should add documentation i will do.

Cizzle:
At `help/core.help` (adding to the `set` help) and `help/set/cmds1.help`.

michaelortmann:
@Cizzle: as far as i understand: `help/core.help` is more about "on the fly changeable" settings; the new settings are STR_PROTECT; so i think were done here. I added help text to `help/set/cmds1.help` as requested. eggdrop.conf was changed accordingly.

vanosg (approved):

thommey (approved):
This is fine but let's hope we don't get TLSv1.4 soon and have to redo this

Issue #520, opened 2018-05-21
REHASH: duplicate chanfile saves & freezes


ghost:
REHASH causes 1 bot I run to freeze.  (As this seems to be a disputed term, let me clarify now before I'm asked, yet again: it. stops. working. and. does. nothing.)

Other bots on the same box do not. Recompiling does not resolve. I have duplicated with NO SCRIPTS LOADED (raw bot).

Walkthrough:
\* EGGDROP v1.8.3
\* TCL v8.6.8
\* $::raw-logs = 1 (so additional console data can be seen; see bug report #519 )

[2018-05-20 18:17:28 -0800] <Bot> [03:17:37] #Domino# console +*
[2018-05-20 18:17:28 -0800] <Bot> Set your console to #bots: mpjkcoblrxsdwvtuhg12345678 (msgs, public, joins, kicks/modes, cmds, misc, bots, linked bot msgs, raw, files, server, debug, wallops, server output, botnet incoming, botnet outgoing, incoming share traffic, outgoing share traffic, level 1, level 2, level 3, level 4, level 5, level 6, level 7, level 8).

I've been told to alter the config file and create log files to demonstrate the information requested by Geo. Apparently, pasting the console output is useless.  [STMelon]

I won't; ".console" exists for a reason.

[2018-05-20 18:17:37 -0800] <Domino> .tcl rehash
[2018-05-20 18:17:37 -0800] <Bot> [03:17:46] Writing user file...
[2018-05-20 18:17:38 -0800] <Bot> [03:17:46] Writing channel file...
[2018-05-20 18:17:38 -0800] <Bot> [03:17:46] Rehashing...
[2018-05-20 18:17:38 -0800] <Bot> [TCL (1ms)] 
[2018-05-20 18:17:38 -0800] <Bot> [03:17:47] Writing channel file...

... and that's it. The bot STOPS there.

[2018-05-20 18:18:01 -0800] <Domino> .tcl rehash
\<bot does not respond\>

Three questions:

(1) want to know what the code makes it do after the second "writing channel file" (so I know where to trouble-shoot, if it's within my ability)
(2) why does it write the chanfile TWICE
(3) why does it return from the TCL command, then continue with another chanfile write (as if it's being done as a background-triggered item) after it has returned

Could it be a similar issue to the UPDATE bug I filed years ago. ( http://cvs.eggheads.org/viewvc/eggdrop1.6/doc/Changes1.6?view=markup#l100 )

[...]

PR #615, opened 2018-09-05
Fix a truncation in delignore()

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: May be relevant to #520

One-line summary:
Fix a truncation in delignore()
use UHOSTLEN, as with addignore, see cmds.c:cmd_pls_ignore():
  char *who, s[UHOSTLEN];
    addignore(s, dcc[idx].nick, par, expire_time ? now + expire_time : 0L);

i was able to .+ignore a long string, so that line
strncpyz(temp, ign, sizeof temp);
in delignore() was truncating the string.
in my case strlen(ign) was 319 and sizeof temp was 256

you could put debug code before line
strncpyz(temp, ign, sizeof temp);
to output strlen(ign)
and if you then do .+ignore a long string
you will reach a strlen > 255
this means truncation

you should also be able to demonstrate a real life bug, if the bot would share userlist to another bot. in this case, the other bot should receive a truncated ignore string


thommey (approved):
LGTM

michaelortmann:
WIP because unsure about max len 320

thommey (approved):

PR #617, opened 2018-09-05
Update compat/md5

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Update compat/md5 from https://openwall.info/wiki/people/solar/software/public-domain-source-code/md5

compiles fine
.tcl md5 fine.
extern openssl md5 and intern md5 produced same result.


thommey (approved):
I wonder if this changes anything... H and H2 seem like optimizations only, the rest looks harmless

PR #619, opened 2018-09-10
(WIP) Fix race open()/chmod(): use umask() instead

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: Race open()/chmod()

One-line summary:
Fixes a race condition between open() and chmod().

Instead of chmod() after _some_ open() for write calles, we will use a global process umask() now.
I guess the description of userfile-perm in eggdrop.conf is still fine, even though it now also applies to all written files, like the share file (the share file was getting hardcoded permissions before).

i was also thinking about other ways, like using posix open() instead of fopen() because its additional perm paramter, but fprintf() wouldnt work and dprintf() is posix 2008 only, and so i reasearched on... to make it short, i really think umask() is the right thing to do here.

this patch is only internal and wont be user visible.

i tested with default
set userfile-perm 0600
and the user and chanfile was written as always, as should, as
-rw-------


thommey:
I'd prefer a different fix for this, because setting the process-wise umask without restoring it immediately can cause undesired side-effects on Tcl scripts creating files. replacing open with fopen adds an annoying conversion of the int fd to filestream, yeah (although it does work both ways, it's really inconvenient).

fchmod() is int fd only, so we might as well use open there... open() then fdopen() to get the stream would be the annoying option. any other ideas? mk*temp() and then moving don't help either because they don't take modes (nor does tmpfile()). I see the problem.

michaelortmann:
i would go for "umask() + fdopen() + restore umask()" now as it looks like the most simple solution.

thommey (changes requested):
restore umask after using it and this is fine

michaelortmann:
From linux uname(2):

It is impossible to use umask() to fetch a process's umask without at the same time changing  it.   A  second call  to umask() would then be needed to restore the umask.  **The nonatomicity of these two steps provides the potential for races in multithreaded programs.**

I need more time to think about this.

Eggdrop is threaded, so its possible, other (Tcl) threads use our temporary umask even if we restore it as fast as possible.

A race.

We cant avoid it. We cant suspend (all) other threads.

Now the question is: is it better to have a race condition with chmod() and other programs or a race condition between eggdrop threads.

My answer is, we take our chance and switch the race over to the eggdrop threads.

That means, we continue going for the "umask() + fdopen() + restore umask()" solution.

michaelortmann:
More news:

There is another difference between chmod() and umask(), chmod() would also effect an already existing file. so we cant simply replace chmod() with umask(), we would have to do a bit more. maybe we must leave the chmod in, additionally to the umask.

the more i think about it. its may as well be best to leave it now as it is.

the good thing is, the analyzation resulted in some findings, like the following:

share.c:` } else if (!copy_to_tmp && !(f = fopen(s, "wb"))) {` we should do better error handling for fopen(), we could also apply chmod(), maybe it should, may be problematic code here.

logfiles are opened like in read/write mode? read? i propose we remove the "+":
src/misc.c: `logs[i].f = fopen(s1, "a+");`

there are places where a file is attempted to open with append mode "a", and if it fails with mode "w". that is not needed. "a" includes "Create the file if it does not exist", see `man fopen`.

PR #628, opened 2018-09-16
Enhance Minix support  - Enable shared build

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Enhance Minix support  - Enable shared build

Also fix configure to find clang in case gcc isnt installed and cc is not set
**please run misc/runautotools after merge**

$ ./configure
$ make config
$ make
$ make install
$ ./eggdrop -nt
Hey -HQ!  My name is minix and I am **running eggdrop v1.8.3+acupdates, on Minix 3.3.0.**


Cizzle:
Might be helpful to add the `CC=clang ./configure` prerequisite (if it is one) to the `doc/COMPILE-GUIDE`.

michaelortmann:
@Cizzle: i propose a more general solution, see #675.

the c compiler clang is not found here, because no gcc and no cc.

usually cc links to gcc or clang, so clang would be found if cc links to clang.

but here under minix cc is no there, only clang, so its not found.

we could change aclocal.m4 to find it. we would have to modify AC_PROG_CC into something like

AC_PROG_CC([clang gcc cc])

im unsure about the search order

What do you think?

Cizzle:
That would be nice, I'd keep the order `gcc cc clang` though.

michaelortmann:
@Cizzle: its done.

Cizzle:
LGTM

PR #634, opened 2018-09-18
Fix buffer size in channels.c:write_channel()

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix buffer size in channels.c:write_channel()

1. No security problem, because chanfile can only be set in eggdrop.conf and is STR_PROTECT, and because ~new is probably only written into buffers which are overwritten anyway, and because ~new is hardcoded and cant be altered by mallory.
2. With this patch comes some cleanup/unification

eggdrop.conf:
set chanfile "123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890"
CFLAGS="-g3 -O2 -D_FORTIFY_SOURCE=2" ./configure
$ ./eggdrop -nt
[...]
Aborted (core dumped)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -87,7 +87,8 @@
 #  define _POSIX_SOURCE 1               /* Solaris needs this */
 #endif
 
-extern char origbotname[], userfile[121], botnetnick[];
+extern char origbotname[], botnetnick[]; 
+extern char userfile[121];        /* 121 = sizeof userfile from users.c */
 extern int dcc_total, conmask, cache_hit, cache_miss, max_logs, quick_logs,
            quiet_save;
 extern struct dcc_t *dcc;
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -87,7 +87,7 @@
 #  define _POSIX_SOURCE 1               /* Solaris needs this */
 #endif
 
-extern char origbotname[], userfile[], botnetnick[];
+extern char origbotname[], userfile[121], botnetnick[];
```
Cizzle: Adding another magic number to fix another?
Probably a better idea to move `userfile` from `users.c` to `userrec.c` and `backup_userfile()` from `main.c` to `userrec.c` as well and getting rid of some of these externs and magic numbers. Not really the scope of this PR though.

vanosg: Documented the magic number source, we can clean it up later if we think the benefit outweighs the risk of breakage

PR #635, opened 2018-09-19
strftime() is posix 2001, cleanup compat, egg_strftime() -> strftime()

michaelortmann:
Found by: vanosg
Patch by: michaelortmann
Fixes: 

One-line summary:
strftime() is posix 2001, cleanup compat, egg_strftime() -> strftime()

Debloating and posixing eggdrop step by step.
Fixes 1 part of #226, don't close #226 yet.
**Please misc/runautotools after merge.**
Please also misc/makedepend before release of next eggdrop stable version.

```
$ misc/runautotools
$ ./eggdrop -nt
[...]
[16:23:00] testbot joined #testchan.
[...]
```
A second test linking 2 sharebots was also successful.


Cizzle:
This will break Tcl scripts using `strftime` as we provide the GNU version with it's extensions (like field width modification) and extra conversion specifiers (like %P, %l, ...). If we want this, it'll be for 1.9 or 2.0 even.

michaelortmann:
I dont agree for `src/compat/strftime.c` and `src/compat/strftime.h` use `#ifndef HAVE_STRFTIME`. So, eggdrop already prefers the systems (maybe non gnu) strftime over of its own gnu version. The following proof is with current git version of eggdrop:

```
$ uname
FreeBSD
$ nm -l eggdrop|grep strftime
                 U strftime
```

If tcl authors also got confused, then we have even more reason to apply this patch.

Cizzle:
True indeed, the system's `strftime` is preferred, meaning Tcl scripts are in fact already sort of broken depending on which system they are sourced.
The question becomes then; do we want uniformity and always use our GNU version or ~~only provide the POSIX version with mention~~ use the system one with disclaimer that only POSIX is guaranteed and that GNU extensions (and possibly others) availability depends on the system?

michaelortmann:
Please use the system version, for Alice may rely on the GNU version with GNU extension and and Bob may relay on another version with other extensions.

But we could - maybe - enhance eggdrop, scan through the tcl, and warn the user, if he uses non posix specifiers.

michaelortmann:
now, **where** should we add the disclaimer?

Cizzle:
In tcl-commands.rst at `strftime`.

thommey (approved):
LGTM

vanosg:
Can @Cizzle  or @michaelortmann  update the docs to reflect the disclaimer you mentioned, and clean up the conflicts? Thanks!

michaelortmann:
@vanosg: doc added and conflicts resolved.

PR #639, opened 2018-09-20
Enhance makepass() and make_rand_str()

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Enhance makepass() and make_rand_str()

1. misc.c: **Rewrote make_rand_str() to be more random and faster.** Entropy was worse because of randint(3) toggle beween lower case letters and digits.
2. dccutil.c: **Rewrote makepass() to be more random and faster.** It returned passwords of len 10 to 15. Now it will always return passwords of len 15. There is no documentation about the maximum password length returned, but this function is exported, so i want to leave it 15 for eggdrop 1.8.
3. dcc.c: ps[20] -> pass[16] because now we know the max length returned from makepass(). While at it, equally renamed the array to pass.
4. share.c: p[32] -> pass[16] because now we know the max length returned from makepass(). While at it, equally renamed the array to pass. **The code was really fooling the reader to think a share password of len 31 would be calculated.**

Test 1: quick compile, run, link to anotherbot, share userfile, ok.
Test 2:
Before:
```
$ cat BotB.user
[...]
BotA       - b                       
--LASTON 1544228471 linked
--PASS 2wr3jmjw5hrz
--BOTADDR 127.0.0.1:3333/3333
--XTRA created 1541221233
```

```
.chpass BotB
[03:58:11] tcl: builtin dcc call: *dcc:chpass -HQ 1 [something]
[03:58:11] #-HQ# chpass BotB [nothing]
Removed password.
.link BotB
```

After:
```
$ cat BotA.user
[...]
BotB       - b                       
--PASS wysjsb8w6nuau73
--LASTON 1544410699 linked
--BOTADDR 127.0.0.1:3343/3343
--XTRA created 1541218572
```


Cizzle:
LGTM and possible for 1.8.4 IMHO.

src/misc.c
```diff
@@ -1438,17 +1438,14 @@ void show_banner(int idx)
...
 {
-  int j;
+  static const char chars[] = "0123456789abcdefghijklmnopqrstuvwxyz";
 
-  for (j = 0; j < len; j++) {
-    if (!randint(3))
-      s[j] = '0' + randint(10);
-    else
-      s[j] = 'a' + randint(26);
+  for (int i = 0; i < len; i++) {
```
thommey: C89 please, declare i at the beginning of the function definition :)

thommey (approved):
I'd prefer a define for the password size or passing "sizeof XXX" to makepass if you agree, instead of the 16 everywhere and relying on the buffer being big enough

src/misc.c
michaelortmann: ty, done.

michaelortmann:
makepass(char *pass) is bad API, should be void makepass(char *pass, size_t len), but its part of module API, so we cant change that for 1.8 and i still would like to see this PR ready / merged for 1.8. Should we put a FIXME in the source or open an Issue to remember? If one day 1.8 will be closed and we are working on 1.9, and this PR is still open, then this message becomes obsolete and it should be changed then. Besides: The define PASWORDLEN is done.

PR #649, opened 2018-09-26
Fix buffer size - fix 2 gcc 8 format-truncation warnings

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix buffer size - fix 2 gcc 8 format-truncation warnings

1. dcc.c:timeout_dupwait(): Replaced hardcoded buffer size 100 with NICKLEN + UHOSTLEN. That gcc warning was no false alarm.
2. chan.c:detect_chan_flood: Fixed buffer size analog (1).

Note1: don't alter CFLAGS, or you may miss this compiler warning

Note2: even gcc 7.3.0 under current cygwin (windrop) is reporting this warning, **in color!**

quick ./eggdrop -nt, was ok, but that didn't traverse the changed code. Still, im feeling fine about this patch.

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -1470,7 +1470,7 @@ static void dcc_dupwait(int idx, char *buf, int i)
  */
 static void timeout_dupwait(int idx)
 {
-  char x[100];
+  char x[NICKLEN + UHOSTLEN];
 
   /* Still duplicate? */
   if (in_chain(dcc[idx].nick)) {
--- a/src/mod/irc.mod/chan.c
+++ b/src/mod/irc.mod/chan.c
@@ -324,10 +324,10 @@ static int detect_chan_flood(char *floodnick, char *floodhost, char *from,
       strcpy(ftype + 4, " flood");
       u_addban(chan, h, botnetnick, ftype, now + (60 * chan->ban_time), 0);
       if (!channel_enforcebans(chan) && (me_op(chan) || me_halfop(chan))) {
-        char s[UHOSTLEN];
+        char s[NICKLEN + UHOSTLEN];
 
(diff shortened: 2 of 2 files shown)


thommey (approved):

vanosg:
build

Cizzle:
Looking at the following defines in eggdrop.h I'd say we got too large buffers now:
```
#define NICKLEN      NICKMAX + 1
#define HANDLEN    32 /* valid values 9->NICKMAX                             */
#define NICKMAX    32 /* valid values HANDLEN->32                            */
#define UHOSTMAX    291 + NICKMAX /* 32 (ident) + 3 (\0, !, @) + NICKMAX */
#define UHOSTLEN     UHOSTMAX + 1
#define NICKLEN      NICKMAX + 1
```
If we do `NICKLEN + UHOSTLEN` we actually get `NICKMAX + 1 + 291 + NICKMAX + 1` or a waste of `NICKMAX + 1` bytes. I'm not even sure why we have UHOSTLEN as UHOSTMAX already takes the `\0` into account (maybe we should remove it from UHOSTMAX).

michaelortmann:
When i saw

egg_snprintf(s, sizeof s, "%s!%s", m->nick, m->userhost);

I blindly calculated size(s) == size(%s!%s) == size(m->nick) + size(m->userhost) == NICKLEN + UHOSTLEN.

See also chan.h:memstruct

Yes, size(s) is probably too long and should be UHOSTLEN only.

But there is consequence:

size(m->userhost) is then too long also, and should be more like UHOSTLEN - NICKLEN.

(all sizes +-1 here ;)

Yes, we should fix all lengths to be accurate to the byte.

Issue #627, opened 2018-09-15
Missing debug log


michaelortmann:
Missing debug log when eggdrop is startet with `./eggdrop -nt`

its missing from console until console +d can be set, which is possible only after eggdrop is fully loaded
console flags for -HQ are not saved to userfile, because -HQ is not saved.

and its missing from log file until eggdrop does "use_stderr = 0;" in main.c
i set +d for logfile in eggdrop.conf

my initial problem was that i would like to see the debug log of tls.c:ssl_init() when i start eggdrop with -nt

is this fixable?
at least for console for start with -nt?
or should we move the call to ssl_init() after "use_stderr = 0;"?

after some more analysis, i thought i could add LOG_DEBUG to EGG_BG_CONMASK in main.h
its hardcoded, but i could live with it
but its also too late:
dcc[term_z].u.chat->con_flags = conmask | EGG_BG_CONMASK;
is executed also only after ssl_init(), and even after "use_stderr = 0;"

my temporary solution is to replace line:
else if ((type & LOG_MISC) && use_stderr) {
with:
else if (((type & LOG_MISC) || (type & LOG_DEBUG)) && use_stderr) {
in misc.c:putlog()

but we still need a good permanent solution to see LOG_DEBUG in console and/or logfile from the start of eggdrop -nt by default or opt in

**Maybe whilefixing #658 i found the cause:** 

misc.c:
```
  if (!backgrd && !con_chan && !term_z)                                          
    dprintf(DP_STDOUT, "%s", out);  
```

Shouldn't that read term_z instead of !term_z ?

michaelortmann:
set console
in eggdrop.conf
was the solution,
ty geo

michaelortmann:
set console
in eggdrop.conf
.store
etc do not solve this issue.

Cizzle:
> Shouldn't that read term_z instead of !term_z ?

No, you'll get every logmessage then: those you haven't set with consoleflags and those you have will even be double.
You're at the right spot though for any mod, though it should be evaluated what the best way would be.

michaelortmann:
**1. With the now "dead" debug log activated a start of eggdrop would look like:**

```
$ ./eggdrop -nt
[03:42:38] LANG: Language loaded: english
[03:42:38] LANG: Section loaded: core
[03:42:38] LANG: 390 messages of 431 lines loaded from ./language/core.english.lang
[03:42:38] LANG: 390 adds, 0 updates to message table

[...]

Issue #402, opened 2017-05-15
Deprecate -nt/-nc as valid command line args


vanosg:
Per discussion in #391, deprecate requiring -n with -c and -t flags by changing help display and docs to indicate just -c or -t as method to activate those functionalities. Show warning to user indicating -nt/-nc will be removed in future release.

PR #658, opened 2018-09-30
Fix: ./eggdrop -nt doesn't start if "listen" and "loadmodule dns" commented out

michaelortmann:
Found by: paigeadelethompson
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix: ./eggdrop -nt doesn't start if "listen" and "loadmodule dns" commented out

If eggdrop is started with `./eggdrop -nt`
Usually the first dcc idx will be set to TELNET (eggdrop.conf: listen)
Then the second dcc idx will be set to DNS (eggdrop.conf: loadmodule dns)
And the third dcc idx will be set to DCC_CHAT (as term_z for foreground user)

If the following 2 eggdrop.conf settings are commented out:
```
# loadmodule dns
# listen
```

eggdrop was fatal exiting with the following misleading error message:
```
[21:22:51] * ERROR: Failed to initialize foreground chat.
[21:22:51] Warning: Attempt to kill un-allocated socket 0!
```

This was due to the code assuming (and commenting) that the term_z / idx would be > 0.
But in this case it is 0.

Since commit https://github.com/eggheads/eggdrop/commit/1e7678c00d7361ce9b59907d39b999e90d49ca82 term_z is dualused as a flag and as an idx. This patch takes account of that and only fixes a corner case.

This fix is related to #402 and #627, but standalone and ready for review now.

So we can easily trigger the error when both settings are commented out in eggdrop.conf:

eggdrop.conf:
```
# loadmodule dns
# listen
```
Before:
```
$ ./eggdrop -nt
[...]
[21:22:51] * ERROR: Failed to initialize foreground chat.
[21:22:51] Warning: Attempt to kill un-allocated socket 0!
```
After:
```
$ ./eggdrop -nt
*** -HQ joined the party line.
```

### DIFF
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -272,7 +272,7 @@ void tell_verbose_uptime(int idx)
   if (backgrd)
     strcpy(s1, MISC_BACKGROUND);
   else {
-    if (term_z)
+    if (term_z >= 0)
       strcpy(s1, MISC_TERMMODE);
     else if (con_chan)
       strcpy(s1, MISC_STATMODE);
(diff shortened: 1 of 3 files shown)


thommey (approved):

vanosg:
build

PR #660, opened 2018-09-30
Fix struct hack / flexible array member

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix struct hack / flexible array member

The struct hack code is "new" in eggdrop and it fails on systems like Haiku alpha1/gcc 2.93.3:
`Fix coredns.c:154: field data has incomplete type`
This general fix uses Macro AC_C_FLEXIBLE_ARRAY_MEMBER, see:
https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/C-Compiler.html
This fix also helps eggdrop to be C89 compatible.
**#548 now also requires this patch, due to it also does struct hack.
please run misc/runautotools after merge**

a quick compile and run was fine

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -97,6 +97,7 @@ AC_CHECK_HEADERS([arpa/inet.h fcntl.h limits.h locale.h netdb.h netinet/in.h std
 AC_C_CONST
 AC_C_BIGENDIAN
 AC_C_INLINE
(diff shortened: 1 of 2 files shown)


Cizzle:
LGTM

PR #661, opened 2018-10-01
Add volatile keyword to do_restart, its used inside signal handler

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Add volatile keyword to do_restart, its used inside signal handler

1. I can't demonstrate a bug here, so this patch is a "better safe than sorry" one.
2. Don't worry about the volatile keyword, configure.ac already uses "AC_C_VOLATILE"

I did only small test: link 2 share bots, join a server, execute a few random tcl commands.


src/main.c
```diff
@@ -141,9 +141,9 @@ int notify_users_at = 0; /* Minutes past the hour to notify users of notes? */
 char version[81];    /* Version info (long form)  */
 char ver[41];        /* Version info (short form) */
 
-int do_restart = 0;       /* .restart has been called, restart ASAP */
-int resolve_timeout = 15; /* Hostname/address lookup timeout        */
-char quit_msg[1024];      /* Quit message                           */
+volatile int do_restart = 0; /* .restart has been called, restart ASAP */
```
thommey: While we're at it, it should be "volatile sig_atomic_t do_restart = 0;" to be correct (I know int is probably always atomic, but that's the proper int size for it)

thommey (changes requested):
fine after changing to sig_atomic_t

src/main.c
michaelortmann: there were more places to change. do_restart is even used in modules. i have no idea, if the volatile status is true including modules or if the module api breaks it. it looks more complicated now. my small test was fine, but my confidence isnt. what is your opinion? could it break anything?

michaelortmann:
@thommey: requested changes are done.

thommey (approved):

vanosg:
build

PR #680, opened 2018-10-05
(Fix #679) Fix "openssl s_client" fails with "invalid handle: 1;2c" under rxvt-unicode

### LINKED ISSUE #679: "openssl s_client" fails with "invalid handle: 1;2c" under rxvt-unicode
i try to login to the bot (current git version) like this:

```
$ openssl s_client -cert my.crt -key my.key -connect 127.0.0.1:2514
[...]
---
read R BLOCK
ÿû

Nickname.
^[[?1;2c
```

after entering my nickname, i see:

```
You don't have access.
closed
```

the bot loggs:

`[04:02:27] Refused telnet@localhost.localdomain (invalid handle: 1;2c[...])`

we need to fix that **1;2c**

or did i do anything wrong?

```
$ echo $TERM
rxvt-unicode-256color
```

setting TERM to xterm or vt100 didn't help.

It happens with rxvt-unicode 9.22 but not with xterm 337.

For rxvt-unicode see http://software.schmorp.de/pkg/rxvt-unicode.html

So, i think its an eggdriop bug, or at least eggdrop should be able to handle such configuration.

The FreeBSD manpage has something on this:

https://www.freebsd.org/cgi/man.cgi?query=urxvt&apropos=0&sektion=7&manpath=FreeBSD+11.2-RELEASE+and+Ports&arch=default&format=html
`
if	[ ""${term_id} = '^[[?1;2C' -a ${DISPLAY:-foo} = foo ];	then`

We probably should just ignore input starting with such control codes.

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #679

One-line summary:
Fix "openssl s_client" fails with "invalid handle: 1;2c" under rxvt-unicode

If Nickname starts with "\e[?1;2c", skip those chars.
This looks like a (too) quick solution to _my_ problem.
So maybe there is more to this and you can provide more detail / intel.
Else, please just merge this in for i dont think im the only one who sufferes from this bug.

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -1561,6 +1561,9 @@ static void dcc_telnet_id(int idx, char *buf, int atr)
     lostdcc(idx);
     return;
   }
+  /* rxvt-unicode */
+  if (!strncmp(buf, "\e[?1;2c", 7))
+    buf += 7;
   dcc[idx].user = get_user_by_handle(userlist, buf);
   get_user_flagrec(dcc[idx].user, &fr, NULL);
 


thommey (approved):

vanosg:
build

PR #688, opened 2018-10-09
(WIP) SASL

### LINKED ISSUE #395: Add SASL support
Use libsasl, or write from scratch?

michaelortmann:
Found by: vanosg
Patch by: michaelortmann
Fixes: #395

One-line summary:
Its becoming useful, any early testing / feedback / input is very welcome

1. SAL needs CAP and the needed CAP stuff was recently merged.
2. **History:** We also need base64 for SASL. Research regarding base64 (C, License, Performance(https://github.com/gaspardpetit/base64/)) pointed to implementations by Apache (<= 2.2) and http://w1.fi/wpa_supplicant/ (Jouni Malinen). OpenBSD has got an implementation, but we would need a fallback then. mbedTLS has got an implementation. And Finally mbedTLS base64 is already proposed through #356 (pbkdf2 module), so we should use that mbedTLS implementation here also. So mbedTLS base64 should be in eggdrop core, to not double that code. Hereby i created directory src/mbedtls and placed/coded all the bits we need regarding base64. Though one thing is open: We could think about making that base64 implementation available for tcl scripts. **Update:** Discussion on IRC resulted in using openssl over mbedtls. pbkdf2 will be based on openssl then. Now we could also use openssl base64 here, but we would need a wrapper for openssl BIO bloat/stuff and an API for our functions. Further research found base64 functions in libc. Its no POSIX, its hidden, no man page, but its there under major systems, such as cygwin, freebsd, linux and openindiana (think of solaris/sun) and we will provide the underlying bsd code that those systems share as fallback for any system lacking libc support here.
3.Its becoming useful, any early testing / feedback / input is very welcome. I already see `SASL authentication successful` with PLAIN mechanism, see Demo 1 below, ECDSA-NIST256P-CHALLENGE mechanism, see Demo 2 below and EXTERNAL, see Demo 3 below.
4. You will need misc/runautotools and misc/makedepend after the fallback base64 is merged.

We need openssl for ECDSA-NIST256P-CHALLENGE, but i didn't find the right (beautiful) solution to if/ifdef yet. Working on it...

We need to make sure we can disable everything for tcl scripts to not be interfered with. Maybe we keep track of state, so we know what cap/sasl we initiated and are waiting for. Then, if we dont initiate in the first place, we wont react to anything.

`.console +d`

**Demo 1 - PLAIN mechanism**
```
.jump 127.0.0.1 6677
[23:23:12] tcl: builtin dcc call: *dcc:jump -HQ 1 127.0.0.1 6677
[23:23:12] #-HQ# jump 127.0.0.1 6677 
Jumping servers......
[23:23:12] Trying server [127.0.0.1]:6677
[23:23:12] CAP: put CAP REQ :sasl
[23:23:12] net: connect! sock 12
[23:23:12] Connected to 127.0.0.1
[23:23:12] -NOTICE- *** Looking up your hostname...
[23:23:12] -NOTICE- *** Checking Ident
[23:23:12] -NOTICE- *** No Ident response
[23:23:12] -NOTICE- *** Found your hostname
[23:23:12] CAP: ACK :sasl 
[23:23:12] zen.localdomain acknowledged sasl 
[23:23:12] SASL: put AUTHENTICATE PLAIN
[23:23:12] SASL: got AUTHENTICATE +
[23:23:12] SASL: put AUTHENTICATE c2FzbHVzZXIAc2FzbHVzZXIAc2FzbHBhc3M=
[23:23:12] SASL: You are now logged in as sasluser.
[23:23:12] SASL: SASL authentication successful
[23:23:13] -NOTICE- *** Notice -- You have an illegal character in your hostname
[23:23:13] triggering bind evnt:init_server
[23:23:13] triggered bind evnt:init_server, user 0.036ms sys 0.054ms
```

**Demo 2 - ECDSA-NIST256P-CHALLENGE mechanism**
```
.jump 127.0.0.1 6677
[23:23:37] tcl: builtin dcc call: *dcc:jump -HQ 1 127.0.0.1 6677
[23:23:37] #-HQ# jump 127.0.0.1 6677 
Jumping servers......
[23:23:37] Trying server [127.0.0.1]:6677
[23:23:37] CAP: put CAP REQ :sasl
[23:23:37] net: connect! sock 12
[23:23:37] Connected to 127.0.0.1
[23:23:37] -NOTICE- *** Looking up your hostname...
[23:23:37] -NOTICE- *** Checking Ident
[23:23:37] -NOTICE- *** No Ident response
[23:23:37] -NOTICE- *** Found your hostname
[23:23:37] CAP: ACK :sasl 
[23:23:37] zen.localdomain acknowledged sasl 
[23:23:37] SASL: put AUTHENTICATE ECDSA-NIST256P-CHALLENGE
[23:23:37] SASL: got AUTHENTICATE +
[23:23:37] SASL: put AUTHENTICATE c2FzbHVzZXIAc2FzbHVzZXI=
[23:23:37] SASL: got AUTHENTICATE JE+FvkBBjqOCfAVA1/fxaLkaGxDXaIcyF8qMZwGkj7s=
[23:23:37] SASL: got AUTHENTICATE Challange
[23:23:37] SASL: put AUTHENTICATE Response MEUCIQDAKilNdXZw/I68IsaYesbNTGJ6SI7+xvCd389BY5gN4AIgZQZFzIcGaYGqxYvKmG1JoutcMn95G8nAoSf8MaOC3E0=
[23:23:37] SASL: You are now logged in as sasluser.
[23:23:37] SASL: SASL authentication successful
[23:23:38] -NOTICE- *** Notice -- You have an illegal character in your hostname
[23:23:38] triggering bind evnt:init_server
[23:23:38] triggered bind evnt:init_server, user 0.027ms sys 0.014ms
```

**Demo 3 - EXTERNAL mechanism**
```
.jump 127.0.0.1 +6697
[22:58:04] tcl: builtin dcc call: *dcc:jump -HQ 1 127.0.0.1 +6697
[22:58:04] #-HQ# jump 127.0.0.1 +6697 
Jumping servers......
[22:58:04] Trying server [127.0.0.1]:+6697
[...]
[22:58:05] CAP: LS :account-notify extended-join identify-msg multi-prefix sasl
[22:58:05] CAP: zen.localdomain supports CAP sub-commands: account-notify extended-join identify-msg multi-prefix sasl
[22:58:05] CAP: Requesting sasl capabilities from server
[22:58:05] CAP: ACK :sasl 
[22:58:05] CAP: Adding cape sasl to negotiated list
[22:58:05] CAP: Current Negotiations sasl with zen.localdomain
[22:58:05] SASL: put AUTHENTICATE EXTERNAL
[22:58:05] SASL: got AUTHENTICATE +
[22:58:05] SASL: put AUTHENTICATE +
[22:58:05] SASL: You are now logged in as BotA.
[22:58:05] SASL: SASL authentication successful
[22:58:14] -NOTICE- *** Notice -- You have an illegal character in your hostname
[22:58:14] triggering bind evnt:init_server
[22:58:14] triggered bind evnt:init_server, user 0.066ms sys 0.008ms
```

Here you can decode base64 online: https://conv.darkbyte.ru/. Most other online converters lack to display chars like \0 but that one has a hex view. though this makes sense only if there is plain text inside.


src/mod/server.mod/servmsg.c
```diff
@@ -1134,6 +1137,147 @@ static int got311(char *from, char *msg)
...
+  FILE *fp;
+  EC_KEY *eckey;
+  EVP_PKEY *privateKey;
+#endif
+
+  putlog(LOG_SERV, "*", "SASL: got AUTHENTICATE %s", msg);
+  if (msg[0] == '+') {
+    s = src;
+    /* Don't use snprintf due to \0 inside */
+    if (!strcmp(sasl_mechanism, "PLAIN")) {
```
grawity: This is inside-out. You should look up the current mechanism first, then check whether received data is valid for that mechanism or not. This will make it easier to support future mechanisms where the server sends non-empty data, as well as multi-step mechanisms, and allows decoupling the individual mechanisms from the core SASL support. For example:

https://github.com/grawity/eggdrop-sasl/blob/master/g_cap.tcl#L254

https://github.com/grawity/eggdrop-sasl/blob/master/g_scram.tcl#L65

src/mod/server.mod/servmsg.c
```diff
@@ -1134,6 +1137,147 @@ static int got311(char *from, char *msg)
...
+      s += strlen(sasl_password);
+    }
+    else if (!strcmp(sasl_mechanism, "ECDSA-NIST256P-CHALLENGE")) {
+      strcpy(s, sasl_username);
+      s += strlen(sasl_username) + 1;
+      strcpy(s, sasl_username);
+      s += strlen(sasl_username);
+    }
+    else /* sasl_mechanism == EXTERNAL */
+      *s++ = '+';
```
grawity: This isn't valid a response for EXTERNAL (unless the server has a user account named "+"). The response needs to either contain the username (like you're doing for PLAIN), or be an empty string.

An empty string only becomes a `+` when it's encoded and sent to the server via AUTHENTICATE (and vice versa, when a `+` is received, it decodes to an empty string).

src/mod/server.mod/servmsg.c
```diff
@@ -1132,6 +1141,115 @@ static int got311(char *from, char *msg)
...
+  putlog(LOG_SERV, "*", "SASL: %s", msg);
+  return 1;
+}
+
+static int got903(char *from, char *msg)
+{
+  newsplit(&msg); /* nick */
+  fixcolon(msg);
+  putlog(LOG_SERV, "*", "SASL: %s", msg);
+  dprintf(DP_MODE, "CAP END\n");
```
vanosg, line 1275: So yes, the SASL spec says to hold off on END until the SASL exchange is completed, but the CAP spec says END should be sent when capability negotiation is done (which I interpret as agnostic from the success or failure of implementing a capability). I think these two statements are at odds with each other. On the bright side, an END sent twice is ignored by the server. We probably need to handle the ENDs in the cap side of the house, so that we can properly track when to send it based on the needs of all requests?

src/mod/server.mod/server.c
```diff
@@ -121,6 +121,15 @@ static void msgq_clear(struct msgq_head *qh);
 static int stack_limit;
 static char *realservername;
 
+static int sasl = 0;
+static int account_notify = 0; //TODO Remove after testing
+static int foober = 1;
+
+static char sasl_mechanism[25];
```
vanosg: static char sasl_mechanism[25] = 0;

default to 0, to match the config setting.

src/mod/server.mod/server.c
```diff
@@ -1386,6 +1400,10 @@ static tcl_strings my_tcl_strings[] = {
   {"connect-server",      connectserver,  120,               0},
   {"stackable-commands",  stackablecmds,  510,               0},
   {"stackable2-commands", stackable2cmds, 510,               0},
+  {"sasl-mechanism",      sasl_mechanism, 24,                0},
+  {"sasl-username",       sasl_username,  NICKMAX,           0},
+  {"sasl-password",       sasl_password,  80,                0},
+  {"sasl-key",            sasl_key,       120,               0},
```
vanosg: sasl-key isn't in config, should this be -ecdsa? (or add a setting in config?)

src/mod/server.mod/servmsg.c
```diff
@@ -1132,6 +1141,115 @@ static int got311(char *from, char *msg)
...
+      strcpy(s, sasl_password);
+      s += strlen(sasl_password);
+    }
+    else if (!strcmp(sasl_mechanism, "ECDSA-NIST256P-CHALLENGE")) {
+      strcpy(s, sasl_username);
+      s += strlen(sasl_username) + 1;
+      strcpy(s, sasl_username);
+      s += strlen(sasl_username);
+    }
+    else /* sasl_mechanism == EXTERNAL */
```
vanosg: Maybe run this as an else if as well, and if something like '4' is added, fail?

src/mod/server.mod/servmsg.c
```diff
@@ -1132,6 +1141,115 @@ static int got311(char *from, char *msg)
...
+  FILE *fp;
+  EC_KEY *eckey;
+  EVP_PKEY *privateKey;
+#endif
+
+  putlog(LOG_SERV, "*", "SASL: got AUTHENTICATE %s", msg);
+  if (msg[0] == '+') {
+    s = src;
+    /* Don't use snprintf due to \0 inside */
+    if (!strcmp(sasl_mechanism, "PLAIN")) {
```
vanosg: We switched this to digits in the config, should be 0, 1, 2?

src/mod/server.mod/servmsg.c
```diff
@@ -1148,6 +1266,138 @@ static int got465(char *from, char *msg)
...
+            putlog(LOG_SERV, "*", "CAP: put CAP REQ :sasl");
+            dprintf(DP_MODE, "CAP REQ :sasl\n");
+            */
+            putlog(LOG_SERV, "*", "SASL: put AUTHENTICATE %s", sasl_mechanism);
+            dprintf(DP_MODE, "AUTHENTICATE %s\n", sasl_mechanism);
+          }
+        }
+      }
+    }
+    dprintf(DP_MODE, "CAP END");
```
vanosg: This is getting sent before the server can reply to finish SASL, thus terminating the negotiation prior to completion and causing an Abort message to be generated.

Maybe the solution lies in a timeout functionality? If no capes requested need negotiation during registration, send the end immediately, otherwise wait up to X seconds for a server response then send the END and move on?

src/mod/server.mod/server.c
michaelortmann: fixed. the code still treated sasl_mechanism as string not enum.

src/mod/server.mod/server.c
michaelortmann: fixed. renamed eggdrop config setting sasl-external into sasl-key

src/mod/server.mod/servmsg.c
michaelortmann: you made the same valid point as grawity. ill leave it as TODO for now.

src/mod/server.mod/servmsg.c
michaelortmann: fixed. the code still treated sasl_mechanism as string not enum.

src/mod/server.mod/servmsg.c
michaelortmann: in my test, sasl plain and sasl ecdsa is working fine. but yes, i will analyze this issue later and leave it as TODO now.

src/mod/server.mod/servmsg.c
michaelortmann: so, nothing to do here and now. you already placed a comment into the code, for future development when more caps will be handled by eggdrop. for now were fine.

src/mod/server.mod/servmsg.c
```diff
@@ -1139,6 +1143,133 @@ static int got311(char *from, char *msg)
...
+      s += strlen(sasl_password);
+      dst[0] = 0;
+      if (b64_ntop((unsigned char *) src, s - src, (char *) dst, sizeof dst) == -1) {
+        putlog(LOG_SERV, "*", "SASL: AUTHENTICATE error: could not base64 encode");
+        /* TODO: send cap end for all error cases in this function ? */
+        return 1;
+      }
+      /* TODO: what about olen we used for mbedtls_base64_encode() ? */
+    }
+    else if (sasl_mechanism == SASL_MECHANISM_ECDSA_NIST256P_CHALLENGE) {
```
vanosg, line 1185: This needs to be ifdef'd out, fp is currently undeclared because the declaration is ifdef'd out above.

src/mod/server.mod/servmsg.c
```diff
@@ -1286,51 +1424,68 @@ static int gotcap(char *from, char *msg) {
      * END. Future eggheads: add support for more than 1 of these async
      * capabilities, right now SASL is the only one so we're OK.
      */
-//    if (strstr(msg, "sasl") != NULL) {
-//      putlog(LOG_SERV, "*", "SASL AUTH CALL GOES HERE!");   //TODO
-//    }
-    if (!strstr(cap.negotiated, "sasl")) {
+    if (strstr(cap.negotiated, "sasl")) {
+      if ((sasl_mechanism != SASL_MECHANISM_ECDSA_NIST256P_CHALLENGE) ||
+          HAVE_OPENSSL_SSL_H) {
```
vanosg: ifdef?

src/mod/server.mod/servmsg.c
```diff
@@ -1286,51 +1425,71 @@ static int gotcap(char *from, char *msg) {
...
      * capabilities, right now SASL is the only one so we're OK.
      */
-//    if (strstr(msg, "sasl") != NULL) {
-//      putlog(LOG_SERV, "*", "SASL AUTH CALL GOES HERE!");   //TODO
-//    }
-    if (!strstr(cap.negotiated, "sasl")) {
+    if (strstr(cap.negotiated, "sasl")) {
+      if ((sasl_mechanism != SASL_MECHANISM_ECDSA_NIST256P_CHALLENGE) 
+#ifdef HAVE_OPENSSL_SSL_H
+        || HAVE_OPENSSL_SSL_H
```
vanosg: Need to handle if sasl-mechanism is 1 (NIST256) but openssl is not on system- either alternate method, or CAP END

src/mod/server.mod/servmsg.c
michaelortmann: Thank you! Fixed with commit 6c48bed8365eeda333b78e106c6231593bf2a9c7.

Issue #3, opened 2010-02-02
Add support for network-autoconfigure using RAW 005


thommey:
| Trac | Data |
| --: | :-- |
| Ticket | 3 |
| Reported by | De Kus |
| Status | assigned |
| Component | Core |
| Priority | blocker |
| Milestone | 1.8.0 |
| Keywords | 005 isupport |
| Version | 1.8.0 CVS |

This actually also involves irc.mod.

I have seen user to struggle with finding the optimum settings and it was often
necessary to patch modes.c and server.c to support new modes. I believe I read
somewhere it was unreliable, but everywhere I connected and checked the
settings, they were pretty much correct. You can learn very much from these
strings. And if you believe some server versions might screw up you have 2
options: user overrides settings in eggdrop.conf, an exception is handled
parsing RAW 004 which contains the server version.

Here are a few examples I collected together for a few server version I believe
to be popular, so you don't have to research on your own. The server version in
() are taken from RAW 004 code. I hope rizons Hybrid patch isn't mature, I
haven't found another server running it atm..
I used PuTTY in RAW protocol as client to be sure to receive raw data:

Unreal (Unreal3.2.6):
:irc.cultvehicles.net 005 De_Kus2 CMDS=KNOCK,MAP,DCCALLOW,USERIP NAMESX
SAFELISTHCN MAXCHANNELS=20 CHANLIMIT=#:20 MAXLIST=b:100,e:100,I:100 NICKLEN=20
CHANNELLEN=32 TOPICLEN=307 KICKLEN=307 AWAYLEN=307 MAXTARGETS=20 :are supported
by this server
:irc.cultvehicles.net 005 De_Kus2 WALLCHOPS WATCH=128 SILENCE=30 MODES=12
CHANTYPES=# PREFIX=(qaohv)~&@%+ CHANMODES=beI,kfL,lj,psmntirRcOAQKVCuzNSMTG
NETWORK=CultvehitlcesIRC CASEMAPPING=ascii EXTBAN=~,cqnr ELIST=MNUCT
STATUSMSG=~&@%+ EXCEPTS :are supported by this server
:irc.cultvehicles.net 005 De_Kus2 INVEX :are supported by this server

Asuka (u2.10.11.07+asuka(1.2.1)):
:port80b.se.quakenet.org 005 DeKus WHOX WALLCHOPS WALLVOICES USERIP CPRIVMSG
[...]

PR #438, opened 2017-08-03
RPL_ISUPPORT parsing, Tcl API and using trivial values

thommey:
Basic RPL_ISUPPORT (IRC raw 005) parsing and Tcl API. Allowed to expand and use the values later, but that is out of the scope of this.

Includes Tcl_Obj API changes that provide a compatibility layer, so we can gradually phase out Tcl's string API.

Fixes #3

PR #694, opened 2018-10-11
(WIP) Add 005 support

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #3

One-line summary:
Start 005 support, find and use NICKLEN

If server sends NICKLEN with 005, eggdrop will use this value instead of default, eggdrop.conf setting or what is hardcoded for net-type setting.

Adapting NICKLEN while server jumping was successful.

### DIFF
--- a/src/mod/server.mod/servmsg.c
+++ b/src/mod/server.mod/servmsg.c
@@ -342,6 +342,23 @@ static int got001(char *from, char *msg)
   return 0;
 }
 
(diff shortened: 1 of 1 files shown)


thommey:
I already have a 005 branch for this, I would prefer to use that as 005 is really not this easy
#438

michaelortmann:
#438

PR #701, opened 2018-10-14
(Fix #457) dns mod: add /etc/hosts lookup before resolving PLUS fix user input validation for .tcl dnslookup

### LINKED ISSUE #457: Check hosts file before DNS lookup?
/etc/hosts has the following entry:
```
172.17.0.2	nosslbot nosslbot eggdrop_botnetnossl
```
yet
```
[21:02:41] #eggdrop# link bot2
[21:02:41] Linking to bot2 at nosslbot:1111 ...
[21:02:41] DNS Resolver: Creating new record
[21:02:41] DNS Resolver: Sent domain lookup request for "nosslbot".
[21:02:41] DNS Resolver: Domain does not exist.
[21:02:41] DNS Resolver: Lookup failed.
[21:02:41] DNS resolve failed for nosslbot
[21:02:41] Failed link to bot2.
```
Should we be checking the hosts file first?

michaelortmann:
Found by: vanosg
Patch by: michaelortmann
Fixes: #457

One-line summary:
dns mod: add a /etc/hosts lookup before resolving PLUS fix user input validation for .tcl dnslookup

As discussed at IRC, we want the /etc/hosts file to be read every time, and not be kept in memory and we prefer maintainability over optimized code
**Optionally** we could add a config setting for a user defined hosts file instead of /etc/hosts.
**Optionally** we could add a config setting for disabling of /etc/hosts.
( That is no requirement for eggdrop 1.9.0, we could add such options later. )

Eggdrops dns mod is based on dns.c from mtr. when mtr had that problem with /etc/hosts they replaced the dns code with new code, that forks, see: https://github.com/traviscross/mtr/issues/78. I don't think we should go this route with eggdrops dns.mod, but its interesting. **Update:** we should go this route, but in core instead of dns.mod, and with threads instead of fork(), see #1015.

Finally let me explain what i mean by "PLUS fix user input validation for .tcl dnslookup":

The user command `.tcl dnslookup` did not validate the user input hostname. if the hostname was longer than 255 chars, which is forbidden by RFC1035, it was still passed on, send to dns server or with this PR searched for in /etc/hosts. So, this PR adds a hostname length validation and rejects too long hostnames with a meaningful error message. **Update:** With this PR we will additionally reject hostnames with bogus chars.

```
$ grep freebsd /etc/hosts
192.168.0.16    freebsd.localdomain      freebsd

.console +d
```
Before:
```
.link freebsd
[20:08:34] tcl: builtin dcc call: *dcc:link -HQ 1 freebsd
[20:08:34] #-HQ# link freebsd
[20:08:34] Linking to freebsd at freebsd:12345 ...
[20:08:34] DNS Resolver: Creating new record
[20:08:34] DNS Resolver: Sent domain lookup request for "freebsd".
[20:08:34] DNS Resolver: Domain does not exist.
[20:08:34] DNS Resolver: Lookup failed.
[20:08:34] DNS resolve failed for freebsd
[20:08:34] Failed link to freebsd.
```
After:
```
.link freebsd
[20:04:19] tcl: builtin dcc call: *dcc:link -HQ 1 freebsd
[20:04:19] #-HQ# link freebsd
[20:04:19] Linking to freebsd at freebsd:12345 ...
[20:04:19] DNS Resolver: Used /etc/hosts: freebsd == 192.168.0.16
[20:04:19] Got STARTTLS from freebsd. Replying...
[...]
[20:04:19] Linked to freebsd.
```
test with too long hostname:
```
.tcl dnslookup a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890a234567890 foo
Tcl error: hostname too long. max 255 chars.
```

test with bogus char in hostname input:
```
.tcl dnslookup abc%def foo
[15:56:19] DNS Resolver: bogus char in hostname input
```

Tiny developer tests under cygwin (windows), freebsd and with --disable-ipv6 were also successful.

You could also test with `.tcl dnslookup <name> foo` instead of `.link`


src/mod/dns.mod/coredns.c
```diff
@@ -1220,6 +1220,36 @@ static void dns_forward(char *hostn)
     }
     return;
   }
+
+  FILE* in;
```
thommey: not C89 compatible, declaration not at beginning of block

thommey (changes requested):

src/mod/dns.mod/coredns.c
michaelortmann: Fixed.

michaelortmann:
TODO: strsep() ist not C89 and not POSIX.2001, so i need to replace it.

michaelortmann:
Ready for review again.

src/mod/dns.mod/coredns.c
```diff
@@ -1196,6 +1196,42 @@ static void dns_lookup(sockname_t *addr)
   sendrequest(rp, T_PTR);
 }
 
+static int dns_hosts(char *hostn) {
+  FILE* f;
+  char line[1024];
```
thommey: is getline() too GNU specific?

src/mod/dns.mod/coredns.c
```diff
@@ -1196,6 +1196,42 @@ static void dns_lookup(sockname_t *addr)
...
+  f = fopen("/etc/hosts", "rb");
+  if (!f) {
+    ddebug1(RES_MSG "fopen(): %s ", strerror(errno));
+    return 0;
+  }
+  while (fgets(line, sizeof line , f)) {
+    if ((ptr = strchr(line, '#')))
+      *ptr = '\0';
+    for(ptr = strstr(line, hostn); ptr; ptr = strstr(ptr + l, hostn)) {
+      if ((isspace(ptr[l]) || !ptr[l]) && ptr != line) {
```
thommey: Is `ptr != line` intended to exclude it as the first word? then this would need space-skipping too. Maybe space-skip "line" as a whole before the loop?

src/mod/dns.mod/coredns.c
michaelortmann: No, this has nothing to do with GNU but with eggdrops POSIX 2001 compatibility. getline() is not posix 2001 but posix 2008. see http://www.unix.org/version3/inttables.pdf and/or https://www.freebsd.org/cgi/man.cgi?query=getline&apropos=0&sektion=0&manpath=FreeBSD+12.0-RELEASE+and+Ports&arch=default&format=html.

src/mod/dns.mod/coredns.c
michaelortmann: Yes and yes.

michaelortmann: btw, strstr(ptr + l) looks like a bug.

src/mod/dns.mod/coredns.c
```diff
@@ -1211,6 +1211,44 @@ static void dns_lookup(sockname_t *addr)
...
+    ddebug1(RES_MSG "fopen(): %s ", strerror(errno));
+    return 0;
+  }
+  while (fgets(line, sizeof line , f)) {
+    ptr1 = line;
+    while (isspace(*ptr1))
+      ptr1++;
+    if ((ptr2 = strchr(ptr1, '#')))
+      *ptr2 = '\0';
+    for(ptr2 = strstr(ptr1, hostn); ptr2; ptr2 = strstr(ptr2 + l, hostn)) {
```
michaelortmann: strstr(ptr2 + l, ...) looks like a bug

michaelortmann: host names are not case sensitive, another bug

michaelortmann: i replaced the parser...

michaelortmann:
New Parser, hopefully less buggy, ready for review again.

michaelortmann:
New Parser, faster and hopefully less buggy, ready for review again.

michaelortmann:
New Parser, faster and hopefully less buggy, **ready for review again.**

vanosg:
Tested both with and without dns module loaded; with ipv4 and ipv6 addresses; and multiple hosts for one IP. Looks good, functionally

michaelortmann:
@thommey:  Changed parser to bookmark last seen newline and hash chars instead of backwards searching for them. Ready for review again.

michaelortmann:
@thommey: Changed parser to bookmark last seen newline and hash chars instead of backwards searching for them in an attempt to ease review / readability / maintainability. Ready for review again.

michaelortmann:
@thommey: Changed parser to bookmark last seen newline and hash chars instead of backwards searching for them in an attempt to ease review / readability / maintainability. Ready for review again.

michaelortmann:
@thommey: Changed parser to bookmark last seen newline and hash-chars instead of backwards searching for them in an attempt to ease review / readability / maintainability. Ready for review again.

src/mod/dns.mod/coredns.c
```diff
@@ -1201,6 +1204,103 @@ static void dns_lookup(sockname_t *addr)
...
+    return 1;
+  }
+  /* due to strncasecmp() and strncmp() being slow if used in loop, precalculate
+   * lower and upper string from hostn and compare with handcrafted code */
+  for (i = 0; i < len; i++) {
+      if (isspace(hostn[i])) {
+        ddebug0(RES_MSG "bogus white-space hostname input");
+        return 1;
+      }
+      hostn_lower[i] = tolower(hostn[i]);
```
thommey: `tolower((unsigned char)hostn[i])`, same with toupper, as described in man page

src/mod/dns.mod/coredns.c
```diff
@@ -1201,6 +1204,103 @@ static void dns_lookup(sockname_t *addr)
...
+      hostn_upper[i] = toupper(hostn[i]);
+  }
+  hostn_lower[i] = 0;
+  hostn_upper[i] = 0;
+  fd = open(PATH, O_RDONLY);
+  if (fd < 0) {
+    ddebug1(RES_MSG "open(" PATH "): %s ", strerror(errno));
```
thommey: remove empty space at the end please, for all those strerror messages

src/mod/dns.mod/coredns.c
```diff
@@ -1201,6 +1204,103 @@ static void dns_lookup(sockname_t *addr)
...
+    close(fd);
+    return 0;
+  }
+  addr = mmap(NULL, sb.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
+  if (addr == MAP_FAILED) {
+    ddebug1(RES_MSG "mmap(" PATH "): %s ", strerror(errno));
+    close(fd);
+    return 0;
+  }
+  last_newline = addr - 1;
```
thommey: this calculates a pointer one before an allocated space, it's undefined behavior

src/mod/dns.mod/coredns.c
```diff
@@ -1201,6 +1204,103 @@ static void dns_lookup(sockname_t *addr)
...
+        last_newline = c;
+        break;
+      case '#':
+        last_hash = c;
+        break;
+      default:
+        /* if (!strncasecmp(c, hostn, len)) { */
+        for (i = 0; (i < len) && (c[i] == hostn_lower[i] || (c[i] == hostn_upper[i])); i++);
+        if ((i == len) &&
+            ((c == (addr + sb.st_size - len - 1)) || isspace(*(c + len))) &&
```
thommey: this can again be invalid (it ca be addr - 1). Example:
file "ex" and we are looking for "ex", addr points to "e", sb.st_size is 2, len is 2: addr + 2 - 2 - 1 => addr - 1 which is undefined behavior to calculate.
Is it correct? I think you intend to check if it's the end of the file, addr - 1 would end up wrong.
isspace needs the unsigned char cast as well

thommey (changes requested):
Can't we just read the file line-by-line with fgets or readline? memory mapping a file under 1 KB seems like an optimization we don't need at the expense of code readability. You have to check if you're at the file end, if it's a comment, where the last newline is, etc. which would all be a lot easier with fgets/readline

src/mod/dns.mod/coredns.c
michaelortmann: as we "agreed" we read this file (which can be bigger than 1k ;) for every dns lookup, not caching (looking at the files modification time). so i thought it would be worth using mmap. but yes, i will change it to using line by line reading now. esp. since the future will hopefully bring async dns with #1015 and make this dns mod code obsolete.

michaelortmann:
as we "agreed" we read this file (which can be bigger than 1k ;) for every dns lookup, not caching (looking at the files modification time). so i thought it would be worth using mmap. but yes, i will change it to using line by line reading now. esp. since the future will hopefully bring async dns with #1015 and make this dns mod code obsolete.

src/mod/dns.mod/coredns.c
michaelortmann: Fixed trailing whitespace in debug log

src/mod/dns.mod/coredns.c
michaelortmann: im sure we can do this "hack" in C. but if you still insist, i will change the code.

michaelortmann: Fixed pointer arithmetic hack

src/mod/dns.mod/coredns.c
michaelortmann: Fixed tolower() and toupper() usage. This was stupid me.

src/mod/dns.mod/coredns.c
michaelortmann: Fixed isspace() by using the already existing egg_isspace() macro if we need to check for space and newline and hostn_isspace() if we only need to check for space. Fixed eof.

michaelortmann:
Meanwhile i committed a new implementation.

This time fgets() is used instead of mmap() because we prefer maintainability over optimized code, even though we read /etc/hosts for every lookup and /etc/hosts can be big, for example if used for stopping banner ads. Example: https://pgl.yoyo.org/adservers/serverlist.php?hostformat=hosts;mimetype=plaintext

I benchmarked this new implementation on a 20 year old machine with an amd k6-ii 400 cpu under freebsd 12.1. I took the host file from the just mentioned example and feeded a hostname to the parser, that is not in that file, so it must parse the whole file. Parsing the whole file took 13ms on that 20 year old machine. So, for me its fast enough to not introduce a new config option to disable /etc/hosts parsing for that reason.

This new implementation was also successfully tested with prefer-ipv6.

michaelortmann:
Meanwhile i committed a new implementation.

This time fgets() is used instead of mmap() because we prefer maintainability over optimized code, even though we read /etc/hosts for every lookup and /etc/hosts can be big, for example if used for stopping banner ads. Example: https://pgl.yoyo.org/adservers/serverlist.php?hostformat=hosts;mimetype=plaintext

I benchmarked this new implementation on a 20 year old machine with an amd k6-ii 400 cpu under freebsd 12.1. I took the host file from the just mentioned example (currently 3557 lines) and feeded a hostname to the parser, that is not in that file, so it must parse the whole file. Parsing the whole file took 13ms on that 20 year old machine. So, for me its fast enough to not introduce a new config option to disable /etc/hosts parsing for that reason.

This new implementation was also successfully tested with prefer-ipv6.

Ready for review again. Please test thoroughly.

michaelortmann:
Im still doing some cleanup/work on this...

michaelortmann:
Meanwhile i committed a new implementation.

This time fgets() is used instead of mmap() because we prefer maintainability over optimized code, even though we read /etc/hosts for every lookup and /etc/hosts can be big, for example if used for stopping banner ads. Example: https://pgl.yoyo.org/adservers/serverlist.php?hostformat=hosts;mimetype=plaintext

I benchmarked this new implementation on a 20 year old machine with an amd k6-ii 400 cpu under freebsd 12.1. I took the host file from the just mentioned example (currently 3557 lines) and feeded a hostname to the parser, that is not in that file, so it must parse the whole file. Parsing the whole file took 13ms on that 20 year old machine. So, for me its fast enough to not introduce a new config option to disable /etc/hosts parsing for that reason.

This new implementation was also successfully tested with prefer-ipv6.

Upfront tab/space is also ignored (*wave* at thommey: postel's law applied).

Currently a debug print is issued, if /etc/hosts lookup was successful. This could be changed into a regular putlog(LOG_MISC). Let me know, if you really want this. Example:
`[05:03:09] DNS Resolver: Used /etc/hosts: jupiter == fd9f:de9c:36c5:eda1::1`

Ready for review again. Please test thoroughly.

michaelortmann:
Meanwhile i committed a new implementation.

This time fgets() is used instead of mmap() because we prefer maintainability over optimized code, even though we read /etc/hosts for every lookup and /etc/hosts can be big, for example if used for stopping banner ads. Examples: https://pgl.yoyo.org/adservers/serverlist.php?hostformat=hosts;mimetype=plaintext https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts. We need to care, other well known projects have / had performance issues with /etc/hosts: https://github.com/ohmyzsh/ohmyzsh/issues/7397 https://github.com/fish-shell/fish-shell/issues/4604.

I benchmarked this new implementation on a 20 year old machine with an amd k6-ii 400 cpu under freebsd 12.1. I took the host file, feeded a hostname to the parser, that is not in that file, so it must parse the whole file. I did that in a loop with 1000 iterations to get accurate timing. Parsing the whole file took:

13ms for https://pgl.yoyo.org/adservers/serverlist.php?hostformat=hosts;mimetype=plaintext
219ms https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts

In case of 219ms, an additional config option to disable /etc/hosts parsing could be helpful.

Current systems, like Ryzen 3700x take only about 0.2ms for first hostfile and 3ms for second hostfile, but we should make sure, older systems are working fine also.

This new implementation was also successfully tested with prefer-ipv6.

Upfront tab/space is also ignored (*wave* at thommey: postel's law applied).

Currently a debug print is issued, if /etc/hosts lookup was successful. This could be changed into a regular putlog(LOG_MISC). Let me know, if you really want this. Example:
`[05:03:09] DNS Resolver: Used /etc/hosts: jupiter == fd9f:de9c:36c5:eda1::1`

Ready for review again. Please test thoroughly. And let me know what you think about an additional "disable /etc/hosts parsing" option.

src/mod/dns.mod/coredns.c
```diff
@@ -1201,6 +1202,109 @@ static void dns_lookup(sockname_t *addr)
...
 
+/* Read /etc/hosts. fgets() is used instead of mmap() because we prefer
+ * maintainability over optimized code, even though we read /etc/hosts for every
+ * lookup and /etc/hosts can be big, for example if used for stopping banner
+ * ads */
+static int dns_hosts(char *hostn) {
+  #define PATH "/etc/hosts"
+  size_t hostn_len;
+  int i;
+  char hostn_lower[256], hostn_upper[256], line[1024], *p1, *p2, *p3, *p4;
```
thommey, line 1210: If we are concerned about /etc/hosts being used for massive dns blocking, we should increase the line buffer size. My gut feeling would be 16 KB. Or at least warn about it if fgets() returns max length (likely truncated) so we can get reports from this happening.

src/mod/dns.mod/coredns.c
```diff
@@ -1201,6 +1202,109 @@ static void dns_lookup(sockname_t *addr)
   sendrequest(rp, T_PTR);
 }
 
+/* Read /etc/hosts. fgets() is used instead of mmap() because we prefer
+ * maintainability over optimized code, even though we read /etc/hosts for every
+ * lookup and /etc/hosts can be big, for example if used for stopping banner
+ * ads */
```
thommey: The real optimization in that case would be caching file content based on mtime, not mmap

thommey (approved):
LGTM but bump line length please (and/or add warning)

src/mod/dns.mod/coredns.c
michaelortmann: yes, and if you what me to implement that, let me know.

src/mod/dns.mod/coredns.c
michaelortmann: Before i hardcoded that 1024 i checked other implementations. Like for example in FreeBSD:
https://github.com/freebsd/freebsd/blob/1d6e4247415d264485ee94b59fdbc12e0c566fd0/libexec/mknetid/mknetid.c#L283
https://github.com/freebsd/freebsd/blob/1d6e4247415d264485ee94b59fdbc12e0c566fd0/lib/libc/net/getaddrinfo.c#L2403
Those operating system implementations also seem to limit the bufsize to 1024 or 8*1024 without any error checking.
So i would like to stick with 1024 here.
But you are right, i forgot to add the ferror() check for fgets(). (Yes, operating systems dont check that either, but we can and we should ;)
Now that this PR is merged, i will probably add a small follow-up PR to add that fgets() ferror() check.

PR #702, opened 2018-10-14
Fix heap bug

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix heap bug

This bug could have made eggdrop fail on hardened systems.
The patch also adds error check for X509_NAME_print_ex()
and enhances performance by some cpu cycles ;)

Test with sharebots over ssl was successful, which tests at least the changed functions but also gives good overall impression.
The following eggdrop debug output is correct:
`[18:47:31] TLS: certificate subject: C=EU, O=Eggheads, OU=Botnet, UID=sslbot`


thommey (dismissed):
// is C99, C89 only has /* */ comments, otherwise the PR is fine

michaelortmann:
@thommey: damn, i did it again. fixed now.

src/tls.c
```diff
@@ -422,10 +422,17 @@ static char *ssl_printname(X509_NAME *name)
...
-  len = BIO_get_mem_data(bio, &data) + 1;
-  buf = nmalloc(len);
-  strlcpy(buf, data, len);
+  if (X509_NAME_print_ex(bio, name, 0, XN_FLAG_ONELINE & ~XN_FLAG_SPC_EQ)) {
+    len = BIO_get_mem_data(bio, &data);
+    buf = nmalloc(len + 1);
+    memcpy(buf, data, len); /* don't strlcpy() for it would read data[len] */
+    buf[len] = 0;
+  }
+  else {
```
Cizzle: Combine with line above.

src/tls.c
```diff
@@ -422,10 +422,17 @@ static char *ssl_printname(X509_NAME *name)
 
   /* X509_NAME_oneline() is easier and shorter, but is deprecated and
      the manual discourages it's usage, so let's not be lazy ;) */
-  X509_NAME_print_ex(bio, name, 0, XN_FLAG_ONELINE & ~XN_FLAG_SPC_EQ);
-  len = BIO_get_mem_data(bio, &data) + 1;
-  buf = nmalloc(len);
-  strlcpy(buf, data, len);
+  if (X509_NAME_print_ex(bio, name, 0, XN_FLAG_ONELINE & ~XN_FLAG_SPC_EQ)) {
+    len = BIO_get_mem_data(bio, &data);
```
Cizzle, line 426: According to docs, this returns a `long` so let's change it while we're at it. It also seems to be signed and though it's unclear if it can get negative, please add a check below.
Same comment for the same codeblocks lower.

Cizzle (changes requested):

src/tls.c
```diff
@@ -416,16 +416,30 @@ static int ssl_verifycn(X509 *cert, ssl_appdata *data)
...
+    if (len > 0) {
+      buf = nmalloc(len + 1);
+      memcpy(buf, data, len); /* don't strlcpy() for it would read data[len] */
+      buf[len] = 0;
+    }
+    else {
```
Cizzle: Please combine with line above. Same for other blocks.

Cizzle (approved):

Cizzle:
LGTM

vanosg:
build

Issue #703, opened 2018-10-14
Compiler Warnings


michaelortmann:
There were/are many compiler warnings, esp. with new versions of gcc. This issue is for coordinating analysis and work to each of them. Some of them are already fixed by (pending) PRs.

The following findings are with current (moving) git version of eggdrop.

**11. 64-bit NetBSD 6.1.5 gcc 4.5.3 Tcl 8.6.8 OpenSSL 1.0.1i CFLAGS default**

Tested eggdrop version: git commit 713cedfa5cd561211b8869d2e9ce8dad0d4d1d91 Oct 17 2018

```
gcc -fPIC -g -O2 -pipe -Wall -I. -I../../.. -I../../..  -I../../../src/mod  -DHAVE_CONFIG_H -I/usr/pkg/include -g3 -DDEBUG -DDEBUG_ASSERT -DDEBUG_MEM -DDEBUG_DNS  -DMAKING_MODS -c .././uptime.mod/uptime.c && mv -f uptime.o ../
.././uptime.mod/uptime.c: In function 'send_uptime':
.././uptime.mod/uptime.c:226:3: warning: overflow in implicit constant conversion
```

This will be fixed by #548

**13. FreeBSD 6.4-RELEASE gcc 3.4.6 Tcl 8.4.19 default cflags**

```
gcc -g -O2 -pipe -Wall -I.. -I..  -DHAVE_CONFIG_H -I/usr/local/include/tcl8.4 -g3 -DDEBUG -DDEBUG_ASSERT -DDEBUG_MEM -DDEBUG_DNS  -c tcl.c
tcl.c:423: warning: dereferencing type-punned pointer will break strict-aliasing rules
gcc -g -O2 -pipe -Wall -I.. -I..  -DHAVE_CONFIG_H -I/usr/local/include/tcl8.4 -g3 -DDEBUG -DDEBUG_ASSERT -DDEBUG_MEM -DDEBUG_DNS  -c userent.c
userent.c: In function `del_entry_type':
userent.c:1395: warning: dereferencing type-punned pointer will break strict-aliasing rules
gcc -fPIC -g -O2 -pipe -Wall -I. -I../../.. -I../../..  -I../../../src/mod  -DHAVE_CONFIG_H -I/usr/local/include/tcl8.4 -g3 -DDEBUG -DDEBUG_ASSERT -DDEBUG_MEM -DDEBUG_DNS  -DMAKING_MODS -c .././server.mod/server.c && mv -f server.o ../
[...]

PR #704, opened 2018-10-14
Define sizeof userfile at the right place

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Define sizeof userfile at the right place

Fixes compiler warning (3) in #703
Fixes magic number in commit 9e6f2790f10dff7f38cbfeaf30acb2e14e89b517
Add some comments


src/userrec.c
```diff
@@ -532,7 +532,7 @@ void sort_userlist()
 void write_userfile(int idx)
 {
   FILE *f;
-  char *new_userfile;
+  char new_userfile[sizeof userfile + 4]; /* 4 = sizeof "~new" */
```
thommey: at least the comment is misleading, 4 = strlen("~new"), 5 = sizeof("~new")

thommey (changes requested):

src/userrec.c
michaelortmann: ty, fixed.

src/main.c
```diff
@@ -615,7 +614,7 @@ static void do_arg()
 
 void backup_userfile(void)
 {
-  char s[sizeof userfile + 4];
+  char s[sizeof userfile + 4]; /* 4 = sizeof "~bak" */
```
Cizzle: Same comment applies here.

michaelortmann: Done.

thommey:
Is defining it in a headerfile really the way to go? It seems unconventional and could define the symbol in multiple translation units without them being related. I don't think this is the right general solution for the issue of all those char arrays.

michaelortmann:
Thank you thommey, i committed a new/better solution, where i moved the function backup_userfile() from main.c to userrec.c, **ready for review again**.

thommey (approved):

PR #709, opened 2018-10-15
Added a note regarding source path

michaelortmann:
Found by: jackal^
Patch by: simple, Robby, michaelortmann
Fixes: 

One-line summary:
Added a note regarding source path and removed too obvious time wasting statement

The issue was:
jackal^> so im in ~, if i git clone, i end up with ~/eggdrop. if i ./configure, i end up with eggdrop binary in source dir

### DIFF
--- a/README
+++ b/README
@@ -100,7 +100,7 @@ GIT USAGE
 
     https://github.com/eggheads/eggdrop/archive/develop.tar.gz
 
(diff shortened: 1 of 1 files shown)


michaelortmann:
Discussion on irc resulted in:
Close this issue, because it's already documented:
https://github.com/eggheads/eggdrop/blob/develop/Makefile.in#L349

PR #714, opened 2018-10-16
Fix 2 Wformat-truncation (gcc 8.2.1 default cflags)

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix 2 Wformat-truncation (gcc 8.2.1 default cflags)

Resized a buffer to be undersized due to sane assumption that struct tm is sane (i mean, we already assumed tm->tm_min would be 0-59, we continue to do so) and at the same time oversized to silence a gcc format-truncation warning (i mean, gcc already seems to assume that struct tm is sane, but it looses its grip, when we do math, like tm->tm_year + 1900).

We could solve this in different ways, but i don't want to waste cpu cycles over it, so resetting the buffer size is it.

While at it, combine 2 functions into 1. It safes some more cpu/bytes in mainloop/check_minutely. So to optimize here should be reasonable. IRC chat with thommey was fine about it.

```
gcc -g -O2 -pipe -Wall -I.. -I..  -DHAVE_CONFIG_H -I/usr/include -g3 -DDEBUG -DDEBUG_ASSERT -DDEBUG_MEM -DDEBUG_DNS  -c tclhash.c
tclhash.c: In function ‘check_tcl_time’:
tclhash.c:1199:50: warning: ‘%04d’ directive output may be truncated writing between 4 and 11 bytes into a region of size between 0 and 6 [-Wformat-truncation=]
   egg_snprintf(y, sizeof y, "%02d %02d %02d %02d %04d", tm->tm_min, tm->tm_hour,
                                                  ^~~~
tclhash.c:1199:29: note: directive argument in the range [-2147481748, 2147483647]
   egg_snprintf(y, sizeof y, "%02d %02d %02d %02d %04d", tm->tm_min, tm->tm_hour,
                             ^~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from compat/compat.h:27,
                 from main.h:102,
                 from tclhash.c:31:
compat/snprintf.h:58:24: note: ‘snprintf’ output between 17 and 60 bytes into a destination of size 18
 #  define egg_snprintf snprintf
tclhash.c:1199:3: note: in expansion of macro ‘egg_snprintf’
   egg_snprintf(y, sizeof y, "%02d %02d %02d %02d %04d", tm->tm_min, tm->tm_hour,
   ^~~~~~~~~~~~
tclhash.c: In function ‘check_tcl_cron’:
tclhash.c:1220:45: warning: ‘%02d’ directive output may be truncated writing between 2 and 11 bytes into a region of size between 0 and 6 [-Wformat-truncation=]
   egg_snprintf(y, sizeof y, "%02d %02d %02d %02d %02d", tm->tm_min, tm->tm_hour,
                                             ^~~~
tclhash.c:1220:29: note: directive argument in the range [-2147483647, 2147483647]
   egg_snprintf(y, sizeof y, "%02d %02d %02d %02d %02d", tm->tm_min, tm->tm_hour,
                             ^~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from compat/compat.h:27,
                 from main.h:102,
                 from tclhash.c:31:
compat/snprintf.h:58:24: note: ‘snprintf’ output between 15 and 60 bytes into a destination of size 15
 #  define egg_snprintf snprintf
tclhash.c:1220:3: note: in expansion of macro ‘egg_snprintf’
   egg_snprintf(y, sizeof y, "%02d %02d %02d %02d %02d", tm->tm_min, tm->tm_hour,
   ^~~~~~~~~~~~
```


Cizzle:
I don't see how this fixes the warnings?

michaelortmann:
thanks, irc discussion resulted in %04d won't truncate to 4 chars, but set a min length instead. so it seems the compiler warning was indeed correct. on the other hand it wasnt, i mean, gcc was trying to silence its own warning over assuming sane values in tm struct, as we, the humans do, but it lost its grip over us doing calculations like year + 1900 on it. My new solution is committed and commented on top of this PR.

Btw, we cant use strftime() instead of snprintf() here, because strftime() for example cant do tm->tm_mon + 1.

Cizzle:
LGTM

vanosg:
build

PR #717, opened 2018-10-17
Fix 1 Wformat-overflow

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix 1 Wformat-overflow with gcc 8.2.1 and CFLAGS -fsanitize=undefined

First i thought we should replace nmalloc(size) with buffer[size] because nmalloc(size) was called with a constant size of 15.
Then i understood that the code block wants to find out the size of the final string to allocate and put in a structure.
So i recoded that part, to clearly show the intention by using the pattern
```
len = snprintf(null, 0, ...)
s = malloc(len + 1)
sprintf(s, ...)
```

```
userrec.c: In function ‘adduser’:
userrec.c:644:5: warning: null destination pointer [-Wformat-overflow=]
     sprintf(now2, "%li", tv);
     ^~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/userrec.c
+++ b/src/userrec.c
@@ -635,17 +635,15 @@ struct userrec *adduser(struct userrec *bu, char *handle, char *host,
   }
   set_user(&USERENTRY_PASS, u, pass);
   if (!noxtra) {
-    char *now2;
+    int l;
     xk = nmalloc(sizeof *xk);
     xk->key = nmalloc(8);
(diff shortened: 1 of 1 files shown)


Cizzle:
LGTM

vanosg:
build

PR #721, opened 2018-10-18
Cleanup opening of logfile

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Don't need the + in a+ here, it could mislead the reader that we also read from a logfile, which we don't.


Cizzle:
LGTM

PR #722, opened 2018-10-18
Fix Wstringop-truncation and Wformat-overflow

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix Wstringop-truncation and Wformat-overflow

The compiler was noisy about a code pattern like
`[egg_]s[n]printf() + str[n]cat()"
that hereby is cleaned up and also more len and null termination safe.


Cizzle:
LGTM

vanosg:
build

PR #726, opened 2018-10-20
Make userrec:expmem_mask() static

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Make userrec:expmem_mask() static

We should always use static if possible
Fix part of #176
http://www.sisyphus.ru/en/srpm/Sisyphus/eggdrop/patches/1
I know, that patch looks like one of the usual "remove inline keyword" patches,
but i also found the +static in there

A small test compiling, starting, share linking, server joining, was successful.

### DIFF
--- a/src/userrec.c
+++ b/src/userrec.c
@@ -74,7 +74,7 @@ void *_user_realloc(void *ptr, int size, const char *file, int line)
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


Cizzle:
LGTM

vanosg:
build

PR #728, opened 2018-10-22
 \e isn't C standard so we use \x1B instead (make error)

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
 \e isn't C standard so we use \x1B instead

commit 1e28068f4d1c00d5e9114b27c36ee15fdf96a494 **broke some compilers**
gcc, clang and tcc are fine, but others like nwcc are not.
see https://en.wikipedia.org/wiki/Escape_sequences_in_C#Non-standard_escape_sequences

```
$ CC=nwcc make
[...]
nwcc -march=native -O2 -pipe -fstack-protector-strong -fno-plt -fomit-frame-pointer -I.. -I..  -DHAVE_CONFIG_H -I/usr/include  -c dcc.c
dcc.c:1561: Error: Invalid escape sequence.
  if (!strncmp(buf, "\e[?1;2c", 7))
                      ^ here
make[1]: *** [Makefile:82: dcc.o] Error 1
```

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -1558,7 +1558,7 @@ static void dcc_telnet_id(int idx, char *buf, int atr)
     return;
   }
   /* rxvt-unicode */
-  if (!strncmp(buf, "\e[?1;2c", 7))
+  if (!strncmp(buf, "\x1B[?1;2c", 7)) /* \e isn't C standard so we use \x1B instead */
     buf += 7;
   dcc[idx].user = get_user_by_handle(userlist, buf);
   get_user_flagrec(dcc[idx].user, &fr, NULL);


Cizzle:
From that wiki page:
> However, these escape sequences can be used on any system with a C compiler, and may map to different values if the system does not use a character encoding based on ASCII.

So it's possible this `\x1B` maps to something differently than `<ESC>`?

michaelortmann:
could be, but its stretched here. if you take a look at the context. what could happen.  we're talking about a string match of some chars, of a string: "\e[?1;2c"

in any other case we do what we did before. so were safe here. if esc isnt 1B we miss it, but then we do as we did before this patch.

Cizzle:
LGTM

vanosg:
build

PR #730, opened 2018-10-22
Fix nwcc compiler error

### LINKED ISSUE #729: Make error with nwcc compiler
Yesterday, while i played with compiler parameters, i think clang, i found a compiler warning for
```
coredns.c:
  if (0 > response + len) {
```

This code is odd. The compiler is right about that.

Today, thanks to goddess Coincidencia, i found out that this is not only a warning but also an error with other compiler, like nwcc 0.8.3:

```
$ CC=nwcc make
[...]
nwcc -fPIC -march=native -O2 -pipe -fstack-protector-strong -fno-plt -fomit-frame-pointer -I. -I../../.. -I../../.. -I../../../src/mod  -DHAVE_CONFIG_H -I/usr/include   -DMAKING_MODS -c .././dns.mod/dns.c && mv -f dns.o ../
.././dns.mod/coredns.c:975: Error: Cannot use pointer type with `>' operator and basic type
    if (0 > response + len) {
          ^ here
/var/tmp/cpp5.cpp - 1 error(s), 0 warning(s)
```

After further analysis, i think the line
`  if (0 > response + len) {`
should be
`  if (c > response + len) {`

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #729

One-line summary:
Fix nwcc compiler error

In my test, if a resolve was successful, c was equal to response + len.
So im confident this is the real solution.
I tested with additional debug print inside code that you won't see here.
What you could see is that the compiler error with nwcc is gone.

### DIFF
--- a/src/mod/dns.mod/coredns.c
+++ b/src/mod/dns.mod/coredns.c
@@ -972,7 +972,7 @@ void parserespacket(uint8_t *response, int len)
       return;
     }
     c += 10 + rr->datalength;
-    if (0 > response + len) {
+    if (c > response + len) {
       ddebug0(RES_ERR "Specified rdata length exceeds packet size.");
       return;
     }


Cizzle:
LGTM

vanosg:
build

PR #756, opened 2018-11-06
Remove compat  str(n)casecmp (eggdrop requires posix 2001)

michaelortmann:
Found by: vanosg
Patch by: michaelortmann
Fixes:

One-line summary:
Fixed part of #226.
Eggdrop depends on POSIX.1-2001. POSIX.1-2001 defines strcasecmp() and strncasecmp(). This patch gets rid of src/compat/strcasecmp.c and src/compat/strcasecmp.h. egg_str(n)casecmp() are left as is, but redefined to directly use str(n)casecmp() with a one-liner define in eggdrop.h. we only keep them for backward compatibility to (older) modules, and could remove it from later eggdrop versions when api breakage becomes possible.

All  egg_str(n)casecmp() were replaced by str(n)casecmp():

find ./ -type f -exec sed -i 's/egg_strcasecmp(/strcasecmp(/g' {} \;
find ./ -type f -exec sed -i 's/egg_strncasecmp(/strncasecmp(/g' {} \;

**Note 1: don't do this in git directory. cd src/ first!**

Note 2: github isn't displaying those commands in full, you need to go to edit mode, to see the backslash in there. if you don't, it wont work, but it also won't also cost you a life.

Please run misc/runautotools after merging this patch.

I did a small test, linked 2 share bots, watched them share the userfile, joined a server and did some tcl commands. So far, so good.


michaelortmann:
My last commit message should read `replaced all egg_str(n)casecmp() with str(n)casecmp() with sed`.

michaelortmann:
The reason i moved ctype.h is that this PR removes src/compat/strcasecmp.h, which includes ctype.h. After removal, some source files (still) need, but don't (indirectly) include ctype.h, including eggdrop.h. So i added ctype.h to eggdrop.h. After that i cleaned up some includes of ctype.h which are no longer needed.

PR #768, opened 2018-11-19
Cleanup src/net.c

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Cleanup src/net.c

1. sizeof "255.255.255.255" -> INET_ADDRSTRLEN
2. The comment "look for \r too cos windows can't follow RFCs" was wrong, was the opposite of what the RFC says
3. Fix leak, thanks Cizzle!
4. Enhanced strdup like code 
5. Misc. cleanup

A tiny compile, run, connect and share test was fine.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -115,7 +115,7 @@ char *iptostr(struct sockaddr *sa)
               s, sizeof s);
   else
 #else
-  static char s[sizeof "255.255.255.255"] = "";
+  static char s[INET_ADDRSTRLEN] = "";
 #endif
     inet_ntop(AF_INET, &((struct sockaddr_in *)sa)->sin_addr.s_addr, s,
               sizeof s);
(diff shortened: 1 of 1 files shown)


src/net.c
```diff
@@ -1025,14 +1025,15 @@ int sockgets(char *s, int *len)
...
           while (*p == '\n' || *p == '\r')
-            *p++ = 0;
-          if (strlen(socklist[i].handler.sock.inbuf) > 510)
-            socklist[i].handler.sock.inbuf[510] = 0;
-          strcpy(s, socklist[i].handler.sock.inbuf);
+            p++;
+          strlcpy(s, socklist[i].handler.sock.inbuf, 511);
           if (*p) {
             px = nmalloc(strlen(p) + 1);
             strcpy(px, p);
```
Cizzle: While you're at it seems we're missing an `nfree` in the `else` clause 3 below this comment (I can't comment on the specific line 1042). Unless I'm mistaken it'd be nice to have this fixed in this PR.

michaelortmann:
gcc and clang with -fsanitize=address didn't find any leak, and i made sure the suspicious block was executed for my test. last times i didn't trust the compilers i was always wrong, but maybe this is an exceptional case where a human is still smarter than the machine, lets find out...

src/net.c
michaelortmann: gcc and clang with -fsanitize=address didn't find any leak, and i made sure the suspicious block was executed for my test. last times i didn't trust the compilers i was always wrong, but inserting nfree() didn't result in a double free, so this time you have beaten the machine. Hooray Cizzle!

Cizzle:
That last commit looks pointless, what's the gain?

michaelortmann:
We already calculate len() for malloc(), so we can use it to memcpy() instead of strcpy() which is more efficient. Usually we would use strdup(), but thats not eggdrop memory debug friendly and therefore undefed. but maybe one day we could have an eggdrop macro like #define egg_strdup(). im evaluating this elsewhere. anyway, i would call my last commit a small code enhancement. less cpu cycles less global warming. ok, thats kind of exaggerated ;)

Cizzle:
LGTM

vanosg:
Looking at old code (circa 2000), the comment made sense- based on what I see there, it looks like Windows would send only a \r, not a \r\n ? Whatever the reason, it was intentional by the coder at the time. However, the code (and Windows?) progressed from there about 2 years later, without updating the comment. No issues reported in the last 16 years, so I think we're safe... :)

This comment has no bearing on anything, I was just curious about why it would have existed. Moving on.

michaelortmann:
rfc1459, rfc2812 and rfc2813: 2.3 (not 2.3.1): IRC messages are always lines of characters terminated with a CR-LF (Carriage Return - Line Feed) pair[...]. So, an IRC server must send \r\n. Regardless of the underlying operating system. Else the IRC server is broken, not the system. Feel free to call this a bad design decision in the rfc, as i do.

PR #769, opened 2018-11-19
Fix starttls or passreq being on same line as stealth-prompt

Cizzle:
Found by: Geo & Cizzle
Patch by: Cizzle
Fixes: 

One-line summary: When a stealth-prompt in conf (with stealth-telnets set to 1) or 0x53e in a language file (with stealth-telnets set to 0) is used without a trailing newline ("\n"), no upgrade to TLS of a botlink or no link at all when TLS is not compiled in happens.

Additional description (if needed): When open-telnets was set to 1 in conf, linking and upgrading happened as expected.

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -1787,7 +1787,8 @@ static void dcc_telnet_pass(int idx, int atr)
       struct threaddata *td = threaddata();
       /* mark socket to read next incoming at reduced len */
       td->socklist[i].flags |= SOCK_SENTTLS;
-      dprintf(idx, "starttls\n");
+      /* Prefix with \n in case of newline-less ending stealth_prompt */
+      dprintf(idx, "\nstarttls\n");
       putlog(LOG_BOTS, "*", "Sent STARTTLS to %s...", dcc[idx].nick);
     }
 #endif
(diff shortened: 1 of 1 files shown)


src/dcc.c
```diff
@@ -1787,7 +1787,8 @@ static void dcc_telnet_pass(int idx, int atr)
       struct threaddata *td = threaddata();
       /* mark socket to read next incoming at reduced len */
       td->socklist[i].flags |= SOCK_SENTTLS;
-      dprintf(idx, "starttls\n");
+      /* Prefix with \n in case of newline-less ending stealth_prompt */
+      dprintf(idx, "\nstarttls\n");
```
michaelortmann, line 1791: maybe only add \n in front of starttls if stealth_prompt really doesn't end with '\n' ? something like if (*stealth_prompt && stealth_prompt[len(stealth_prompt) - 1] != '\n') ?

src/dcc.c
```diff
@@ -2414,12 +2416,16 @@ static void dcc_telnet_got_ident(int i, char *host)
   /* This is so we don't tell someone doing a portscan anything
    * about ourselves. <cybah>
    */
-  if (stealth_telnets)
+  if (stealth_telnets) {
+    /* Show here so it doesn't interfere with newline-less stealth_prompt */
+    if (allow_new_telnets)
+      dprintf(i, "(If you are new, enter 'NEW' here.)\n");
```
michaelortmann, line 2422: Looks like a visible change to me. Do we really want to display this for case stealth telnets?

michaelortmann (commented):
small questions

src/dcc.c
Cizzle: Unnecessary, "starttls" and "passreq" are only sent to bots, our socket reader skips all "\n" (and "\r").

src/dcc.c
Cizzle: It has been displayed before this PR so the only change is it's location. While it would be handy to be able to configure this (either by defining your own "new_telnet_prompt" or with a toggle setting), it shouldn't be part of this PR.

michaelortmann: Yes and No. It has been displayed before regardless of stealth_prompt. So, your patch here is an enhancement. But: Its showing the new_telnet_prompt in the wrong case. It shows it when stealth_prompt is set instead of if steal_prompt is unset. So if you would switch it to the else case, it would really be a nice enhancement :)

Issue #757, opened 2018-11-07
Bad code style: sizeof info lost with "extern char foo[]"


michaelortmann:
Hei hei,

Eggdrop is using extern char foo[] to use arrays across c files.
So sizeof info is lost, sizeof foo is not useful anymore.
To avoid workarounds and use sizeof in a clean and secure ways across the code, i propose to change all those externs, and put all those char arrays into header files.
This is kind of restructuring / of a bigger change.
But we don't need individual header files per c file, we can put them all into one.
eggdrop.h? main.h? a new one?

PR #781, opened 2018-11-25
Add oident and builtin ident support as module

### LINKED ISSUE #259: Add proper oident support
by writing localip:localport -> remoteip:remoteport { reply "foo" } into the config, not overwriting the whole thing with a global {} directive

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #259

One-line summary:
Add oident support

First this PR was a commit of oidentd support to eggdrop core. Then i wanted to add builtin ident support. IRC discussion result: This PR should be a commit as an eggdrop module.

The oidentd part could deprecate tcls like http://www.tclarchive.org/download.php?id=728 and http://www.tclscripts.net/files/file/33-oidentdtcl/. Regarding the second one, i don't know, if the author is the same, or if it is a gpl violation.

The oidentd part is done.

The builtin ident part could deprecate winident.tcl / part of windrop. It could also be used under (some) unix. See CAP_NET_BIND_SERVICE / http://man7.org/linux/man-pages/man7/capabilities.7.html / setcap / https://packages.debian.org/stretch/libcap2-bin. I don't think many people should / would run eggdrop with escalated caps, but i wanted to document it.

**The builtin ident part is WIP but a first demo was successful, see tests below**

This module solution has advantage over current solutions via tcl in that it minimizes race between oident config file write / setup of identd and connect to irc server. Current tcls can only bind to connect-server and/or init-server. Some do both and / or write a config file twice.

Currently the new event bind before connect (after successful dns resolve etc) is called "ident" but i think you can come up with a better name. At least ident won't confuse scripterz who already know preinit-server.

**Test 1 - oidentd**

This test was carried out with oidentd version 2.3.1.

eggdrop.conf:
```
set username "BotA"
loadmodule ident
```

```
$ ./eggdrop -nt BotA.conf
Module loaded: ident
[03:28:46] Connected to 127.0.0.1
```

```
$ id
uid=1000(michael) gid=1000(michael) groups=1000(michael)
$ ls -la ~/.oidentd.conf 
-rw----r-- 1 michael michael 20 Dec  9 03:28 /home/michael/.oidentd.conf
$ cat ~/.oidentd.conf
global{reply "BotA"}
```

```
$ irc testuser 127.0.0.1
/whois BotA
*** BotA is BotA@localhost (/msg BotA hello)
*** on irc via server zen.localdomain (Example Geographic Location, Planet Earth)
*** BotA has been idle 1 minutes
```

**Test 2 - builtin oidentd**

eggdrop.conf:
```
set username "BotA"
loadmodule ident
set ident-method 1
```

Enable eggdrop to bind to lower ports (linux):
`root$ setcap 'cap_net_bind_service=+ep' /home/michael/eggdrop/eggdrop-1.8.3`

```
$ ./eggdrop -nt BotA.conf
Module loaded: ident
[05:55:35] Connected to 127.0.0.1
.status
    Online as: BotA!BotA@localhost (/msg BotA hello)
    Server [127.0.0.1]:6667 (connected for 00:00)
    Server queue is at 0% (1 msgs)
```

The following was read from and written to wire:
```
read: 47733 , 6667
write: 47733 , 6667 : USERID : UNIX : BotA
```


src/mod/server.mod/servmsg.c
```diff
@@ -1279,6 +1280,28 @@ static void server_resolve_failure(int servidx)
   lostdcc(servidx);
 }
 
+static void do_oident() {
+  char *home = getenv("HOME");
+  char path[121], buf[(sizeof "global{reply \"\"}") + 10];
```
thommey: I don't know if eggdrop limits the setting itself to 10, but I am sure there are ircds that support more than that. can we just put the whole string into it and let oidentd or the ircd limit if necessary?

michaelortmann: Yes, eggdrop limits, see https://github.com/eggheads/eggdrop/blob/develop/src/net.c#L65.

If ircds support more than that, what does it help, if eggdrop limits its username, which we use here for oident.

I would have replaced the magic number 10 with a sizeof or something, that would have answered your question and kept the size dynamic, but thats an eggdrop design problem we should touch at a larger scale, see #757.

thommey (approved):
LGTM

thommey (approved):
LGTM

thommey (approved):
LGTM

src/mod/server.mod/servmsg.c
```diff
@@ -1279,6 +1280,28 @@ static void server_resolve_failure(int servidx)
...
+  char path[121], buf[(sizeof "global{reply \"\"}") + USERLEN];
+  int nbytes;
+  int fd;
+
+  if (!home) {
+    putlog(LOG_SERV, "*",
+           "do_oident(): getenv(): variable HOME is not in the current environment");
+    return;
+  }
+  snprintf(path, sizeof path, "%s/.oidentd.conf", home);
```
janikrabe: I suggest checking the return value of `snprintf` here; otherwise, if the path doesn't fit into the buffer, eggdrop could write to the wrong file.

src/mod/server.mod/server.c
```diff
@@ -1422,6 +1423,7 @@ static tcl_ints my_tcl_ints[] = {
   {"stack-limit",       &stack_limit,               0},
   {"exclusive-binds",   &exclusive_binds,           0},
   {"msg-rate",          &msgrate,                   0},
+  {"oident",            &oident,                    0},
```
janikrabe: Minor nitpick — it's *oidentd*, even when not referring to the daemon itself.

janikrabe (changes requested):

src/mod/server.mod/servmsg.c
michaelortmann: Thank you. done.

src/mod/server.mod/server.c
michaelortmann: Thank you, done.

janikrabe (approved):
LGTM, thanks!

src/mod/ident.mod/ident.c
```diff
@@ -0,0 +1,202 @@
...
+  char buf2[128], *pos;
+  ssize_t i;
+
+  s = answer(dcc[idx].sock, &dcc[idx].sockname, 0, 0);
+  if ((i = read(s, buf2, 64)) < 0) {
+    putlog(LOG_MISC, "*", "Ident error: %s", strerror(errno));
+    return;
+  }
+  buf2[i - 1] = 0;
+  if (!(pos = strpbrk(buf2, "\n\r\\0"))) {
```
janikrabe: Is the double backslash in `\\0` intentional?

src/mod/ident.mod/ident.c
```diff
@@ -0,0 +1,202 @@
...
+    putlog(LOG_MISC, "*", "Ident error: couldnt read request");
+    return;
+  } 
+  snprintf(pos, (sizeof buf2) - (pos - buf2), " : USERID : UNIX : %s\n\r", botname);
```
janikrabe: RFC 1413 requires `\r\n` (CR LF) instead of `\n\r` (LF CR).

src/mod/ident.mod/ident.c
```diff
@@ -0,0 +1,202 @@
...
+  dcc[idx].sock = s;
+  /* dcc[idx].timeval = now; */
+  strcpy(dcc[idx].nick, "(ident)");
+  
+  add_builtins(H_raw, ident_raw);
+}
+
+static void ident_ident()
+{
+  if (ident_method == IDENT_METHOD_OIDENT)
```
janikrabe: A `switch` would work well here.

src/mod/ident.mod/ident.c
michaelortmann: i think strpbrk as a string function cant take \0 as parameter anyway. its was coded too fast and ill need to rework some stuff, thats why i maked the PR as WIP now.

src/mod/ident.mod/ident.c
michaelortmann: Ty! fixed.

src/mod/ident.mod/ident.c
michaelortmann: With only 2 checks, is shouldn't make much difference. A question of style. So you like switch more.

src/mod/ident.mod/ident.c
janikrabe: You're right, `\0` wouldn't have the desired effect either. Just don't forget to change this, since it could cause some hard-to-diagnose issues with port numbers containing zeros.

michaelortmann:
WIP means more doc and testing/feedback needed.

doc:

```
doc/sphinx_source/coreDocs/ident.rst
doc/sphinx_source/coreDocs/modules.rst
doc/settings/mod.ident
doc/settings/CONTENTS
doc/MODULES
maybe help/ ?
src/mod/ident.mod/modinfo
src/mod/ident.mod/help
```

PR #782, opened 2018-11-26
Rewrote dcc relay filter

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Rewrote dcc relay filter

Bogus filter code in dcc_relay was rewritten.
First only the strcpy()s were rewritten as memmove()s, then the whole filter code was rewritten.
Also fixes ancient bug, see http://lists.eggheads.org/pipermail/eggheads/2014-January/012719.html.

I successfully tested:
1. A 'normal' relay from BotA to BotB.
2. Filtering of (multiple) ESC.
3. Filtering of Esc sequences of form _ESC_[...m
  Python test program line was `conn.sendall(b'1 abc abc abc def\x1b[ghimjkl abc abc abc\n')`
  Result (debug printf() only inserted for this test) was `1 abc abc abc defjkl abc abc abc`

### DIFF
--- a/src/botnet.c
+++ b/src/botnet.c
@@ -1468,36 +1468,44 @@ static void eof_dcc_relaying(int idx)
 
 static void dcc_relay(int idx, char *buf, int j)
 {
-  unsigned char *p = (unsigned char *) buf;
-  int mark;
+  unsigned char *src, *dst;
 
   for (j = 0; dcc[j].sock != dcc[idx].u.relay->sock ||
        dcc[j].type != &DCC_RELAYING; j++);
-  /* If redirecting to a non-telnet user, swallow telnet codes and
-   * escape sequences. */
+  /* If redirecting to a non-telnet user, swallow telnet IAC, escape sequences
+   * and CR. */
   if (!(dcc[j].status & STAT_TELNET)) {
-    while (*p != 0) {
-      while (*p != 255 && (*p != '\033' || *(p + 1) != '[') && *p != '\r' && *p)
-        p++;                    /* Search for IAC, escape sequences and CR. */
-      if (*p == 255) {
-        mark = 2;
-        if (!*(p + 1))
-          mark = 1;             /* Bogus */
-        if ((*(p + 1) >= 251) || (*(p + 1) <= 254)) {
-          mark = 3;
-          if (!*(p + 2))
-            mark = 2;           /* Bogus */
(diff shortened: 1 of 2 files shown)


src/botnet.c
```diff
@@ -1468,36 +1468,44 @@ static void eof_dcc_relaying(int idx)
...
+    dst = (unsigned char *) buf;
+    while (*src) {
+      /* Search for IAC, escape sequences and CR. */
+      if (*src == TLN_IAC) {
+        src++;
+        if ((*src >= TLN_WILL) && (*src <= TLN_DONT))
+          src++;
+          if (*src)
+            src++;
+        else if (*src)
```
thommey, line 1489: this is always false because \0 <= TLN_DONT

thommey (changes requested):

src/botnet.c
Cizzle: But \0 !>= TLN_WILL?

michaelortmann: yes, the if is only taken, if BOTH is true
if ((*src >= TLN_WILL) **&&** (*src <= TLN_DONT))
if \0 would be between TLN_WILL and TLN_DONT, which \0 isnt.
and because BOTH must be true, i couldn't easily convert the lines into a switch/case.
so, i think were fine here.
Thanks for reviewing and feeback as always!

src/botnet.c
```diff
@@ -1468,36 +1468,45 @@ static void eof_dcc_relaying(int idx)
...
+          src++;
+      } else if (*src == '\033') { /* ESC */
+        src++;
+        if (*src == '[') { /* CSI */
+          src++;
+          /* Search for the end of the escape sequence. */
+          while (*src && *src != 'm')
+            src++;
+        }
```
Cizzle, line 1497: If I'm not mistaken, the `'m'` is not skipped and will, in the next loop, be added to dst.

Cizzle (changes requested):

src/botnet.c
michaelortmann: Thanks for the finding. I committed a fix and did a read life test.

Cizzle (approved):

Cizzle:
LGTM

thommey (approved):

PR #790, opened 2018-12-04
(WIP) Add egg_strdup() for core and modules and converted 1 example for each

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Add egg_strdup() for core and modules and converted 1 example for each

After this patch, about 100 more code blocks like the few ones done here, can be converted.
Less magick numbers, like in:
```
module_list->name = nmalloc(8);
strcpy(module_list->name, "eggdrop");
```
Maybe also Code like "malloc_strcpy()" in filesys module. 

**Now i wonder, if the nmalloc() memory accounting also / still works, when the nmalloc() is wrapped in a misc.c func.** Maybe the egg_strdup() won't work and this PR is garbage. And i dont want to re-introduce keyword `inline` here. I have the strange feeling, i've dont this before, read about this idea and problem before. **Deja vu.**

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -1064,6 +1064,18 @@ void help_subst(char *s, char *nick, struct flag_record *flags,
   }
 }
 
+/* As strdup() but with nmalloc() */
+char *egg_strdup(const char *str)
+{
+  size_t len;
+  char *copy;
+
+  len = strlen(str) + 1;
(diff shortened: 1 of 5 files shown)


michaelortmann:
egg_strdup() wont be eggdrop memory debug friendly.

PR #793, opened 2018-12-06
Fix getting our ip for dcc

Cizzle:
Found by: Geo
Patch by: Cizzle
Fixes: 

One-line summary: rework getting our ip needed for DCC as it failed in some scenarios.

Additional description (if needed): In particular, it failed when `nat-ip` was not set and `listen-addr` was set to `0.0.0.0` and either IPV6 was disabled (`./configure --disable-ipv6`) or `prefer-ipv6` was set to 1.
The `prefer-ipv6` setting was removed from that part as well as other functions used within make sure we get what we prefer.

Test cases demonstrating functionality (if applicable): Set up two bots, link them and have them share userfiles.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -707,13 +707,13 @@ int getdccfamilyaddr(sockname_t *addr, char *s, size_t l, int restrict_af)
 #ifdef IPV6
      /* If it's listening on an IPv6 :: address,
         try using vhost6 as the source IP */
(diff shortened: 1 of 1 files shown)


vanosg:
I was only able to test on a machine with ipv4 addresses available, but it worked in all cases for that.

thommey (approved):
LGTM

PR #799, opened 2018-12-13
Fix getrandom() syscall not existing while function exists in header.

Cizzle:
Found by: jack3
Patch by: Cizzle
Fixes: 

One-line summary: On systems running Linux older than 3.17 and using glibc 2.25 or newer, th getrandom() function exists in the system header but the system call does not exist.

Test cases demonstrating functionality (if applicable): running eggdrop should not be fatal anymore.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1021,12 +1021,20 @@ int mainloop(int toplevel)
 static void init_random(void) {
   unsigned int seed;
 #ifdef HAVE_GETRANDOM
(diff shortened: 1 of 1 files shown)


thommey:
jack3 confirmed that it is fixed with this patch

vanosg:
LGTM

vanosg:
Sorry, can't resist

> running eggdrop should not be fatal anymore.

I should hope not. Sounds like a really bad movie... Attack of the Killer Eggdrops? RIP jack3 ....

PR #803, opened 2018-12-31
Enhance security by using explicit_bzero()

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Enhance security by using explicit_bzero() for clearing sensitive memory such as a password.

Add compat explicit_bzero() from libsodium.

This patch was startet 20180829, and it would have taken much longer to pick it up again, if there wasn't memsad as a reminder. Eggdrop doesn't suffer from the bug described by memsad, because we never used memset() to clear sensitive memory. Probably because we eggheads all know it would be optimized out. But eggdrop doesn't clear sensitive memory at all.

See also:
1. https://www.usenix.org/sites/default/files/conference/protected-files/usenixsecurity17_slides_zhaomo_yang.pdf
2. https://media.ccc.de/v/35c3-9788-memsad

**Please misc/runautotools and misc/makedepend after merge**


michaelortmann:
Last merge was for conflict resolving. Milestone 1.9? Ready for review again.

PR #826, opened 2019-07-22
dont hang on set sasl 1 if cap is merged but sasl not yet

michaelortmann:
Found by: 
Patch by:
Fixes: 

One-line summary:


michaelortmann:
no need, lets block

PR #839, opened 2019-08-08
Add sasl timeout

michaelortmann:
Found by:
Patch by: michaelortmann
Fixes: 

One-line summary:

To test timeout code disable line https://github.com/eggheads/eggdrop/blob/02a73d075ad0c5223f3e4790884eaa03ec92b681/src/mod/server.mod/servmsg.c#L1299 or simulate loss of authenticate messages from server by making function gotauthenticate() instantly return and not progressing with sasl. The 2nd test would result in something like:
```
[05:26:14] SASL: put AUTHENTICATE ECDSA-NIST256P-CHALLENGE
[05:26:24] SASL: timeout
[05:26:24] -NOTICE- *** Notice -- You have an illegal character in your hostname
[05:26:24] SASL: SASL authentication aborted
[05:26:24] triggering bind evnt:init_server
[05:26:24] triggered bind evnt:init_server, user 0.469ms sys 0.039ms
[05:26:24] -NickServ (NickServ@services.int)- This nickname is registered. Please choose a different nickname, or identify via /msg NickServ identify <password>.
[05:26:24] -NickServ (NickServ@services.int)- You are now identified for BotA.
```
Note that i also tested the eggdrop.conf option sasl-timeout = 10, which would otherwise default to 15.

### DIFF
--- a/src/mod/server.mod/server.c
+++ b/src/mod/server.mod/server.c
@@ -143,6 +143,13 @@ static int burst;
 #include "cmdsserv.c"
 #include "tclserv.c"
 
+/* Available sasl mechanisms. */
+char const *SASL_MECHANISMS[SASL_MECHANISM_NUM] = {
+  [SASL_MECHANISM_PLAIN]                    = "PLAIN",
+  [SASL_MECHANISM_ECDSA_NIST256P_CHALLENGE] = "ECDSA-NIST256P-CHALLENGE",
+  [SASL_MECHANISM_EXTERNAL]                 = "EXTERNAL"
+};
+
 
 static void write_to_server(char *s, unsigned int len) {
   char *s2 = nmalloc(len + 2);
--- a/src/mod/server.mod/server.c
+++ b/src/mod/server.mod/server.c
@@ -129,6 +129,7 @@ static char sasl_password[81];
 static int sasl_continue = 1;
 static char sasl_ecdsa_key[121];
 static int sasl_timeout = 15;
+static int sasl_timeout_time = 0;
 
 #include "servmsg.c"
 
@@ -1686,6 +1687,8 @@ static void server_secondly()
   deq_msg();
   if (!resolvserv && serv < 0)
     connect_server();
+  if (sasl_timeout_time == time(NULL))
+    debug0("TODO: got sasl timeout, handle it here");
 }
 
 static void server_5minutely()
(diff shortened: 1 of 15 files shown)


src/mod/server.mod/servmsg.c
```diff
@@ -1464,15 +1476,10 @@ static int gotcap(char *from, char *msg) {
         putlog(LOG_SERV, "*", "SASL: put AUTHENTICATE %s",
             SASL_MECHANISMS[sasl_mechanism]);
         dprintf(DP_MODE, "AUTHENTICATE %s\n", SASL_MECHANISMS[sasl_mechanism]);
+        sasl_timeout_time = time(NULL) + sasl_timeout;
```
thommey: This is not robust against clock adjustments. It would be better to either do:
```c
sasl_timeout_time = sasl_timeout;
/* in secondly() */
if (!--sasl_timeout) { /* timeout */ }
```
or use clock_gettime with CLOCK_MONOTONIC

src/mod/server.mod/server.c
```diff
@@ -1684,6 +1687,8 @@ static void server_secondly()
   deq_msg();
   if (!resolvserv && serv < 0)
     connect_server();
+  if (sasl_timeout_time == now)
```
thommey: not robust against us missing a second (clock adjustment or blocking execution of anything), better check >= here, but see the other comment for alternative approaches

thommey (changes requested):
I think this would be way better if we used either clock_gettime(CLOCK_MONOTONIC), or counting down 15 times (default) in a secondly() hook than comparing actual timestamps. Eggdrop also has its own timer management with (u)timers exposed to Tcl, but not sure how suitable it is to hook into the same one from C.

thommey:
> I think this would be way better if we used either clock_gettime(CLOCK_MONOTONIC), or counting down 15 times (default) in a secondly() hook than comparing actual timestamps. Eggdrop also has its own timer management with (u)timers exposed to Tcl, but not sure how suitable it is to hook into the same one from C.

just to be clear: the usual eggdrop way is to decrement counters in secondly() hooks, clock_gettime would be a new approach

src/mod/server.mod/server.c
michaelortmann: yes. fixed.

src/mod/server.mod/servmsg.c
michaelortmann: yes. fixed.

michaelortmann:
**Ready for review again.**

thommey (approved):

PR #842, opened 2019-08-17
List bots in alpha order

### LINKED ISSUE #743: Feature add: List bots on partyline in Alpha order when doing .bots or .botinfo
This would be nice to have the ability to list the bots on the party line in order. 

IE
.bots
Bots: 1zark, Cras, Ewalk, Nope

instead of
.bots
Ewalk, 1zark, Nope, Cras

michaelortmann:
Found by: d3vils
Patch by: michaelortmann
Fixes: #743

One-line summary:
make tandem bots list a sorted list

the output of .bots and the like is <this bot> <linked bots in sorted order>

linked 3 bots and checked that .bots output was the same regardless of link order

### DIFF
--- a/src/botnet.c
+++ b/src/botnet.c
@@ -85,20 +85,43 @@ tand_t *findbot(char *who)
  */
 void addbot(char *who, char *from, char *next, char flag, int vernum)
 {
-  tand_t **ptr = &tandbot, *ptr2;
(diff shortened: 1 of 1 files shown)


thommey:
I don't think this is a decent solution. Sorting on "addbot" does not cover loading the userfile or renaming the bots, it would be significantly easier to sort on output. and that doesn't need to be implemented in C. Tcl solution posted in #743

PR #845, opened 2019-08-19
Enhance memory management

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Enhance memory management of botnet party array

botnet parties are stored in an array of party_t. eggdrop started with array size 50 (even for single/unlinked bots, this wastes 4K of memory) and if it gets full adds 50 more. With this patch, eggdrop will start with size 1 and if it gets full doubles size. This is a good algorithm we already use for memtbl. Additionally this patch also makes use of the known lengths of the tand bot list, so we dont need to iterate there.

```
make debug
.console +d
```
Before:
```
.debug
[...]
File 'botnet.c  ' accounted for 5600/5600 (ok)
```
After:
```
.debug
[...]
File 'botnet.c  ' accounted for 0/0 (ok)
[...]
.link BotA
[...]
[00:00:09] Linked to BotA.
[...]
.debug
[...]
File 'botnet.c  ' accounted for 198/198 (ok)
[...]
[00:05:13] botnet: party size doubled to 2.
*** (BotB) michael has joined the party line.
[...]
.debug
[...]
File 'botnet.c  ' accounted for 327/327 (ok)
```

### DIFF
--- a/src/botnet.c
+++ b/src/botnet.c
@@ -37,7 +37,7 @@ extern Tcl_Interp *interp;
 
 tand_t *tandbot;                   /* Keep track of tandem bots on the botnet */
 party_t *party;                    /* Keep track of people on the botnet */
-static int party_size = 1;
+static int party_size = 0;
 int tands = 0;                     /* Number of bots on the botnet */
 int parties = 0;                   /* Number of people on the botnet */
 char botnetnick[HANDLEN + 1] = ""; /* Botnet nickname */
@@ -66,7 +66,6 @@ int expmem_botnet()
 void init_bots()
 {
   tandbot = NULL;
-  party = nmalloc(party_size * sizeof(party_t));
 }
 
 tand_t *findbot(char *who)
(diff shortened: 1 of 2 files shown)


src/botnet.c
```diff
@@ -169,9 +164,14 @@ int addparty(char *bot, char *nick, int chan, char flag, int sock,
...
   /* New member */
-  if (parties == maxparty) {
-    maxparty += 50;
-    party = (party_t *) nrealloc((void *) party, maxparty * sizeof(party_t));
+  if (!party_size) {
+      party_size = 1;
+      party = nmalloc(party_size * sizeof(party_t));
+  }
+  else if (parties == party_size) {
+    party_size <<= 1;
```
thommey, line 172: *=2 compiles to the same instruction and is less confusing for humans to read but ok :)

thommey (approved):

src/botnet.c
michaelortmann: yes and yes, old habit, next time ill change or remove something like this i may remember and change all such occurrences.

PR #859, opened 2019-09-01
Fix oversize

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix oversize

NICKLEN is already NICKMAX + 1, see: https://github.com/eggheads/eggdrop/blob/develop/src/eggdrop.h#L78. So NICKLEN + 1 is oversized.


thommey (approved):
LGTM

PR #866, opened 2019-09-20
Allow space in passwds

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Allow space in passwds

One more char means higher entropy means eggdrop passwds become slightly more secure by this tiny patch.


michaelortmann:
maybe bad idea, when i think of us sharing passwds to bots in cleartext a space may disrupt our botnet parser.

PR #786, opened 2018-11-30
Enhance mtmtbl with dynamic size

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Enhance mtmtbl with dynamic size

Additional description (if needed):
What value should MEMTBLSIZE be? 25000? 250000? If too small eggdrop would fatal error when mtmtbl is full, if too big, eggdrop will waste memory. No! This patch replaces static MEMTBLSIZE with MEMTBLSIZE_START == (1 << 12) == 4096 and MEMTBLSIZE_MAX == (1 << 20) == 1048576. memtbl size will start with MEMTBLSIZE_START and double in size every time it is full, still with an upper limit of MEMTBLSIZE_MAX.

Test cases demonstrating functionality (if applicable):
RSS = resident set size
VSZ = virtual size in Kbytes (alias vsize)

Before:
```
$ ps u
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
michael  27396  0.0  0.1  40844 20188 pts/4    Sl+  22:21   0:00 ./eggdrop -nt
```

```
.status
Memory table: 3639/250000 (1.5% full)
Memory fault: only accounting for 113753/113938 (99.8%)
Memory table itself occupies an additional 9765k static
```
After:
```
$ ps u
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
michael  28007  0.0  0.0  31240 10380 pts/4    Sl+  22:31   0:00 ./eggdrop -nt
```

```
.status
Memory table: 3639/4096 (88.8% full)
Memory fault: only accounting for 113753/113938 (99.8%)
Memory table itself occupies an additional 0k static
```

PR #868, opened 2019-09-21
Fix memory table status report

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix memory table status report

Bug was my fault, introduced with #786, truly sorry.

Before:
```
.status
Memory table: 3256/4096 (79.5% full)
Memory table itself occupies an additional 0k static
```
After:
```
.status
Memory table: 3256/4096 (79.5% full)
Memory table itself occupies an additional 160k static
```

### DIFF
--- a/src/mem.c
+++ b/src/mem.c
@@ -79,7 +79,7 @@ void init_mem()
   size_t size;
   int i;
 
-  size =  memtbl_size * sizeof(*memtbl);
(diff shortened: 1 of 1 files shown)


vanosg (approved):
LGTM

PR #894, opened 2019-10-30
(WIP) Add bind nickonce

michaelortmann:
Found by: Robby
Patch by: michaelortmann
Fixes: 

One-line summary:
Add bind nickonce

Nick binding doesnt trigger per user but per chan. An additional nickonce binding may be useful. Lets discuss it. And whats better to discuss than some actual code ;)

Not tested yet.

### DIFF
--- a/src/mod/irc.mod/chan.c
+++ b/src/mod/irc.mod/chan.c
@@ -2190,13 +2190,13 @@ static int gotnick(char *from, char *msg)
       }
     }
   }
-  u = get_user_by_host(from);
   if (!found) {
+    u = get_user_by_host(from);
     s1[0] = '*';
     s1[1] = 0;
     check_tcl_nick(nick, uhost, u, s1, msg);
   }
-  check_tcl_nickonce(nick, uhost, u, msg);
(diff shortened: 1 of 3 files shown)


thommey:
I see the valid use-case, but I think there are other solutions for it. Such as triggering the bind with "*" as channel name once before triggering the channel binds. "nickonce" is too long for a bind name, they were limited to 4 chars last I checked.

Or maybe a "return 1" to stop triggering more of it (for the other channels)

vanosg:
Correct on 4-char bind limits. This came from a question asked in channel about a NICK bind triggering multiple times if a user is in multiple chans with the bot. If * as the channel bind works (@Robby- have you tried this?) then I think this is not needed, but I would want to confirm * works the right way.

vanosg:
```bind nick -|- * procname  ``` is the bind line in use (haven't tested first hand, but thats what Robby said). So this would imply * as a channel doesn't stop it. return 1 as part of the nick bind might be a pretty decent fix

vanosg:
Addressed by 86de9fc ; in short use the RAWT bind with NICK as a  keyword to only trigger once per nick change, regardless of the number of channels the eggdrop and user share.

Issue #854, opened 2019-08-28
trace-Functions called to many times


michaelortmann:
I see trace-Functions and related functions called to many times, for example during startup of a bot:

https://github.com/eggheads/eggdrop/tree/cbb5dad2566a2b57d6cb8574724433b80827441d/src/mod/server.mod/server.c#L1368 5 times
https://github.com/eggheads/eggdrop/tree/cbb5dad2566a2b57d6cb8574724433b80827441d/src/mod/server.mod/server.c#L1330 6 times
https://github.com/eggheads/eggdrop/tree/cbb5dad2566a2b57d6cb8574724433b80827441d/src/mod/irc.mod/irc.c#L1019 3 times

It's not really a bug, but a waste of CPU.

michaelortmann:
Looks like we set TCL_TRACE_READS too often.

PR #895, opened 2019-11-03
Do not set TCL_TRACE_READS for Tcl_TraceVar(net-type)

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
traced_nettype() is unnecessarily called for TCL_TRACE_READS

This looks like one example of unnecessary calles to (or processings within) trace-Functions. The other cases will be tracked and fixed with #854.

Before (Function is called many times):
```
$ ./eggdrop -nt BotA.conf
[...]
Module loaded: server          
WARNING: Using an integer for net-type is deprecated and will be
         removed in a future release. Please reference an updated
         configuration file for the new allowed string values
Module loaded: ctcp
[...]
use '.help userinfo' for commands.
WARNING: Using an integer for net-type is deprecated and will be
         removed in a future release. Please reference an updated
         configuration file for the new allowed string values
WARNING: Using an integer for net-type is deprecated and will be
         removed in a future release. Please reference an updated
         configuration file for the new allowed string values
Module loaded: ident
```
After (Function is called one time only):
```
Module loaded: server          
WARNING: Using an integer for net-type is deprecated and will be
         removed in a future release. Please reference an updated
         configuration file for the new allowed string values
Module loaded: ctcp
```

### DIFF
--- a/src/mod/server.mod/server.c
+++ b/src/mod/server.mod/server.c
@@ -2071,7 +2071,7 @@ static char *server_close()
                  TCL_TRACE_READS | TCL_TRACE_WRITES | TCL_TRACE_UNSETS,
                  traced_serveraddress, NULL);
   Tcl_UntraceVar(interp, "net-type",
-                 TCL_TRACE_READS | TCL_TRACE_WRITES | TCL_TRACE_UNSETS,
+                 TCL_TRACE_WRITES | TCL_TRACE_UNSETS,
                  traced_nettype, NULL);
   Tcl_UntraceVar(interp, "nick-len",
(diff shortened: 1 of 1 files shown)


thommey:
The concept idea was:
* Write trace to give that warning and update the C variable.
* Read trace to return a string to Tcl whenever the variable is read instead of the internal number.

The second part is just missing it seems like, because if you read ${::net-type} from Tcl now, it will always be the internal number, and never a string, even if you set it to one.

thommey:
On second thought, existing scripts might rely on this being a number from Tcl's point of view. This is fine until we decide to change that.

PR #905, opened 2019-11-26
Update autobotchk

TehPeGaSuS:
Use modern crontab syntax

Shorter and cleaner


michaelortmann:
Portability? Freebsd and Linux seem fine, but what about, lets say, minix? its man page doesnt mention the new syntax: https://man.minix3.org/cgi-bin/man.cgi?query=crontab&apropos=0&sektion=5&manpath=Minix+3.3.0&arch=default&format=html

Cizzle:
Agree with @michaelortmann .

PR #906, opened 2019-11-27
pbkdf2 mod - new password hashing algorithm, update to #356

### LINKED ISSUE #810: [1.8.4] password is limited to 16 characters
This was also valid for 1.6.x and is not mentioned anywhere (not in docs or during chpass). Password is just cut to the first 16 characters.
```
.+user testpass
.chpass testpass 123456789012345678

.tcl passwdok testpass 123456789012345
Tcl: 0
.tcl passwdok testpass 1234567890123456
Tcl: 1
.tcl passwdok testpass 1234567890123457
Tcl: 0
.tcl passwdok testpass 12345678901234567
Tcl: 0
.tcl passwdok testpass 123456789012345678
Tcl: 0
```
This also happens for `/msg op` resulting with "failed OP" for valid password.

The cause of this lies in:
```
static void cmd_chpass(struct userrec *u, int idx, char *par)
[...]
      l = strlen(new = newsplit(&par));
      if (l > 16)
        new[16] = 0; 
```

### LINKED ISSUE #811: [1.8.4] problem with encpass and repeated characters
This should not be the correct behaviour, there should be some variation involved:
```
.tcl encpass x
Tcl: +3Fi9q.glmnd0
.tcl encpass xx
Tcl: +3Fi9q.glmnd0
.tcl encpass xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Tcl: +3Fi9q.glmnd0
```

michaelortmann:
Found by: Various
Patch by: thommey and michaelortmann
Fixes: #810 #811

One-line summary:

TODO:
- Write more Doc (Geo?)

Changelog since #356:
- Merge remote-tracking branch 'upstream/feature/pass2' into pass2 (based on current develop 08-Sep-2019 2caaa82b7c87b72a40e4840767b870eaa47744af)
- Remove mbedtls (we talked about this in IRC)
- Cleanup Makefile and strings like 1.9.0 and 2020
- misc/makedepend
- Update ax_check_openssl.m4
- Do not load mod if necessary openssl stuff not found
- Remove gap in modvals.h
- Honor pbkdf2 module when main.c checks for an encryption hook
- Add `#loadmodule pbkdf2` to eggdrop.conf and eggdrop-basic.conf
- Add config option `pbkdf2-rounds` to eggdrop.conf
- Add config option `pbkdf2-method` to eggdrop.conf. Valid values are output of `openssl list -digest-algorithms`
- Use b64_ntop() and b64_pton() instead of openssl glue
- Macros B64_NTOP_CALCULATE_SIZE(x) and B64_PTON_CALCULATE_SIZE(x) in base64.h
- Enabled password encoding (enabled HOOK_ENCRYPT_PASS2 and added USERENTRY_PASS2)
- cycles -> rounds
- PHC string format (b64 padding = signs are omitted)
- Add benchmark debug for pbkdf2, so one can see how `rounds` effects the time needed to hash a password. My developer system for example shows 5ms for default values (sha512 with 5000 rounds):
```
.chpass testuser 123456  
[10:35:15] pbkdf2 method SHA512 rounds 5000, user 5.284ms sys 0.000ms
```
- Add random generator check and log
- Enabled password verification (userrec.c:u_pass_match())
- Add explicit_bzero()
- Cleaned up all return codes
- Do not load pbkdf2 if configured with --disable-tls
- Add userfile share functionality (There is a small design problem in eggdrop. dup_userlist() calls adduser() and adduser() is eggdrop module api and takes only one password parameter, but we need to set 2 already encrypted passwords. so we cant do it with adduser() and pass_set(). solution is, to write PASS2 after adduser() in  dup_userlist().)
- Clean / Modern OpenSSL library initialization
- Use ERR_error_string() where available
- Add re-encoding of password if pbkdf2-method and / or pbkdf2-rounds change
- Add eggdrop.conf option pbkdf2-re-encode (default = 0) to enable / disable auto re-encoding of password if pbkdf2-method and / or pbkdf2-rounds change
- Add pbkdf2 password if a user logs in that only has a blowfish password and both modules (blowfish and pbkdf2) are loaded
- Verified https://tools.ietf.org/html/rfc6070, that was never a problem, for we use openssl PKCS5_PBKDF2_HMAC().
- Increase PASSWORDMAX to 30 and replace magick 30 with PASSWORDMAX in pass_set() and u_pass_match() and define/use CHARSET_PASSWORD. As a result, automatically generated botlink passwords are now 180 instead of 82 bits of entropy.
- `.tcl getuser <handle> PASS` will now return PASS2, if pbkdf2 loaded and PASS2 set, and return PASS otherwise:
```
.tcl getuser testuser PASS
Tcl: $pbkdf2-SHA512$rounds=5000$hipwbiJOXV3Vw4wpzIsgYQ$J1avFbXvJzlbYUzVFoIoB6tK4tJPskouaBruCvyNUVwQrpxa9Pw9Ex1KPknCYeGM/CKLpLIiEuAeZ9jN/vQLeQ
.tcl getuser testuser PASS2
Tcl: Please use PASS instead.
```
- Added tcl encpass2:
```
.tcl encpass2 test
[22:45:28] tcl: builtin dcc call: *dcc:tcl -HQ 1 encpass2 test
[22:45:28] tcl: evaluate (.tcl): encpass2 test
[22:45:28] pbkdf2 method SHA512 rounds 5000, user 5.380ms sys 0.000ms
Tcl: $pbkdf2-SHA512$rounds=5000$vIvkqmr5MSKeP74gSr2gJA$OyuzcmX6v8K/x5QK1sINI5wN2VIDFHKVGP0YIKPf/NsUVLUl/QqIDgmb0VaP1xu/rgoUUYs7k69JEZD49CgFIA
```

PHC means https://en.wikipedia.org/wiki/Password_Hashing_Competition. The PHC String Format https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md is an attempt to specify a common hash string format that’s a restricted and well defined subset of the Modular Crypt Format. New hashes are strongly encouraged to adhere to the PHC specification, rather than the much looser Modular Crypt Format. The hash strings generated by this module are in the following format:
`$id$rounds=<rounds>$<salt>$<hash>` with `id = pbkdf2-<digest>` for now.
Rounds param name and default value 5000 is taken from https://github.com/freebsd/freebsd/blob/master/lib/libcrypt/crypt-sha256.c. Also equal to https://github.com/lattera/glibc/blob/master/crypt/sha512-crypt.c. Salt len value 16 is taken from freebsd and  NIST SP 800-132. glibc has salt len 8 hardcoded https://github.com/linux-pam/linux-pam/blob/master/modules/pam_unix/passverify.c#L409 and its fine to hardcode it https://bugzilla.redhat.com/show_bug.cgi?id=1229472.

Security-Benchmark Eggdrop-Blowfish vs. PBKDF2-SHA512:
```
$ john --test --format=bfegg
Will run 16 OpenMP threads
Benchmarking: bfegg, Eggdrop [Blowfish 32/64]... (16xOMP) DONE
Raw:	598016 c/s real, 37481 c/s virtual

$ john --test --format=PBKDF2-HMAC-SHA512
Will run 16 OpenMP threads
Benchmarking: PBKDF2-HMAC-SHA512, GRUB2 / OS X 10.8+ [PBKDF2-SHA512 256/256 AVX2 4x]... (16xOMP) DONE
Speed for cost 1 (iteration count) of 1000
Raw:	37440 c/s real, 2348 c/s virtual
```

Because we default to 5000 rounds (iteration count), john would hash 37440 / 5 = 7488 c/s

So: PBKDF2-SHA512 will be 598016 / 7488 = 79 times more secure by this "silly" benchmark. Additionally: The salt makes it harder / impossible to crack a userfile with precalculated tables. Rounds can be tuned to make it more secure, but also slower / blocking. Eggdrop-Blowfish has a short maximum length of passwords, where PBKDF2 could potentially handle unlimited length. (Numbers are for a consumer grade 65W CPU like AMD 3700X, graphic cards and other stuff is faster, but i just wanted to measure some ratio here, to be done with it)

Another security measure is how much entropy is in automatically generated botnet passwords.
Until now, eggdrops used a small charset and 16 chars:
```
>>> math.log2(len("0123456789abcdefghijklmnopqrstuvwxyz") ** 16)
82.718800023077
```
So, that equals 82 bits of entropy.
From now on, eggdrop will use a big charset and 30 chars:
```
>>> math.log2(len("0123456789?ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz") ** 30)
180.0
```
So, that equals 180  bits of entropy.
It doesnt matter much for eggdrops security, but hey.

```
.+user testuser
.chpass testuser 123456
.save
```
eggdrop.user:
```
testuser   - -      
--PASS2 $pbkdf2-SHA512$rounds=5000$K8QXC+2JFJQQvgLINeMebA$Ipg8bgf/Veavxr+0ELBowYnKj0wYxcRf67J3abjWbQpULuMzIGGJwcz0kK8DqnT1x5kK9U20AW/+u24xwiZl1Q
--PASS +9F93o1OxwgK1
--XTRA created 1579508105
```
Test with openssl < 1.0.0 (OpenBSD 4.7 with OpenSSL 0.9.8k)
Also test with without working openssl (i temporarily renamed /usr/include/openssl/ssl.h)
```
$ make config
[...]
checking whether compiling and linking against OpenSSL works... no
configure: warning:

  Cannot find openssl library, its development files or the necessary openssl
  functions EVP_sha256() and PKCS5_PBKDF2_HMAC(). The pbkdf2 module will
  therefore be disabled.

configure: creating ./config.status
```
Test of .chpass was successful for the following cases:
1. only blowfish loaded
2. only pbkdf2 loaded
3. both blowfish and pbkdf2 loaded

Note: PKCS5_PBKDF2_HMAC() needs openssl 1.0.0 (released 29-Mar-2010) or higher. Developer test with an old openssl 1.0.0.c system was successful.

Note2: We will not give an option to remove old blowfish passwords in the process of password migration from blowfish to pbkdf2. We will do so with eggdrop 2.0.0 and/or a script that will alter the userfile.

Note3: USERENTRY_PASS2 becomes part of Module API. But no 3rd partymodule should use it blindly. Its only exported for our share mod. It doesnt do all the checks on user input that USERENTRY_PASS does. Normally the use of USERENTRY_PASS is enough, for it automatically sets PASS2 if needed. In Future, pass_set(), pass2_set() and its usage could be refactored, but that would probably be a bigger change.

Note4: Botlink passwords are kept in PASS only. Keep that in mind, if you plan on blindly removing all "PASS" from userfile.

Note5: Automigration means, if both modules are loaded, and PASS2 is not set (yet) and a user loggs in via PASS, PASS2 is automatically created at this point. Its only working for PASS -> PASS2, not PASS2 -> PASS, because that would have made the code much more complicated with current design.


michaelortmann:
What should happen with --disable-tls?
1. remove --disable-tls with within this PR
2. built eggdrop without tls, but try to built pbkdf2 anyway?
3. dont built/load pbkdf2 in this case, instead fallback to the "intermediately available blowfish (which is probably later going away)" or a stub doing no encryption?

eggdrop.conf
```diff
@@ -615,6 +615,19 @@ die "Please make sure you edit your config file completely."
 set mod-path "modules/"
 
 
+#### PBKDF2 MODULE ####
+#loadmodule pbkdf2
+
+# Encryption method used. openssl list -digest-algorithms
+#set pbkdf2-method "SHA512"
```
bdrewery: About the comment, SHA512 isn't an encryption method.

eggdrop.conf
```diff
@@ -615,6 +615,19 @@ die "Please make sure you edit your config file completely."
...
+
+# Enable re-encoding of password if pbkdf2-method and / or pbkdf2-rounds change.
+#set pbkdf2-re-encode 0
+
+# Number of rounds.
+#set pbkdf2-rounds 5000
```
bdrewery: Assume I'm an idiot. What is rounds and why is 5000 a good default? Just asking for an improved comment for the user.

eggdrop.conf
bdrewery: I know I'm being pedantic but this whole change is for a salted hash which isn't encryption because it cannot be decrypted.

I support the change in principle, one-way hashing is good. There's no reason a password needs to be decrypted by eggdrop.

eggdrop.conf
michaelortmann: the pbkdf2 method adds CPU penalty to an attacker by not only encryptiong a pwd once, but n-times. The n is called rounds or iteration count.

For example glibc uses rounds=5000

https://github.com/lattera/glibc/blob/master/crypt/sha256-crypt.c#L88

The more rounds, the more CPU you put use to crack or to encode a password. You can always go higher, but remember: eggdrop is single threaded. Any passowrd encoding in eggdrop will not only slow the encode down, but block eggdrop as long as it takes. Therefore i also implemented a small debug log showing how many millisecs it took.

michaelortmann: I hope,Geo jumps in and does write some nice documentation for the soon to be finished (?) pbkdf2 module thereby explaining questions like that.

michaelortmann: Oh, and i tried to explain rounds and its effect in the Text in this PR. I copy the text here also:

    Add benchmark debug for pbkdf2, so one can see how rounds effects the time needed to hash a password. My developer system for example shows 5ms for default values (sha512 with 5000 rounds):

.chpass testuser 123456  
[10:35:15] pbkdf2 method SHA512 rounds 5000, user 5.284ms sys 0.000ms

eggdrop.conf
michaelortmann: Yeah, ill change that, thanks!

src/mod/pbkdf2.mod/pbkdf2.c
```diff
@@ -0,0 +1,302 @@
...
+    return -1;
+  c = strchr(target, '=');
+  if (c) {
+    *c = 0;
+    targsize = (c - target);
+  }
+  return targsize;
+}
+
+/* Return
```
thommey, line 60: The name of the parameter "rounds" is "c" in the PBKDF2 specification, and https://www.npmjs.com/package/@phc/format uses "i". Where did "rounds" come from as parameter name?
I like that you found the PHC format, great find and great idea to use it

src/mod/pbkdf2.mod/pbkdf2.c
```diff
@@ -0,0 +1,302 @@
...
+ *   NULL = error
+ */
+static char *pbkdf2_hash(const char *pass, const char *digest_name,
+                         const unsigned char *salt, unsigned int saltlen,
+                         unsigned int rounds)
+{
+  const EVP_MD *digest;
+  int digestlen, ret;
+  int outlen, restlen;
+  static char *out; /* static object is initialized to zero (Standard C) */
```
thommey: Can we change this to have the caller pass a dstbuf instead? Re-entrancy probably won't matter but it's an unusual calling convention "result is valid until function is called again". If not, that should at least be boldly documented, because it's new to eggdrop.
If worried about dst buffer size, we can require the dst buffer size being passed as well and being at least the max length of the PHC entry, just document it with a comment

src/mod/pbkdf2.mod/pbkdf2.c
```diff
@@ -0,0 +1,302 @@
...
+       b64salt[B64_NTOP_CALCULATE_SIZE(PBKDF2_SALT_LEN) + 1],
+       b64hash[B64_NTOP_CALCULATE_SIZE(256) + 1];
+  char format[40];
+  unsigned int rounds;
+  const EVP_MD *digest;
+  unsigned char salt[PBKDF2_SALT_LEN + 1];
+  int saltlen;
+  static char *buf;
+
+  if (snprintf(format, sizeof format,
```
thommey, line 180: Just as a note: this might not be worth the confusion, splitting on $ is a lot clearer, and this can't parse arbitrary PHC strings, just the exact ones we wrote (no other parameters etc.). There's room for improvement here later then I guess. But fine for now, of course.

src/userent.c
```diff
@@ -237,38 +238,117 @@ struct user_entry_type USERENTRY_INFO = {
...
+  else /* remove PASS2 */
+    e->u.extra = new;
+  return 0;
+}
+
+static int def_tcl_null(Tcl_Interp * irp, struct userrec *u,
+                        struct user_entry *e, int argc, char **argv)
+{
+  Tcl_AppendResult(irp, "Please use PASS instead.", NULL);
+  return TCL_OK;
```
thommey: return TCL_ERROR; should be better, as it raises an error on the Tcl level, TCL_OK does not and might never show this error

src/userent.c
```diff
@@ -237,38 +238,117 @@ struct user_entry_type USERENTRY_INFO = {
...
+  if (e->u.extra) {
+    explicit_bzero(e->u.extra, strlen(e->u.extra));
+    nfree(e->u.extra);
+  }
+  if (new) { /* set PASS2 */
+    e->u.extra = user_malloc(strlen(new) + 1);
+    strcpy(e->u.extra, new);
+  }
+  else /* remove PASS2 */
+    e->u.extra = new;
```
thommey: Readability suggestion: e->u.extra = NULL

thommey (commented):
Looks great, thanks!

src/mod/pbkdf2.mod/pbkdf2.c
michaelortmann: Parapeter rounds comes from glibc https://github.com/lattera/glibc/blob/master/crypt/sha512-crypt.c#L83 see also man 3 crypt.

src/mod/pbkdf2.mod/pbkdf2.c
michaelortmann: After some debate with thommey over irc, its now fixed with commit 1986367b458bc6c48c23d1c986fdfdc2e6681a4f.

michaelortmann: oh, and  came up with a 5th solution: do it like snprintf, if you give that function a NULL instead of a buffer, it could return the size, like snprintf() does. so you can call it twice.

src/userent.c
michaelortmann: Fixed with commit fd3d0a5d715439954827555c611efdaeb3ca08cc

src/userent.c
michaelortmann: Fixed with commit a0823f67db9ba0e30f43a317553c83b11c322d4a.

src/mod/pbkdf2.mod/pbkdf2.c
michaelortmann: So, how will you track this "wishlist item"? Are you gonna open an Issue for it?

vanosg:
Githubism: Fixes #811

michaelortmann:
Option 1: Keep PASS API, implement pbkdf2 under PASS2 API (implementation 1). 
---------
- pbkdf2 does PASS2 entries
- allows compatible transition from PASS API
- requires PASS module to be loaded to use encpass, encrypt, decrypt functions
- deprecates encpass, encrypt, decrypt functions
- adds encpass2, encrypt2, decrypt2 functions
- No "breakage" as blowfish is still available
- Remove blowfish module from repo in Eggdrop 2.0

Option 2: Keep PASS API, implement pbkdf2 under PASS2 API (implementation 2). 
---------
- pbkdf2 does PASS2 entries
- allows compatible transition from PASS API
- requires PASS module to be loaded to use encpass, encrypt, decrypt functions
- deprecates encpass, encrypt, decrypt functions
- adds encpasss2 function
- ignores encrypt2, decrypt2 functions
- Remove blowfish module from repo in Eggdrop 2.0
- No "breakage" as blowfish is still available

Option 2.1: Keep PASS API, implement pbkdf2 under PASS2 API (implementation 2). 
---------
- pbkdf2 does PASS2 entries
- allows compatible transition from PASS API
- requires PASS module to be loaded to use encpass, encrypt, decrypt functions
- deprecates encpass, encrypt, decrypt functions
- adds encpasss2 function
- ignores encrypt2, decrypt2 functions
- Remove blowfish module from repo in Eggdrop 2.0
- No "breakage" as blowfish is still available
- Add "mortmanns king of the day" config option to remove PASS entries, when/while corresponding PASS2 entries are created/found (this is the only difference to Option 2 and not the same as a cleanup-tcl, which would not hinder the creation of new PASS entries)

Option 3: Implement PBKDF2 under PASS API
---------
- pbkdf2 does PASS entries
- no compatible transition from PASS API
- replaces encpass, encrypt, decrypt function with pbkdf2 hashes instead of blowfish
- No "breakage" as blowfish is still available, just used in lieu of pbkdf2
- mortman is pissed b/c all his hard transition work is gone

Option 4: Crippled blowfish
---------
- Package PBKDF2 under PASS2 API (no encrypt/decrypt functions?)
- Create blowfish.tcl to offer (override?) encrypt/decrypt functions

PR #913, opened 2019-12-10
Fix type

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
base64 functions can return -1 which is checked for


thommey:
If olen2 is set by ECDSA_sign to be greater than INT_MAX, it will be negative for the b64 call, is this what we want?

michaelortmann:
@thommey: thx. both functions use unsigned/size_t (i cant bolden the ECDSA_sing() type, because of the stars (*) and github markup parser ;)

        int ECDSA_sign(int type, const unsigned char *dgst, int dgstlen,
                       unsigned char *sig, unsigned int *siglen, EC_KEY *eckey);

int b64_ntop(u_char const *, **size_t**, char *, size_t);

fixed. ready for review again.

michaelortmann:
last commit was for conflict resolving with develop only. still ready for review again.

thommey (approved):
LGTM

PR #922, opened 2020-01-02
Rip out starttls.

### LINKED ISSUE #535: SSL

Cizzle:
Found by:
Patch by: Cizzle
Fixes: #535 

One-line summary: Now only TLS-to-TLS and plain-to-plain botlinks are allowed; although plain links can still be upgraded manually.

Also see #552.

### DIFF
--- a/doc/sphinx_source/mainDocs/tls.rst
+++ b/doc/sphinx_source/mainDocs/tls.rst
@@ -1,5 +1,5 @@
 TLS support
-Last revised: Oct 17, 2010
+Last revised: Jan 2, 2020
 
 ===========
 TLS support
@@ -36,9 +36,9 @@ custom one, as they take precedence over any system-wide paths.
 Usage
 -----
 
-By default, without additional configuration, TLS support will provide
-opportunistic encryption for botnet links. For other connection types,
-TLS must be requested explicitly.
+By default and before 1.9.0, without additional configuration, TLS support
+will provide opportunistic encryption for botnet links. From 1.9.0 onward
+or for other connection types, TLS must be requested explicitly.
 
 Secure connections are created the same way as plaintext ones. The only
 difference is that you must prefix the port number with a plus sign.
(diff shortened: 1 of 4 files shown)


vanosg:
For documentation purposes, 1.8 link tree:

1.8 | 1.9 | Result
----|-----| --------
Links to ()port | Listens with ()port | &#x1F49A; 1.8 upgrades to TLS with STARTTLS
Links to ()port | Listens with +port | &#x1F49A; Tries plain/STARTTLS, fails, then tries TLS, succeeds
Links to +port | Listens with ()port | &#x1F534; Tries TLS to plain port, Fails
Links to +port | Listens with +port | &#x1F49A; Succeeds
 | |
Listens with ()port | Links to ()port | &#x1F49A; 1.8 sends STARTTLS, upgrades to TLS
Listens with ()port | Links to +port | &#x1F534; Tries TLS to plain port, fails
Listens with +port | Links to ()port | &#x1F534; Timeout (Tries plain to TLS port, fails)
Listens with +port | Links to +port | &#x1F49A; Succeeds

In short, a 1.8 bot will still require TLS for a link to a 1.9 bot

vanosg:
LGTM :+1:

vanosg (approved):

PR #929, opened 2020-02-19
Minor changes to botchk

astrorigin:
Found by: Stanislas Marquis <stan@astrorigin.com>
Patch by: Stanislas Marquis <stan@astrorigin.com>
Fixes: -

One-line summary:
Mainly add a reminder that config file must be executable with a proper shebang.

If for some reason the config file is not executable, botchk will fail. A simple reminder should be enough to prevent that problem.
Also took the opportunity to simplify instructions given for crontab.

### DIFF
--- a/scripts/botchk
+++ b/scripts/botchk
@@ -10,10 +10,10 @@
 #
 # To check for your bot every 10 minutes, put the following line in your
 # crontab:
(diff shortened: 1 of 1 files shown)


vanosg:
Thanks for taking the time to put this together! However, similar to a previous issue submitted, this formatting is not portable. I hope we'll see more input from you in the future!

PR #938, opened 2020-03-09
Restrict b/e/I to <63 char hostmask

### LINKED ISSUE #887: Does eggdrop truncate valid bans longer than 70 chars?
**This looks like a question, but i think its an Issue.**

In a quick test on freenode i was was able to set the following valid ban:

`1234567890123456!1234567890@1234567890.123456790.123456790.123456790.123456790.123456790.12`

which is 91 chars long:

```
$ python -c 'print(len("1234567890123456!1234567890@1234567890.123456790.123456790.123456790.123456790.123456790.12"))'
91
```

Does eggdrop truncate valid bans longer than 70 chars?

I think so because of https://github.com/eggheads/eggdrop/blob/develop/src/mod/channels.mod/cmdschan.c#L132

vanosg:
Found by: @michaelortmann 
Patch by: Geo
Fixes: #887 

One-line summary:
Restrict b/e/I to <63 char hostmask

As pointed out in the thread started by @michaelortmann that led to spirited discussion in channel, the ban/exempt/invite maximum of 70 did not seem to reflect the rfc. 2812 states 
```<hostname> has a maximum length of 63 characters.  This is a limitation of the protocol as internet hostnames (in particular) can be longer.  Such restriction is necessary because IRC messages are limited to 512 characters in length. Clients connecting from a host which name is longer than 63 characters are  registered using the host (numeric) address instead of the host name.``` 
As previously written, eggdrop was checking a ban mask length in its entirety, to include nick and ident, and then truncating to a 70-char length for the mask as a whole. This patch updates the check to only look at the hostname portion of a mask, not the mask in its entirety, and then truncate appropriately. 

```
.+ban 123456789!987654321@023456789.123456789.223456789.323456789.423456789.523456789.623456789
New ban: 123456789!987654321@023456789.123456789.223456789.323456789.423456789.523456789.62* (requested)

```
```
.+invite 123456789!987654321@023456789.123456789.223456789.323456789.423456789.523456789.623456789
New invite: 123456789!987654321@023456789.123456789.223456789.323456789.423456789.523456789.62* (requested)
```
```
.+exempt 123456789!987654321@023456789.123456789.223456789.323456789.423456789.523456789.623456789
New exempt: 123456789!987654321@023456789.123456789.223456789.323456789.423456789.523456789.62* (requested)
```

### DIFF
--- a/src/eggdrop.h
+++ b/src/eggdrop.h
@@ -56,6 +56,7 @@
 
 
 /* Handy string lengths */
+#define HOSTMAX     63            /* DNS RFC 1035, IRC RFC 2812          */
 #define UHOSTMAX    291 + NICKMAX /* 32 (ident) + 3 (\0, !, @) + NICKMAX */
 #define DIRMAX      512           /* paranoia                            */
 #define LOGLINEMAX  9000          /* for misc.c/putlog() <cybah>         */
--- a/src/mod/channels.mod/cmdschan.c
+++ b/src/mod/channels.mod/cmdschan.c
@@ -29,7 +29,7 @@ static struct flag_record victim = { FR_GLOBAL | FR_CHAN, 0, 0, 0, 0, 0 };
 
 static void cmd_pls_ban(struct userrec *u, int idx, char *par)
 {
-  char *chname, *who, s[UHOSTLEN], s1[UHOSTLEN], *p, *p_expire;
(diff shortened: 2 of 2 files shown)


vanosg:
When comparing the diffs, you can skip the garbage in lines 146-~170 cmd_pls_ban- its just reformatting the incorrect existing space formatting and the GH diff is freaking out over it.

PR #943, opened 2020-03-22
ident mod - remove oidentd.conf after use

### LINKED ISSUE #942: Remove $HOME/.oidentd.conf after successful connection to server
Right now, on develop branch, eggdrop has support for oidentd.
The problem is that the bot creates a `$HOME/oidentd.conf` which will be read by the oidentd daemon to reply to identd queries from the IRC server.
So far, so good. The problem is that if any other kind of bot that doesn't have support for this will end up having the same ident has the eggdrop.
The solution would be to remove the `$HOME/oidentd.conf` after the bot successfully connected to the IRC server.

michaelortmann:
Found by: PeGaSuS-Coder
Patch by: michaelortmann
Fixes: #942

One-line summary:
Remove oidentd.conf after use

So far so good, but note, that this opens a tiny race for when another user / process writes .oidentd.conf between eggies creation / writing and removing. No problem for eggdrop, but the other user / process could have its file removed. We could do a sanity check before removing the file. That would minimize the race time window, but its not possible to close it. Let me know, if you want that sanity check which means, eggdrop would check, if the file it will remove looks like the one it created / wrote.

```
[01:11:29] Trying server 127.0.0.1:6667
[01:11:29] Ident: Creating / writing /home/michael/.oidentd.conf.
[01:11:29] net: open_telnet_raw(): idx 3 host 127.0.0.1 port 6667 ssl 0
[01:11:29] net: connect! sock 8
[01:11:29] Connected to 127.0.0.1
[01:11:30] Ident: Removing /home/michael/.oidentd.conf.
[01:11:30] triggering bind quotepass_unbind
[01:11:30] triggered bind quotepass_unbind, user 0.027ms sys 0.000ms
[01:11:30] triggering bind evnt:init_server
[01:11:30] triggered bind evnt:init_server, user 0.076ms sys 0.000ms
```

### DIFF
--- a/src/mod/ident.mod/ident.c
+++ b/src/mod/ident.mod/ident.c
@@ -52,8 +52,8 @@ static cmd_t ident_raw_builtin_off[] = {
 };
 
 static cmd_t ident_raw_oident_off[] = {
-  {"001", "",   (IntFunc) ident_oident_off,  "ident:001"},
-  {NULL,  NULL, NULL,                        NULL       }
+  {"001", "",   (IntFunc) ident_oident_off, "ident:001"},
+  {NULL,  NULL, NULL,                       NULL       }
 };
 
 char path[121]; /* oidentd.conf */
--- a/src/mod/ident.mod/ident.c
+++ b/src/mod/ident.mod/ident.c
@@ -44,11 +44,19 @@ static tcl_ints identints[] = {
 };
 
 static void ident_builtin_off();
+static void ident_oident_off();
 
-static cmd_t ident_raw[] = {
+static cmd_t ident_raw_builtin_off[] = {
   {"001", "",   (IntFunc) ident_builtin_off, "ident:001"},
-  {NULL,  NULL, NULL,                        NULL}
+  {NULL,  NULL, NULL,                        NULL       }
 };
+
+static cmd_t ident_raw_oident_off[] = {
+  {"001", "",   (IntFunc) ident_oident_off,  "ident:001"},
+  {NULL,  NULL, NULL,                        NULL       }
+};
+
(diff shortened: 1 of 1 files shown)


thommey:
I would rather restrict the oidentd.conf entries from "global" to specify the source-port and destination ip-port combination. that should be unique and won't conflict with another bot.

deleting this file makes debugging more difficult, and it is a known limitation of oidentd.conf applications that they have race conditions

michaelortmann:
We could add an eggdrop.conf option under ident module section that enabled/disabled removing of oident file. Yes or No ?

michaelortmann:
+1 for the restrict idea.

vanosg:
I suggest switching to a directive-based format instead of removing the file. This would allow the user to have other programs using oidentd as well that we don't blow away. Using a format of 
```
to my.irc.org fport 6667 from vhost4 lport <the local port> { reply myIdent } ### pidfilename (see comment in writeup below)
```
would prevent other bots from being pushed to use the provided ident. When the bot connects, check for the servername and port combo the bot is connecting to and, if found, remove it and replace it with the updated directive (most likely just chaging lport). This could/would end up with some clutter in the .conf file, but I think I would be OK with that. Maybe we could also add a check at Eggdrop startup that scans the config file for its pidfilename and removes everything it finds, before ever connecting to a server? It wouldn't be 100%, but could help manage some of the clutter.

Alternatively, we could just write the same directive followed by just a "### Eggdrop" comment and replace that single line every time a bot connects, but if 4 bots were to start at once, for example after a reboot, there would be competition for owning that line and all the Eggdrop's would likely end up with the same ident.

vanosg:
OBE by #1042

PR #909, opened 2019-11-28
Enhance password checking, logging and rejecting

michaelortmann:
Found by: wilkowy
Patch by: michaelortmann
Fixes: #810 

One-line summary:
Enhance password len checking, logging, rejecting (instead of truncating)

Additional description (if needed):
1. The upper limit for password length within eggdrop was never consistent, like /msg cutted to 15 chars, chpass to 16 chars.
2. Unify with commit 103128c93f31bb09bb0a8149cc1dac51f5625858
3. Later, #906 will probably lift the upper limit to password length, which will be easier, once this PR unified password length handling.
4. Fixed return codes in msgcmds.c:msg_pass() (was not really broken, but hey ;)
5. Eggdrop distinguishes plain from encrypted passwords (in pass_set()) with first char == '+'. At the same time it allows users to enter such + passwords. Those would not only end up as plain text in userfile, but would also forbid the user to use it / login. This PR fixes this. It rejects such + passwds, see test case below.
6. Less code doubling by using a new common function. Sadly, it cant be used for /msg also, because of the different logging/noticing.

Test cases demonstrating functionality (if applicable):
**Test 1 - + password - Before**
```
Before:

.chpass testuser +12345
Changed password.

Userfile:
testuser   - hp
--PASS +12345
--XTRA created 1585178624

$ telnet 127.0.0.1 3333
Trying 127.0.0.1...
Connected to 127.0.0.1.
Escape character is '^]'.


Nickname.
testuser

Enter your password.
Negative on that, Houston.
Connection closed by foreign host.
```
**Test 1 - Partyline - After**
```
.chpass testuser 12345
[02:59:06] tcl: builtin dcc call: *dcc:chpass -HQ 1 [something]
Please use at least 6 characters.
.chpass testuser 123456
[02:59:11] tcl: builtin dcc call: *dcc:chpass -HQ 1 [something]
[02:59:11] #-HQ# chpass testuser [something]
Changed password.
.chpass testuser 123456789012345
[02:59:16] tcl: builtin dcc call: *dcc:chpass -HQ 1 [something]
[02:59:16] #-HQ# chpass testuser [something]
Changed password.
.chpass testuser 1234567890123456
[02:59:17] tcl: builtin dcc call: *dcc:chpass -HQ 1 [something]
Please use at most 15 characters.
.chpass testuser +12345
[02:59:22] tcl: builtin dcc call: *dcc:chpass -HQ 1 [something]
Please do not use + as first character.
```
**Test 2 - /msg - After**
```
[...]

PR #947, opened 2020-03-25
Password: Please do not use + as first character.

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix for passwords starting with +

Eggdrop distinguishes plain from encrypted passwords (in pass_set()) with first char == '+'. At the same time it allows users to enter such + passwords. Those would not only end up as plain text in userfile, but would also forbid the user to use it / login.

With this PR, + passwords will be rejected.

**Before:**
```
.chpass testuser +12345
Changed password.

Userfile:
testuser   - hp
--PASS +12345
--XTRA created 1585178624

$ telnet 127.0.0.1 3333
Trying 127.0.0.1...
Connected to 127.0.0.1.
Escape character is '^]'.


Nickname.
testuser

Enter your password.
Negative on that, Houston.
Connection closed by foreign host.
```
**After:**
```
.chpass testuser +12345
[00:28:17] tcl: builtin dcc call: *dcc:chpass -HQ 1 [something]
Please do not use + as first character.
```

### DIFF
--- a/src/cmds.c
+++ b/src/cmds.c
@@ -407,6 +407,10 @@ static void cmd_newpass(struct userrec *u, int idx, char *par)
     dprintf(idx, "Please use at least 6 characters.\n");
     return;
   }
+  if (new[0] == '+') {
+    dprintf(idx, "Please do not use + as first character.\n");
+    return;
+  }
   set_user(&USERENTRY_PASS, u, new);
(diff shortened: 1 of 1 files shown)


michaelortmann:
merged into #909

PR #950, opened 2020-03-29
Clean / Modern OpenSSL library initialization

michaelortmann:
Found by: eggheads/eggdrop
Patch by: eggheads/eggdrop
Fixes: 

One-line summary:
Clean / Modern OpenSSL library initialization

See also: https://wiki.openssl.org/index.php/Library_Initialization

A small developer test linking two eggdrop bots via ssl was successful for OpenSSL 1.1.1e and 1.0.2p.

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -125,7 +125,7 @@ int ssl_init()
 #if OPENSSL_VERSION_NUMBER < 0x10100000L /* 1.1.0 */
   SSL_library_init();
   SSL_load_error_strings();
-  OpenSSL_add_all_algorithms
+  OpenSSL_add_all_algorithms();
 #endif
   if (ssl_seed()) {
     putlog(LOG_MISC, "*", "ERROR: TLS: unable to seed PRNG. Disabling SSL");
--- a/src/tls.c
+++ b/src/tls.c
@@ -120,10 +120,13 @@ static int ssl_seed(void)
  */
 int ssl_init()
 {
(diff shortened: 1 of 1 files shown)


src/tls.c
```diff
@@ -120,10 +120,13 @@ static int ssl_seed(void)
...
-  /* Load SSL and crypto error strings; register SSL algorithms */
-  SSL_load_error_strings();
+  /* OpenSSL library initialization
+   * If you are using 1.1.0 or above then you don't need to take any further steps. */
+#if OPENSSL_VERSION_NUMBER < 0x10100000L /* 1.1.0 */
   SSL_library_init();
-
+  SSL_load_error_strings();
+  OpenSSL_add_all_algorithms();
+#endif
```
thommey, line 129: Shouldn't there be `#else OPENSSL_init_ssl()`? Where's the stuff that's called for 1.1.0 or higher?

michaelortmann: Thats the thing. For openbsd 1.1.0+ there is no initialization needed anymore :)
See also: https://wiki.openssl.org/index.php/Library_Initialization

michaelortmann: The manpage to OPENSSL_init_ssl also sais:

As of version 1.1.0 OpenSSL will automatically allocate all resources that it needs so no explicit
       initialisation is required. Similarly it will also automatically deinitialise as required

PR #956, opened 2020-03-30
Add Tcl isaway command

vanosg:
Found by: thommey
Patch by: Geo
Fixes: 

One-line summary:
Add a Tcl 'isaway' command.

Track away status via 352 and ircv3 AWAY messages
Moved AWAY tracking from server.mod to irc.mod

```
### .channel output:
.channel #eggheads
(n = owner, m = master, o = op, d = deop, b = bot)
 NICKNAME    HANDLE     JOIN  IDLE         USER@HOST
 nick1       *          ---                <- it's me!
 nick2       *          ---    44m         foo
 nick3       *          ---    14m         foo
 nick4       *          ---    44m (away)  foo
 nick5       handle     ---  N 37m         foo
 nick6       *          ---    44m         foo
 nick7       *          ---    44m         foo

### bot is on channel, user is not away:
.tcl isaway nick
Tcl: 0
### user then uses away command (away-notify not enabled):
.tcl isaway nick
Tcl: 0
### bot refreshes channel member list:
.tcl resetchan #foober w
Tcl: 
.tcl isaway nick
Tcl: 1

```
Now, with away-notify enabled:
```
### away-notify enabled:
.tcl cap active
Tcl: away-notify
### user is not away:
.tcl isaway nick
Tcl: 0
### user goes away
.tcl isaway nick
Tcl: 1
```


vanosg:
One possible limitation to this PR- .channels will be accurate immediately after the bot joins a channel and does a WHO, but the 'away' list will grow stale unless manually refreshed. Perhaps adding a specific 'chanreset' call that doesn't update the memberlist (currently the 'w' flag), but only update away status, like an 'a' flag? Using the w flag as it exists now would reset the idle counter in .channels.

doc/sphinx_source/mainDocs/tcl-commands.rst
```diff
@@ -1068,6 +1068,14 @@ isvoice <nickname> [channel]
 
   Module: irc
 
+^^^^^^^^^^^^^^^^^
+isaway <nickname>
```
thommey: this should be `isaway <nickname> [channel]` like all the other commands

src/mod/irc.mod/tclirc.c
```diff
@@ -434,6 +434,30 @@ static int tcl_ischaninvite STDVAR
...
+{
+  struct chanset_t *chan;
+  memberlist *m;
+
+  BADARGS(2, 2, " nick");
+
+  for (chan = chanset; chan; chan = chan->next) {
+    m = ismember(chan, argv[1]);     /* In my channel list copy? */
+    if (m) {
+        if chan_ircaway(m) {
```
thommey: where are the paranthesis around the if-condition? the indentation is weird, too

thommey (changes requested):

doc/sphinx_source/mainDocs/tcl-commands.rst
vanosg: isaway is a global state, not on a per-channel basis. I can't be away on one channel, but not another. It is server-level.

doc/sphinx_source/mainDocs/tcl-commands.rst
```diff
@@ -1068,6 +1068,14 @@ isvoice <nickname> [channel]
...
+^^^^^^^^^^^^^^^^^
+
+  Returns: 1 if Eggdrop is currently tracking someone by that nickname is marked 'away' by an IRC server; 0 otherwise. IMPORTANT: this command will function properly when used in conjunction with the IRCv3 away-notify capability, if away-notify is enabled on the IRC server (if you didn't add this to your config file, it isn't enabled). Without away-notify functionality, this command may not be accurate. For best results without away-notify enabled, use 'resetchan <channel> w' on a channel the user is on to refresh the channel list. This will obtain the current away status for the user.
```
thommey: add comment about how to check if the cap is enabled via Tcl here :-)

doc/sphinx_source/mainDocs/tcl-commands.rst
thommey: we do not track users server-wide. that data structure does not exist in eggdrop. we track channelusers.

src/mod/channels.mod/channels.h
```diff
@@ -38,7 +38,8 @@
 #define CHAN_RESETBANS    0x08
 #define CHAN_RESETEXEMPTS 0x10
 #define CHAN_RESETINVITED 0x20
-#define CHAN_RESETALL     0x3F
+#define CHAN_RESETAWAY    0x40
```
thommey, line 41: please call it RESETWHO or similar, this can be used for many more things (like accounts)

vanosg: RESETWHO is already in use 5 lines above this. It updates the entire member record from a 352 message, which involves clearing out the entire channel list and resetting it- to include IDLE times maintained by eggdrop. The RESETAWAY flag does not clear the record like WHO does, only update the away field of the user record (thereby preserving the idle time). The resetchan command is structured to be used for many more things via the use of flags, you can see those already currently available in the tcl-command doc. Something like account-notify could/should be easily fit into that structure

vanosg:
```
### Checks when user is not away
.tcl isaway geo
Tcl: 0
.tcl isaway geo #eggdroptest
Tcl: 0

### User goes away, resetchan because no away-notify...
.tcl resetchan #eggdroptest a
Tcl: 
.tcl isaway geo
Tcl: 1
.tcl isaway geo #eggdroptest
Tcl: 1

### And it doesn't work when it's not supposed to work.
.tcl isaway
Tcl error: wrong # args: should be "isaway nick ?channel?"
```

vanosg:
```
.channel
Channel #eggdroptest, 2 members, mode +tn:
(n = owner, m = master, o = op, d = deop, b = bot)
 NICKNAME  HANDLE     JOIN  IDLE         USER@HOST
 User       User      ---  N         3m       foo@user.com
 BeerBot   *          ---            3m        <- it's me!
End of channel info.
```
user goes away, then:

```
.tcl refreshchan #eggdroptest w
Tcl: 
.channel
Channel #eggdroptest, 2 members, mode +tn:
(n = owner, m = master, o = op, d = deop, b = bot)
 NICKNAME  HANDLE     JOIN  IDLE         USER@HOST
 User       User      ---  N  5m (away)  foo@user.com
 BeerBot   *          ---                <- it's me!
End of channel info.
```
away status is updated, idle time not reset.
user returns from away:
```
.tcl resetchan #eggdroptest w
Tcl: 
.channel
Channel #eggdroptest, 2 members, mode +tn:
(n = owner, m = master, o = op, d = deop, b = bot)
 NICKNAME  HANDLE     JOIN  IDLE         USER@HOST
 User       User       ---  N             user@foo.com
 BeerBot   *          ---                <- it's me!
End of channel info.
```
User away status is updated, idle times are reset.

PR #957, opened 2020-03-30
Module API Cleanup

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Yet another tiny cleanup: #define strncpyz strlcpy directly in header file


src/modules.c
```diff
@@ -598,7 +596,7 @@ Function global_table[] = {
   (Function) 0,
 #endif
   /* 304 - 307 */
-  (Function) strncpyz,
+  (Function) 0,                   /* was strncpyz() -- use strlcpy() instead */
```
thommey: Is this necessary? Keeping it in there would allow loading modules compiled for an older version of eggdrop. Removing it makes them crash when they use strncpyz. The rest of the changes is fine.

michaelortmann: We did this also with functions like my_memcpy() and egg_strcasecmp().

michaelortmann: maybe it was a mistake not to do this also for my_memcpy() and friends.
so lets do it proper from now on.
done, ready for review again.

PR #960, opened 2020-03-31
Enhance mem / status report

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Enhance mem / status report

Before:
`Memory table: 3605/4096 (88.0% full)`
After:
`Memory table: 3616/4096 (0.3% of max 1048576 full)`

`max == mem.c:MEMTBLSIZE_MAX`

Before:
```
$ ./eggdrop -nc BotA.conf
I am BotA, running eggdrop v1.9.0+msgtags: 17 users (mem: 266k).
Online for 00:00 (status mode) - CPU: 00:00.02 - Cache hit:  0.0%
Configured with: 'CC=clang -fsanitize=address'
Admin: Lamer <email: lamer@lamest.lame.org>
Config file: BotA.conf
OS: Linux 5.5.11-arch1-1
Process ID: 699526 (parent 558778)
Tcl library: /usr/lib/tcl8.6
Tcl version: 8.6.10 (header version 8.6.10)
Tcl is threaded.
TLS support is enabled.
TLS library: OpenSSL 1.1.1e  17 Mar 2020
IPv6 support is enabled.
Socket table: 20/100
    No server currently.
    Active CAP negotiations: None
Memory table: 3605/4096 (88.0% full)
Memory table itself occupies an additional 160k static
```
After:
```
$ ./eggdrop -nc BotA.conf
I am BotA, running eggdrop v1.9.0+msgtags: 17 users (mem: 266k).
Online for 00:00 (status mode) - CPU: 00:00.01 - Cache hit:  0.0%
Configured with: 
Admin: Lamer <email: lamer@lamest.lame.org>
Config file: BotA.conf
OS: Linux 5.5.11-arch1-1
Process./eggdrop -nc BotA.conf ID: 708312 (parent 707370)
Tcl library: /usr/lib/tcl8.6
Tcl version: 8.6.10 (header version 8.6.10)
Tcl is threaded.
TLS support is enabled.
TLS library: OpenSSL 1.1.1e  17 Mar 2020
IPv6 support is enabled.
Socket table: 20/100
    No server currently.
    Active CAP negotiations: None
Memory table: 3616/4096 (0.3% of max 1048576 full)
Memory table itself occupies an additional 160k static

$ ./eggdrop -nc BotA.conf
.status
I am BotA, running eggdrop v1.9.0+msgtags: 18 users (mem: 266k).
Online for 00:00 (terminal mode) - CPU: 00:00.01 - Cache hit: 66.7%
Configured with: 
Admin: Lamer <email: lamer@lamest.lame.org>
Config file: BotA.conf
OS: Linux 5.5.11-arch1-1
Process ID: 708464 (parent 707370)
Tcl library: /usr/lib/tcl8.6
Tcl version: 8.6.10 (header version 8.6.10)
Tcl is threaded.
TLS support is enabled.
TLS library: OpenSSL 1.1.1e  17 Mar 2020
IPv6 support is enabled.
Socket table: 20/100
Memory table: 3625/4096 (0.3% of max 1048576 full)
Memory table itself occupies an additional 160k static
Loaded module information:
    No server currently.
    Active CAP negotiations: None
```

### DIFF
--- a/src/mem.c
+++ b/src/mem.c
@@ -23,7 +23,7 @@
  */
 
 #define MEMTBLSIZE_START (1 << 12) /* 4096    */
-#define MEMTBLSIZE_MAX   (1 << 20) /* 1048576 */
+#define MEMTBLSIZE_MAX   1048576   /* (1 << 20), must be STRINGIFY()able */
 #define COMPILING_MEM
 
 #include "main.h"
@@ -97,9 +97,10 @@ void tell_mem_status(char *nick)
 #ifdef DEBUG_MEM
   float per;
 
-  per = ((lastused * 1.0) / (memtbl_size * 1.0)) * 100.0;
-  dprintf(DP_HELP, "NOTICE %s :Memory table usage: %d/%d (%.1f%% full)\n",
-          nick, lastused, memtbl_size, per);
+  per = 100.0 * (lastused * 1.0) / (MEMTBLSIZE_MAX * 1.0);
+  dprintf(DP_HELP, "NOTICE %s :Memory table usage: %d/%d (%.1f%% of max "
+          STRINGIFY(MEMTBLSIZE_MAX) " full)\n", nick, lastused, memtbl_size,
+          per);
 #endif
   dprintf(DP_HELP, "NOTICE %s :Think I'm using about %dk.\n", nick,
           (int) (expected_memory() / 1024));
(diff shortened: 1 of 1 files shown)


src/mem.c
```diff
@@ -97,9 +97,10 @@ void tell_mem_status(char *nick)
 #ifdef DEBUG_MEM
   float per;
 
-  per = ((lastused * 1.0) / (memtbl_size * 1.0)) * 100.0;
-  dprintf(DP_HELP, "NOTICE %s :Memory table usage: %d/%d (%.1f%% full)\n",
-          nick, lastused, memtbl_size, per);
+  per = 100.0 * (lastused * 1.0) / (MEMTBLSIZE_MAX * 1.0);
```
thommey, line 100: the reason 100.0 was at the end was overflow-prevention (100 * lastused can overflow while lastused/MEMTBLSIZE_MAX * 100 might not)
`per = 1.0 * lastused / MEMTBLSIZE_MAX * 100` would work, too, or cast one to (double), or revert to what it was :-)

src/mem.c
```diff
@@ -97,9 +97,10 @@ void tell_mem_status(char *nick)
...
+  dprintf(DP_HELP, "NOTICE %s :Memory table usage: %d/%d (%.1f%% of max "
```
thommey, line 101: I don't see the benefit of STRINGIFY when we can just use %zu and print MEMTBLSIZE_MAX, then it doesn't need to be stringify'able

src/mem.c
michaelortmann: isn't the result kinda wrong in both cases if something maxes out? i was optimizing for rounding errors, thats why i always multiply before division. but i see your point. the error margin is bigger for overflow. so i will change the order again...

src/mem.c
michaelortmann: it already is stringifyable. we dont have an extra variable here.

PR #948, opened 2020-03-27
(WIP) Fix early startup logging

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #627

One-line summary:
Add -d (Show debug info) command line argument and fix putlog() to respect conmask.

Additional description (if needed):
Eggdrop reads console setting with log levels from config file.
But it is using putlog() with LOG_DEBUG before reading it.
There was no way to dynamically tell eggdrop, when LOG_DEBUG is wanted and when not for this early startup period.
So it was a static setting in eggdrop effectively disabling LOG_DEBUG until config file is read.
Because the early startup period was buggy regarding logging, another bug found its way into then code: putlog() did not respect conmask.
This PR adds -d (Show debug info) command line argument and fixes putlog() to respect conmask.

Test cases demonstrating functionality (if applicable):
**Test 1 - BotA.conf console setting without d**
Before:
```
$ ./eggdrop -nt BotA.conf 

Eggdrop v1.9.0+msgtags (C) 1997 Robey Pointer (C) 2010-2020 Eggheads
--- Loading eggdrop v1.9.0+msgtags (Fri Mar 27 2020)
Listening for telnet connections on 0.0.0.0:3333 (all).
Module loaded: blowfish        
Module loaded: dns             
Module loaded: channels        
Module loaded: server          
[...]
```
After:
```
$ ./eggdrop -nt BotA.conf 

Eggdrop v1.9.0+msgtags (C) 1997 Robey Pointer (C) 2010-2020 Eggheads
--- Loading eggdrop v1.9.0+msgtags (Fri Mar 27 2020)
Listening for telnet connections on 0.0.0.0:3333 (all).
Module loaded: blowfish        
Module loaded: dns             
Module loaded: channels        
Module loaded: server
[...]

$ ./eggdrop -ntd BotA.conf 
LANG: Language loaded: english
LANG: Section loaded: core
LANG: 391 messages of 432 lines loaded from ./language/core.english.lang
LANG: 391 adds, 0 updates to message table

Eggdrop v1.9.0+msgtags (C) 1997 Robey Pointer (C) 2010-2020 Eggheads
Allocated bind table unld (flags 1)
Allocated bind table time (flags 1)
Allocated bind table cron (flags 1)
Allocated bind table note (flags 0)
Allocated bind table nkch (flags 1)
Allocated bind table load (flags 1)
Allocated bind table link (flags 1)
Allocated bind table filt (flags 1)
Allocated bind table disc (flags 1)
[...]

PR #967, opened 2020-04-05
Update inet_aton.c from FreeBSD uptream removes advertising clause

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
The new https://github.com/freebsd/freebsd/blob/455de2f90f71a495eb5068b060c17de9bfcffda6/lib/libc/inet/inet_addr.c **removes advertising clause** from license.

The only user of inet_aton() within eggdrop is one function in dns module. I tested this function and enhanced its logging. But you cant see the `DNS: Valid dns-server` message, because #948 is not merged yet. egg_inet_aton() now uses isdigit(), islower(), isspace() instead of egg_*() due to all of them being posix 2001.
Removal of the advertising clause helps with #226 see https://github.com/eggheads/eggdrop/issues/226#issuecomment-295974628_

```
$ grep INET_ATON config.h
#define HAVE_INET_ATON 1
$ nm eggdrop|grep aton
                 U inet_aton@@GLIBC_2.2.5

$ grep INET_ATON config.h
/* #define HAVE_INET_ATON 1 */
$ nm eggdrop|grep aton
00000000000589c0 T egg_inet_aton
```


src/compat/inet_aton.c
```diff
@@ -99,19 +102,14 @@ static char rcsid[] = "$-Id: inet_addr.c,v 1.11 1999/04/29 18:19:53 drepper Exp
...
  */
-int egg_inet_aton(cp, addr)
-const char *cp;
-struct in_addr *addr;
-{
-  static const uint32_t max[4] = { 0xffffffff, 0xffffff, 0xffff, 0xff };
-  uint32_t val;       /* changed from u_long --david */
-  int base;
-  int n;
+int
```
vanosg: multi-line

vanosg (commented):
Please test the cases that are affected by these changes and show code equivalency

vanosg:
This is more than just removing the advertisement, there are some semi-significant updates to the code being run. This would require some extensive testing to validate- please provide test cases for the new code in use. Thanks!

src/compat/inet_aton.c
michaelortmann: done

michaelortmann:
vanosg: As we spoke, the multi line fix is committed and thus this PR is ready for review again.

PR #999, opened 2020-07-01
Fix u_char->uint8_t, because it is not c99 nor POSIX, fix for diet libc

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
This patch helps portability, esp. for diet libc.

### DIFF
--- a/src/mod/server.mod/servmsg.c
+++ b/src/mod/server.mod/servmsg.c
@@ -20,6 +20,8 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
  */
 
+#undef answer /* before resolv.h because it could collide with src/mod/module.h
+		 (dietlibc) */
 #include <resolv.h> /* base64 encode b64_ntop() and base64 decode b64_pton() */
 #ifdef TLS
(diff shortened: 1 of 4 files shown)


src/mod/server.mod/servmsg.c
```diff
@@ -20,6 +20,8 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
  */
 
+#undef answer /* before resolv.h because it could collide with src/mod/module.h
```
vanosg, line 23: guard with an "#ifdef answer" or similar

vanosg (changes requested):

src/mod/server.mod/servmsg.c
michaelortmann: is there proof we need a guard around undef? i did a small test that shows we can undef arbitrary strings that are not defined:
```
$ cat test.c
#undef tommyknockers23
int main() {
  return 0;
}
$ gcc test.c -o test
```
if you still want me to guard it, let me know and ill do it.

PR #1004, opened 2020-07-12
(Low Prio) Fix dead nested assignment

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann, Cizzle
Fixes: 

One-line summary:
Fix dead nested assignment

Although the value stored to 'chname' is used in the enclosing expression, the value is never actually read from 'chname'

No functional change


Cizzle:
LGTM

PR #1007, opened 2020-07-16
Update userinfo.tcl

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Added new fields YOUTUBE and TWITCH

First i removed ICQ, because ICQ is dead since 2018/2019 when its OSCAR protocol was disabled. But after vanosg stated it is not dead yet, i readded ICQ, to let that zombie live on.

### DIFF
--- a/scripts/userinfo.tcl
+++ b/scripts/userinfo.tcl
@@ -1,4 +1,4 @@
-# userinfo.tcl v1.06 for Eggdrop 1.4.3 and higher
+# userinfo.tcl v1.08 for Eggdrop 1.4.3 and higher
 #           Scott G. Taylor -- ButchBub!staylor@mrynet.com
 #
 # v1.00      ButchBub     14 July      1997 -Original release.  Based on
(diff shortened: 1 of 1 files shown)


scripts/userinfo.tcl
```diff
@@ -61,18 +63,19 @@
...
+#   YOUTUBE YouTube channel
+#   TWITCH  Twitch channel
 
 
 ################################
 # Set your desired fields here #
 ################################
 
-set userinfo-fields "URL BF GF IRL EMAIL DOB PHONE ICQ"
+set userinfo-fields "URL BF GF IRL EMAIL DOB PHONE YOUTUBE TWITCH"
```
vanosg: please leave ICQ in (it has not in fact gone away, and users may have used that field for something else)

michaelortmann: ok ok :)

PR #1014, opened 2020-08-11
Fix declaration of host

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
in case of error, an undefined host would be returned, which looks problematic, better be safe than sorry here.

a hostname is max 256 chars, not UHOSTLEN. UHOSTLEN is greater than 256 currently, but also, better be safe than sorry.


vanosg (approved):
LGTM

PR #1016, opened 2020-08-13
Enhance status with reporting maximum number of open files (sockets)

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:

While working on dns async code with threads and pipes it came to me, that enhancing .status with reporting maximum number of open files (sockets) could be useful. but i wanted the dns PR to be slim, so i copid the work for .status over to this new PR.

```
.status
[...]
Socket table: 20/100
Memory table: 4421/8192 (54.0% full)
[...]
```

```
.status all
[...]
Socket table: 20/100
Maximum number of open files (sockets): soft limit 1024 hard limit 524288
Memory table: 4421/8192 (54.0% full)
[...]
```


src/chanprog.c
```diff
@@ -366,6 +367,12 @@ void tell_verbose_status(int idx)
                "Threaded DNS core is disabled.\n"
 #endif
                "Socket table: %d/%d\n", threaddata()->MAXSOCKS, max_socks);
+
+  if (!getrlimit(RLIMIT_NOFILE, &rlp)) /* TODO: should be posix 2001
+                                        * make sure this is as portable as expected
+                                        * and maybe also add reporting of
+                                        * current amount of open files */
+    dprintf(idx, "Maximum number of open files (sockets): cur %ju max %ju\n", rlp.rlim_cur, rlp.rlim_max);
```
thommey: This does need casts to (uint_max) because variadic functions don't auto-cast and the size they fetch depends on the specifier

michaelortmann: ty, done.

thommey (approved):
LGTM

vanosg:
Can we move this to .status all ? .status is becoming bloated.

michaelortmann:
> Can we move this to .status all ? .status is becoming bloated.

done. this makes this PR even more useful, because now we could easily hide more details, that are currently shown for `.status` and could be shown only for `.status all`. but this is future talk. the function `tell_status() `with its new parameter `detail 0|1` would make it easy now.

vanosg:
I like where your head is at and appreciate you roughing in code for future work, but I think we should either implement such a change completely, or don't. Looking over how this fits in larger code, I think I'd suggest we just leave it like it is but I'm open to other thoughts or reasons to move. I think anything we move from .status to .status all can just slide right into the existing tell_verbose_status(); the existing format (I think) was created the way it was because some of these functions are called from other sections of code. @thommey thoughts?

michaelortmann:
@Geo: done, the status boilerplate code is now unaltered. open files/sockets are only shown for `.status **all**` with minimal amount of code added.

PR #1015, opened 2020-08-12
Make core dns async

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: lots of dns issues

**One-line summary:**
Make core dns async by using threads + pipes

**Additional description (if needed):**
Please run **misc/runautotools** when merging this PR.

We discussed on IRC:
- that we don't want fork() or new external libs, if threads would work
- that we don't want / can't use signals with threads
- that i would come up with a proof of concept for threads + pipes, one thread per request, one pipe per thread, pipes are used to trigger the mainloop select(). meanwhile this PR has outgrown the status of "proof of concept" and is ready for review.

Status:
- Ready for review. pls test and gimme feedback.
- both functions, hostbyip and ipbyhost, in core are async by using threads + pipes (one thread per hostip / ipbyhost)
- successful developer test below
- i was wondering if we need to do anything special on .rehash, so i tested it successfully and looked at the code, so far so good
- i was also wondering, if the new async code plays nice with signals, so i tested kill -HUP and kill to eggdrop, so far so good. (as a reminder for myself: i dont expect problems, but if a problem with signal occurs, we should remember, that threads can be created with sigmask)

The idea is to make eggdrops core dns use libc resolver functions in a non-blocking way. This would make the dns mod  superflous, but there is one thing the dns mod has and the core has not, which is quite useful: defining nameservers in eggdrop.conf. in any other case, the new core dns (once ready) should be made default, because the dns mod has severall issues, is hard to maintain and may one day become a portability problem, if it aint already.

This solution uses pthreads, which is available under posix 2001 and cygwin, so we are fine using it now. Back when dns mod was created that was not such a good option, but now it is.

This PR also shows, that the thread / pipe solution is not that much code after all.

The dns_threads handling is in linked list instead of array. there is a good reason to prefer the linked list: https://stackoverflow.com/questions/166884/array-versus-linked-list/167016#167016. In other words: any dns worker thread has one element in that data structure, be it array or linked list. and a thread may and will modify its own element at any time. so when we remove elements in the main thread, and we would have an array, we would need to shuffle elements around. This would break our simple multithreading implementation. With linked list we are just fine.

Maybe we are also able to use eggdrops own dns servers, if we simply manipulate _res.nsaddr_list and friends (if available - to be fully portable), before we call libc resolve functions. This would make the dns mod finaly obsolete.

[...]

PR #1019, opened 2020-08-18
(RFC) CWE-350 add option to check for dns spoofing

### LINKED ISSUE #13: Reverse DNS Lookup Spoofing
| Trac | Data |
| --: | :-- |
| Ticket | 13 |
| Reported by | BarkerJr |
| Status | assigned |
| Component | Core |
| Priority | major |
| Milestone | 1.8.0 |
| Version | 1.8.0 CVS |

It turns out that Eggdrop does not check to see if the DNS is spoofed when
accepting a telnet session.

Telnet connection: barkerjr.ircd/56705
Denied telnet: barkerjr@barkerjr.ircd, No Access

Now, this would be fine if the host was actually valid for forward resolving. 
However...
- Dns resolved 69.50.185.193 to barkerjr.ircd
- Dns unable to resolve barkerjr.ircd

Since it's not verifying the forward DNS, this could be a problem.  Reverse DNS
is specified by the net-block administrator, so it can easily be invalid, by
mistake or intentionally.

So, this could be a problem, as any net-block admin can spoof hostnames and
telnet to your bot as a known user's hostname, rendering protect-telnet ineffective.

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #13

One-line summary:
Add option dns-spoof-protect (default 0) to check for dns spoofing.

If dns-spoof-protect is set to 1 then for any successful reverse lookup from ip to hostname, a lookup from that hostname to its ip-list is done and it is checked that the ip is in the ip-list. Failed lookup / spoof is logged and eggdrop will use the ip instead of the reverse lookedup hostname.
Based on #1015
This will fix the dns resolver in eggdrops core that is using libc functions.
The dns mod could be having other dns spoofing problems more similar to what a dns server could have.
This (any such) check can only detect some kinds of dns spoofing by design.

Tested with dns server unbound adblocking s3-1-w.amazonaws.com (mapping to 127.0.0.1):
```
.tcl dnslookup 52.216.85.251 foo
[05:38:22] tcl: builtin dcc call: *dcc:tcl -HQ 1 dnslookup 52.216.85.251 foo
[05:38:22] tcl: evaluate (.tcl): dnslookup 52.216.85.251 foo
Tcl: 
[05:38:22] net: dns spoof protect: host s3-1-w.amazonaws.com could not be resolved to ip 52.216.85.251
[05:38:22] Tcl error [foo]: invalid command name "foo"
invalid command name "foo"
    while executing
"foo 52.216.85.251 52.216.85.251 0"
```

### DIFF
--- a/eggdrop.conf
+++ b/eggdrop.conf
@@ -93,6 +93,8 @@ set offset "5"
 # is not supported, the other one will be tried.
 set prefer-ipv6 0
 
+set dns-spoof-protect 0
+
 # If you want to have your Eggdrop messages displayed in a language other
 # than English, change this setting to match your preference. An alternative
 # would be to set the environment variable EGG_LANG to that value.
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -286,7 +286,6 @@ void tell_verbose_status(int idx)
   time_t now2 = now - online_since, hr, min;
   double cputime, cache_total;
   struct utsname un;
-  struct rlimit rlp;
 
   if (uname(&un) < 0) {
     vers_t = " ";
@@ -381,12 +380,6 @@ void tell_verbose_status(int idx)
   dprintf(idx, "IPv6 support is not available.\n"
 #endif
                "Socket table: %d/%d\n", threaddata()->MAXSOCKS, max_socks);
-
-  if (!getrlimit(RLIMIT_NOFILE, &rlp)) /* TODO: should be posix 2001
-                                        * make sure this is as portable as expected
-                                        * and maybe also add reporting of
-                                        * current amount of open files */
-    dprintf(idx, "Maximum number of open files (sockets): cur %ju max %ju\n", rlp.rlim_cur, rlp.rlim_max);
 }
 
 /* Show all internal state variables
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -286,6 +286,7 @@ void tell_verbose_status(int idx)
   time_t now2 = now - online_since, hr, min;
   double cputime, cache_total;
   struct utsname un;
+  struct rlimit rlp;
 
   if (uname(&un) < 0) {
     vers_t = " ";
@@ -380,6 +381,12 @@ void tell_verbose_status(int idx)
   dprintf(idx, "IPv6 support is not available.\n"
 #endif
                "Socket table: %d/%d\n", threaddata()->MAXSOCKS, max_socks);
+
+  if (!getrlimit(RLIMIT_NOFILE, &rlp)) /* TODO: should be posix 2001
+                                        * make sure this is as portable as expected
+                                        * and maybe also add reporting of
+                                        * current amount of open files */
+    dprintf(idx, "Maximum number of open files (sockets): cur %ju max %ju\n", rlp.rlim_cur, rlp.rlim_max);
 }
 
 /* Show all internal state variables
--- a/src/dns.c
+++ b/src/dns.c
@@ -494,13 +494,12 @@ void *thread_dns_ipbyhost(void *arg)
   int i;
   sockname_t *addr = &dtn->addr;
 #ifdef IPV6
-  int found;
+  int found = 0;
 #endif
 
   i = getaddrinfo(dtn->host, NULL, NULL, &res0);
   if (!i) {
     memset(addr, 0, sizeof *addr);
-    found = 0;
     for (res = res0; res; res = res->ai_next) {
       if (res->ai_family == AF_INET) {
         addr->family = res->ai_family;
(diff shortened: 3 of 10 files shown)


h1z1:
o_O  This isn't dns spoofing..   You're effectively breaking an established part (dare I say feature), of the internet.   Hosts are not required to have matching forward/reverse dns.

michaelortmann:
No that is not required. So the optional spoof check doesn't reject any ip missing reverse resolution or any host missing forward resolution. It rejects only the case that a hostname, resolved from an ip, doesn't resolve back to that ip. There are the following cases of reverse resolving, and thats how they are handled with this PR:
```
ip without reverse lookup -> bot will use ip
ip with reversed lookup:
  if the lookedup hostname can be resolved back to that ip-> bot will use that hostname
  if the lookedup hostname cant be resolved back to that ip -> bot will use ip and write a log about it
```

Its the only way to detect what CWE-350 describes with (2):
[...]
The software performs reverse DNS resolution on an IP address to obtain the hostname and make a security decision, but it does not properly ensure that the IP address is truly associated with the hostname.
[...]
Attackers can spoof DNS names by either (1) compromising a DNS server and modifying its records (sometimes called DNS cache poisoning), or (2) having legitimate control over a DNS server associated with their IP address.
[...]

And its the reason, why i named it dns spoofing:
- Cache poisoning is DNS spoofing, but thats not what this PR is for.
- Hostname spoofing is DNS spoofing, thats what this PR is for.

IRC servers are using the same method to detect spoofed hostnames:
https://superuser.com/questions/160718/how-does-hostname-spoofing-customization-work-on-irc-networks/161040#161040
So i think we can also.

h1z1:
> No that is not requred. But its the only way the bot can detect what CWE-350 describes:
> The software performs reverse DNS resolution on an IP address to obtain the hostname and make a security decision, but it does not properly ensure that the IP address is truly associated with the hostname.

It doesn't need to nor can it.  Hostname based auth of any kind was never  really meant to be secure, some protocols like ssh warns very loudly about it too.  SSH of course has a cert it can verify but even smtp has preflight data it can use (the helo for example).

> If the term dns spoofing doesnt cover this case, then what is the correct term?

That would depend how you're configuring the connect string  and what your expected risk is.  The risk in the attack is not so much a spoof  but poison.   I can mount an offline attack against your resolver (dnssec is a whole other issue).

> The **optional** check also doesnt reject ips missing forward/reverse dns, it only rejects reverse lookedup hostnames that dont resolve back to the ip, there are the following cases now:
> 
> ```
> ip without reverse lookup -> bot will use ip
> ip with reversed lookup:
>   if the lookedup hostname can be resolved back to that ip-> bot will use that hostname
>   if the lookedup hostname cant be resolved back to that ip -> bot will use ip and write a log about it
> ```

I realize it's somewhat pedantic, it is worth pointing out though that even ssh's assumption about this is annoying because ips do not need to resolve to their host.

From the linked bug:

> So, this could be a problem, as any net-block admin can spoof hostnames and
telnet to your bot as a known user's hostname, rendering protect-telnet ineffective.

That still remains especially if they are an admin or some third party upstream.

tl;dr is don't use hostnames alone for anything security related :)

michaelortmann:
Sry, i modified my comment between your answer. I think we talk about 2 different things. Cache poisoning cs. hostname spoofing by modifying the name server of my ip.

Yes, i fully agree about the tl;dr. So let me add, that we don't rely on hostname alone, we rely on it additionally to username/pass/cert. And we are not talking only about authentification, we are talking also about spoofing hostnames in logfiles, which can also be dangerous. This PR is one mitigation of one specific problem and its, yeah, optional.

michaelortmann:
If you modify your name server **and** poinson mine, yes, then im lost, even with this PR.

michaelortmann:
The alternatives would be: Reject #13 with words like: the dns system is broken by design, wont try to fix any part of it in eggdrop. In this case we could discuss adding some warning like better not using hostnames, but IPs for telnet access to the bot or warning about a matched hostname during access. We could also change this PR is not rejecting such hostnames, but only warn about the fact that it cant be resolved back. Or we could make the new config option have 3 choices: OFF, WARN(LOG) and REJECT(LOG).

Thank you h1z1 for your input.

h1z1:
>     Hostname spoofing is DNS spoofing, thats what this PR is for.
>   IRC servers are using the same method to detect spoofed hostnames:

ircds do attempt to validate but doesn't really trust the result for anything.  You're not given +O for example without further attestation.   I'm sure there are still networks using  broken y-lines..


> If you modify your name server **and** poinson mine, yes, then im lost, even with this PR.

I don't need to modify mine.  One surprisingly still effective attack is simply flooding yours to brute force the id (16 bit).   Need not even go that far if I''m upstream from you.

> In this case we could discuss adding some warning like better not using hostnames, 

Or "it's 2020 you really, really shouldn't be using telnet" :)

michaelortmann:
CANT FIX
reasoning:
http://jdebp.info/FGA/dns-avoid-double-reverse.html

PR #1028, opened 2020-09-12
(WIP) Remove compat snprintf() and vsnprintf() (eggdrop requires posix 2001)

michaelortmann:
Found by: vanosg
Patch by: michaelortmann
Fixes: #226

One-line summary:
Eggdrop depends on POSIX.1-2001. POSIX.1-2001 defines snprintf() and vsnprintf(). This patch gets rid of src/compat/snprintf*. egg_snprintf() and egg_vsnprintf() are left as is, but redefined to directly use snprintf() and vsnprintf() with a one-liner define in eggdrop.h. we only keep egg_snprintf() and egg_vsnprintf() for backward compatibility to (older) modules, and could remove it from later eggdrop versions when api breakage becomes possible.

All egg_snprintf() and egg_vsnprintf() were replaced by snprintf() and vsnprintf():
```
find ./ -type f -exec sed -i 's/egg_snprintf(/snprintf(/g' {} \;
find ./ -type f -exec sed -i 's/egg_vsnprintf(/vsnprintf(/g' {} \;
```
**Note: don't do this in git directory. cd src/ first!**

**With this PR you must run misc/runautotools and misc/makedepend.**

To help with reviewing, i separated the "mass replace with find" into commit d8586f4cd5da585c9cf17eaff446070fc8308dae, everything else  i put into commit 9c447512cc7ec1a6d3b06a484deff9467f2ac655.

This PR is still WIP, because switching to directly using snprintf() helps the compiler analyzing the code and reporting back more warnings. Now i want to check those warning, maybe fix those within this PR. Those compiler warnings are not about new code, or something off with this PR, the code was always there and the same, but compiler warnings were shadowed by our egg_snprintf() magick and now we get to finally see them in all their glory.

Le me know, if you want
(1) me to fix all new compiler warnings within this PR before merging this
(2) me to fix all new compiler warnings in separate PRs, that will look like #1029, before merging this
(3) to merge this PR now (and fix the compiler warnings later)
(4) me to be king of the day again ;)


michaelortmann:
This PR was big, so it should be done in 2 parts instead. part 1 is PR #1523 cleaning up `(egg_)vsnprintf()`, part 2 is to be done cleaning up (egg_)snprintf(). Then we should start fixing the prototypes, see #1940.

Issue #650, opened 2018-09-26
Fix mem waste / Cleanup buffer size and functions for "nick!userhost" glueing


michaelortmann:
There are many places within eggdrop, where "nick!userhost" is glued togather like this:

egg_snprintf(s, sizeof s, "%s!%s", m->nick, m->userhost);

Often the used function differs. Sometimes sprintf, sometimes, egg_snprintf, maybe others as well.

Often  the used buffer sitze differ. Sometimes static 512, sometimes dynamic, maybe others as well.

It should always be exactly NICKLEN + UHOSTLEN.

And we should always use the same function, be it egg_snprintf or sprintf.

**By fixing the buffer size, we fix truncation or regain precious memory, depends if the buffer was too small or too big**

For 2 such cases it will be done by #649.

Cizzle:
As mentioned in another issue/pr; from eggdrop.h:
```
#define HANDLEN    32 /* valid values 9->NICKMAX                             */
#define NICKMAX    32 /* valid values HANDLEN->32                            */
#define UHOSTMAX   291 + NICKMAX /* 32 (ident) + 3 (\0, !, @) + NICKMAX */
#define NICKLEN    NICKMAX + 1
#define UHOSTLEN   UHOSTMAX + 1
```
Meaning NICKLEN + UHOSTLEN doesn't give what you think you'll get.

michaelortmann:
Yes, thats why we need to fix the definitions in eggdrop.h, starting with anylyzing what UHOSTLEN should be instead of what it is now. where does the 291 even come from? isn't the whole nick+userhost == nick!user@host limited to the sum of its lenghts, so we need to start with the parts of this nickuserhost... like... hostname should be 63 chars max (rfc).  nick is well.. the max length of the nickname we already have as an eggdrop setting (005 later) but ok. 32 is also a sane upper bound we dont have to care anymore, maybe that 32 fuer nickmax/handlen is the only value in eggdrop.h in this block, that actually makes sense. and user certainly has a limit as well, that we should code into eggdrop.h and all others can be sums of those...

UHOSTMAX is dual used now.
once it is used as length for user@host
and then it is also used for nick!user@host cases.
that is the problem rephrased, that needs to be solved here.

The use of UHOSTMAX and UHOSTLEN within eggdrop isnt clear
so we must go over all those places and check what is meant:
max or len
user@host or nick!user@host.

And propose to start from something like this:
```
#define NICKMAX 32 /* unchanged */
[...]

PR #1034, opened 2020-10-05
Fix truncation / buffer sizes

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix truncation / buffer sizes

Found with #1028, but can/should be merged independently before #1028.

it should be possible to demonstrate a truncated notice in https://github.com/eggheads/eggdrop/blob/d5eff2a548841471b5fbd95f2c54176accbad737/src/mod/irc.mod/msgcmds.c#L491 and to demonstrate that this PR will fix this. But it should be enough to know, that this PR will fix compiler warnings that would occur after #1028.


src/mod/irc.mod/msgcmds.c
```diff
@@ -351,7 +351,7 @@ static int msg_who(char *nick, char *host, struct userrec *u, char *par)
   struct chanset_t *chan;
   struct flag_record fr = { FR_GLOBAL | FR_CHAN, 0, 0, 0, 0, 0 };
   memberlist *m;
-  char s[UHOSTLEN], also[512], *info;
+  char s[NICKLEN + UHOSTLEN], also[512], *info;
```
vanosg, line 354: #define UHOSTMAX    291 + NICKMAX /* 32 (ident) + 3 (\0, !, @) + NICKMAX */
#define UHOSTLEN     UHOSTMAX + 1

Should be just UHOSTLEN?

src/mod/irc.mod/msgcmds.c
```diff
@@ -431,7 +431,7 @@ static int msg_who(char *nick, char *host, struct userrec *u, char *par)
 
 static int msg_whois(char *nick, char *host, struct userrec *u, char *par)
 {
-  char s[UHOSTLEN], s1[81], *s2;
+  char s[UHOSTLEN + 1], s1[143], *s2, stime[14];
```
vanosg, line 434: #define UHOSTMAX    291 + NICKMAX /* 32 (ident) + 3 (\0, !, @) + NICKMAX */
#define UHOSTLEN     UHOSTMAX + 1

Should be just UHOSTLEN?

vanosg (commented):
good catches! can you double-check these values and adjust if necessary?

src/mod/irc.mod/msgcmds.c
michaelortmann: In a sane world, yes. But eggdrops usage of UHOSTLEN and friends is borked. This here is an example of something that can be read quite often within eggdrop. Later in the code there is a line like:
`snprintf(s, sizeof s, "%s!%s", m->nick, m->userhost);`
_Note: If egg_snprintf would have been sprintf, the compiler would have catched it. This is what i am trying to to with #1028.
That snprintf cant fit into UHOSTLEN.
There is no better fix currently than oversizing the destination array.
In some distant future i hope to fix the real problem, but that would requite a quite large PR and some tests to make sure its stable. 
OK, in other words, eggdrop needs to define one UHOSTLEN without NICK, and one with NICK. That is currently not the case.

src/mod/irc.mod/msgcmds.c
michaelortmann: Similar to my comment above.
Here are the compiler warnings that would occur, if snprintf instead of egg_snprintf would have been used with a destination buffer of only UHOSTLEN:
```
gcc -fPIC -g -O2 -pipe -Wall -I. -I../../.. -I../../.. -I../../../src/mod  -DHAVE_CONFIG_H -I/usr/include -g3 -DDEBUG -DDEBUG_ASSERT -DDEBUG_MEM -DDEBUG_DNS  -DMAKING_MODS -c .././irc.mod/irc.c && mv -f irc.o ../
In file included from .././irc.mod/irc.c:69:
.././irc.mod/msgcmds.c: In function ‘msg_whois’:
.././irc.mod/msgcmds.c:465:37: warning: ‘snprintf’ output may be truncated before the last format character [-Wformat-truncation=]
  465 |         snprintf(s, sizeof s, "%s!%s", par, m->userhost);
      |                                     ^
.././irc.mod/msgcmds.c:465:9: note: ‘snprintf’ output 2 or more bytes (assuming 325) into a destination of size 324
  465 |         snprintf(s, sizeof s, "%s!%s", par, m->userhost);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.././irc.mod/msgcmds.c: In function ‘msg_who’:
.././irc.mod/msgcmds.c:385:31: warning: ‘%s’ directive output may be truncated writing up to 323 bytes into a region of size between 291 and 323 [-Wformat-truncation=]
  385 |     snprintf(s, sizeof s, "%s!%s", m->nick, m->userhost);
      |                               ^~
.././irc.mod/msgcmds.c:385:5: note: ‘snprintf’ output between 2 and 357 bytes into a destination of size 324
  385 |     snprintf(s, sizeof s, "%s!%s", m->nick, m->userhost);
      |
```

michaelortmann:
Geo: For the uhostlen mess see also #650. Many compiler warnings are currently shaddowed by using `egg_snprintf()` instead `snprintf()`. For that see also #1028. There is more to come, but i want to see this PR merged rather soon instead of making one big PR, because not all compiler warnings about buffer sizes / truncation are equal. This here is one of the more serious ones and not only cosmetic.

src/mod/irc.mod/msgcmds.c
vanosg: To summarize this in one line, m->userhost is defined as UHOSTLEN, even though UHOSTLEN is sized to allow the nick!user@host format not the user@host format we write to it. So in theory m->userhost could be longer than expected, even though we only ever write the user@host portion to it

PR #1042, opened 2020-10-09
Parse/modify oident.conf file

### LINKED ISSUE #942: Remove $HOME/.oidentd.conf after successful connection to server

vanosg:
Found by:
Patch by:
Fixes: #942 

One-line summary:
Parse and modify oident.conf entries

Instead of just overwriteing oident.conf, this PR will write Eggdrop entries to oident.conf in a specific format that will allow us to parse the file. This should allow us to co-exist with other programs using oident.conf, as well as help mitigate race conditions when using multiple bots from the same account. This will also check and prune the oident.conf file for 'stale' entries no longer needed based on a timestamp. 

The one-liner oident.conf entry looks like:
```to irc.com lport 5555 from myhost fport 6667 { reply "myident" } ### eggdrop_pidfile !1602208657```
where pidfile is the pidfile of the eggdrop, and the !number string is unixtime.

### DIFF
--- a/src/mod/ident.mod/ident.c
+++ b/src/mod/ident.mod/ident.c
@@ -171,14 +171,14 @@ static void ident_oidentd()
 putlog(LOG_MISC, "*", "family is %d\n", addr.addr.sa.sa_family);
     if (addr.addr.sa.sa_family == AF_INET) {
       fprintf(fd, "lport %i from %s { reply \"%s\" } "
-                "### eggdrop_%s !%ld\n", ntohs(addr.addr.s4.sin_port),
-                inet_ntop(AF_INET, &(addr.addr.s4.sin_addr), s, INET_ADDRSTRLEN),
-                botuser, pid_file, time(NULL));
+            "### eggdrop_%s !%ld\n", ntohs(addr.addr.s4.sin_port),
+            inet_ntop(AF_INET, &(addr.addr.s4.sin_addr), s, INET_ADDRSTRLEN),
+            botuser, pid_file, time(NULL));
     } else if (addr.addr.sa.sa_family == AF_INET6) {
       fprintf(fd, "lport %i from %s { reply \"%s\" } "
-                "### eggdrop_%s !%ld\n", ntohs(addr.addr.s6.sin6_port),
-                inet_ntop(AF_INET6, &(addr.addr.s6.sin6_addr), s, INET_ADDRSTRLEN),
-                botuser, pid_file, time(NULL));
+            "### eggdrop_%s !%ld\n", ntohs(addr.addr.s6.sin6_port),
+            inet_ntop(AF_INET6, &(addr.addr.s6.sin6_addr), s, INET6_ADDRSTRLEN),
+            botuser, pid_file, time(NULL));
     } else {
       putlog(LOG_DEBUG, "*", "IDENT: Error writing oident.conf line");
     }
--- a/src/mod/ident.mod/ident.c
+++ b/src/mod/ident.mod/ident.c
@@ -103,10 +103,14 @@ static struct dcc_table DCC_IDENTD = {
 static void ident_oidentd()
 {
   char *home = getenv("HOME");
-  char path[121], buf[(sizeof "global { reply \"\" }") + USERLEN];
-  int nbytes;
-  int fd;
+  FILE *fd;
+  long filesize;
+  char *data = NULL;
+  extern char pid_file[];
+  char path[121], line[256], buf[256], identstr[256];
+  int prevtime;
 
+  snprintf(identstr, sizeof identstr, "### eggdrop_%s", pid_file);
   if (!home) {
     putlog(LOG_MISC, "*",
            "Ident error: variable HOME is not in the current environment.");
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by michaelortmann** [comment]
Geo: getsockname() after connect() works. Demo code (ipv4 only and not polished):
https://github.com/eggheads/eggdrop/compare/develop...michaelortmann:geoident

This works, but has small glitch that could make it unusable. it will get the clients ip, but not as seen from the server. example, if you are behind NAT, you will have a local ip, like:

```
[23:19:50] DNS resolved irc.fu-berlin.de to 130.133.8.2
[23:19:50] net: open_telnet_raw(): idx 5 host irc.fu-berlin.de port 6667 ssl 0
Local ip      = 192.168.1.3
Local port is = 58775
```

first i thought that 192.168.1.3 is no use to us here, because its a localnet address behind NAT in my case. but oidentd is also running with this localnet ip. still, oident could compare this local address to a public ip address inside the ident packet data. i dunno. but it looks like this local ip may be a problem.

i dont know why oidentd should need our ip. **we better give oidentd only our port**. because any ip address inside the identd packet data would be what the server sees and thus probably not match any local ip. sure, a port could also be translated by some router/firewall between client/server, but we cant do anything about that case and it should be rare. NAT on the other hand is not rare. and a local port should be enough to disinguish between different eggdrops/irc clients on one machine. an additional ip would do us nothing better / useful.

now we have 2 options, both have ups and downs. overwrite the oidentd.conf like we do now without any ip/port, or set our (ip and) port like what this PR is for. **let the user be spoilt for choice**. i vote for an eggdrop option.

File: src/mod/ident.mod/ident.c [hunk 1]
```diff
@@ -116,14 +122,73 @@ static void ident_oidentd()
...
-    return;
+  fd = fopen(path, "r");
+  if (fd != NULL) {
+    /* Calculate file size for buffer */
+    if (fseek(fd, 0, SEEK_END) == 0) {
+      filesize = ftell(fd);
+      if (filesize == -1) {
+        putlog(LOG_MISC, "*", "IDENT: Error reading oident.conf");
+      }
+      data = nmalloc(sizeof(char) * (filesize + 256)); /* Room for Eggdrop adds */
```
michaelortmann: Please add `*data = 0;`
Else `strlen(data)` in line 142 does heap overflow

src/mod/ident.mod/ident.c
```diff
@@ -116,14 +122,73 @@ static void ident_oidentd()
...
+        putlog(LOG_MISC, "*", "IDENT: Error setting oident.conf file pointer");
+      }
+      while(fgets(line, 255, fd)) {
+        /* If it is not an Eggdrop entry, don't mess with it */
+        if (!strstr(line, "### eggdrop_")) {
+          strncat(data, line, ((filesize + 256) - strlen(data)));
+        } else {
+          /* If it is Eggdrop but not me, check for expiration and remove */
+          if (!strstr(line, identstr)) {
+            strncpy(buf, line, sizeof buf);
```
michaelortmann: Please use `strlcpy()` instead of `strncpy()` if possible.

src/mod/ident.mod/ident.c
```diff
@@ -103,9 +104,14 @@ static struct dcc_table DCC_IDENTD = {
 static void ident_oidentd()
 {
   char *home = getenv("HOME");
-  char path[121], buf[(sizeof "global { reply \"\" }") + USERLEN];
-  int nbytes;
-  int fd;
+  FILE *fd;
+  long filesize;
+  char *data = NULL;
+  char path[121], line[256], buf[256], identstr[256], s[INET_ADDRSTRLEN];
```
michaelortmann: Please declare `s` like this:
```
#ifdef IPV6
  char s[INET6_ADDRSTRLEN];
#else
  char s[INET_ADDRSTRLEN];
#endif
```
Else `inet_ntop(..., s, ...)` in line 180 does stack overflow

src/mod/ident.mod/ident.c
```diff
@@ -116,14 +122,73 @@ static void ident_oidentd()
...
+          }
+        }
+      }
+    }
+    fclose(fd);
+  } else {
+    putlog(LOG_MISC, "*", "IDENT: Error opening oident.conf for reading");
+  }
+  servidx = findanyidx(serv);
+  addr.addrlen = sizeof addr.addr.sa;
```
michaelortmann: In our case, getsockname() should not read but write to addr.addrlen, so i dont think we need this line.

michaelortmann: My comment was wrong here, and the cause of my confusion. Didnt see that getsockname() also reads from addr.addrlen. Geo is correcting the value right now... thx thommey.

src/mod/ident.mod/ident.c
```diff
@@ -116,14 +122,73 @@ static void ident_oidentd()
...
+      if (filesize == -1) {
+        putlog(LOG_MISC, "*", "IDENT: Error reading oident.conf");
+      }
+      data = nmalloc(sizeof(char) * (filesize + 256)); /* Room for Eggdrop adds */
+
+      /* Read the file into buffer */
+      if (fseek(fd, 0, SEEK_SET) != 0) {
+        putlog(LOG_MISC, "*", "IDENT: Error setting oident.conf file pointer");
+      }
+      while(fgets(line, 255, fd)) {
```
michaelortmann: Please consider changing this to:
while(fgets(line, sizeof line - 1, fd)) {

vanosg:
Testing:

* Compiles with --disable-ipv6
* When using an IPv4 IP, writes to file:
```lport 55035 from 192.168.1.163 { reply "myIdent" } ### eggdrop_pid.Lamestbot !1603332198```
* When using an IPv6 IP, writes to file:
```lport 47983 from 2b01:559:c010:3325::1 { reply "myIdent" } ### eggdrop_pid.Lamestbot !1603332361```
* Preserves non-Eggdrop lines
* Removes lines containing ### eggdrop_ that have a timestamp > 300
* Works with IPv4 behind a NAT (local IPs in .conf file)
* Works when starting multiple bots at same time

PR #1051, opened 2020-10-19
Remove the limit of 1024 chars for input/output of add_cr() and escape_telnet()

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Remove the limit of 1024 chars for input/output of add_cr() and escape_telnet()

Additionally refactor dccutil.c:add_cr() and dcc.c:escape_telnet() into a single misc.c:add_cr()

Connecting to the bot via telnet was successful and output was the same as before.


thommey:
why is this useful? aren't those putlog lines limited anyway?

michaelortmann:
> why is this useful? aren't those putlog lines limited anyway?

what about https://github.com/eggheads/eggdrop/blob/901c3a129ca5126b061786498c0fc63701eeed00/src/eggdrop.h#L62 ?

PR #1055, opened 2020-10-30
Feat/pbkdf2

### LINKED ISSUE #810: [1.8.4] password is limited to 16 characters

vanosg:
Found by: Various
Patch by: thommey and michaelortmann
Fixes: #810 #811

Adds PBKDF2 support for hashing. The intent/hope here is that someone will write a proper PASS2 "umbrella" module in the future which will serve as a register for modules such as PBKDF2, and pass values for hashing to the appropriate sub-module based on the PHC string format used for the hash (ie, $pbkdf2-SHA256$rounds=5000$hipwbiJ.... goes to PBKDF2 if loaded, $scrypt-SHA256$rounds=5000$hipwbiJ.... goes to scrypt if loaded, etc).

PASSWORDMAX is now 30. Rounds is defaulted to 16,000, default method is SHA256.

I won't re-hash the docs in this commit message, but it is important to note that this module was written for compatibility to be co-loaded with the blowfish module; it will maintain both the PASS and PASS2 fields. The expectation is that this feature will be removed for Eggdrop 2.0, WHICH IS DEFINITELY GOING TO HAPPEN IN MY LIFETIME. No co-load of blowfish = no maintenance of the PASS field.

If the method used for hashing changes (ie, SHA256 to SHA512), the hashes will be updated to the new method when the user next logs in.

There is no encrypt/decrypt functionality at this time; this is definitely a punt on our part, apologies to the community. Feel free to submit another PR :)


thommey:
blowfish.mod does 2 things: hash/verify passwords for PASS entries and provide Tcl encrypt/decrypt functions.

We want to enable users to load pbkdf2 and not use blowfish hashing anymore, while they still might need Tcl encrypt/decrypt (hashing and encryption are 2 separate things and it makes sense they are in separate modules now).

Proposal:
* blowfish is now optional, but still loaded by default to provide encrypt/decrypt (and a comment in the config warns about removing it in case there are scripts that use encrypt/decrypt - these would fail in a subtle way with Tcl errors)
* we need a new setting to decide if PASS is still generated and/or removed when we have PASS2, could default to 1 in 1.9 and 0 in 2.0, but it's necessary because loading blowfish does not mean I want it for passwords anymore. doing both is necessary for compatibility and transitioning

In the future we might also want to provide a different encrypt/decrypt version. That could be encrypt2/decrypt2 but the signature of the function (key plaintext/ciphertext) wouldn't change in an incompatible way even if we add more parameters (iv). Blowfish has already been modified to accept "magic key" prefixes "cbc:key" or "ecb:key", that could be a way forward where encrypt/decrypt is in the core and modules hook into their respective "magic key". Leaves the issue of subtle failure if a module is not present but that could be resolved by scripts explicitely using "loadmodule rsa" if they want rsa encryption.

michaelortmann:
That new config setting is the same as my king of the day option under Option 2.1 here: https://github.com/eggheads/eggdrop/pull/906#issuecomment-667529964

Implementation with #1064

michaelortmann:
pbkdf2.mod uses a different openssl detection. i could need some help here. build system is quite complex.

michaelortmann:
@vanosg:

Your commen "Rounds is defaulted to 16,000, default method is SHA256." is not correct. The default is still 5000. i changed eggdrop.conf to note that the default is 5000. But it may not be what you intended.

Your comment "If the method used for hashing changes (ie, SHA256 to SHA512), the hashes will be updated to the new method when the user next logs in." is also not correct. The default is currently pbkdf2_re_encode = 0.

doc/sphinx_source/mainDocs/pbkdf2.rst
```diff
@@ -0,0 +1,94 @@
...
+
+**************
+PBKDF2 Hashing
+**************
+
+    With the release of Eggdrop 1.9.0, the development team is releasing an updated crytopgraphy module, PBKDF2, for use with Eggdrop. PBKDF2 is a one-way hashing algorithm used to protect the contents of the password file, as well as for use via the Tcl interface.
+
+Background
+==========
+    Prior to Eggdrop 1.9.0, the blowfish module was included with Eggdrop to protect sensitve data such as passwords stored in the userfile. While there are no known practical attacks against blowfish at the time of this writing, it was decided that a more modern crypto solution was desirable to be included with Eggdrop. The PBKDF2 (Password-based Key Derivation Function 2) uses a password and salt value to create a password hash (the salt value ensures that the hashes of two identical passwords are different). This process is one-way, which means the hashes cannot be cryptographically reversed and thus are safe for storing in a file.
```
michaelortmann: typo: sensitve ==> sensitive

doc/sphinx_source/mainDocs/pbkdf2.rst
```diff
@@ -0,0 +1,94 @@
...
+
+#. Sit back and enjoy a cold beverage, you did it! Now encourage the rest of your users to log in so that their passwords are updated to the new format as well.
+
+Solo configuration
+------------------
+
+    With a solo configuration, Eggdrop will only run the pbkdf2 module. Eggdrop will not be able to authenticate against passwords in an already-existing userfile and thus will require every user to set a password again, as if they were just added to Eggdrop. This can be done via the PASS msg command (/msg bot PASS <password>) or by having a user with appropriate permissions (and an already-set password) log into the partyline and use the '.chpass' command.
+
+    SECURITY CONSIDERATION: This configuraion is not ideal for transitioning an existing userfile to PBKDF2. Without the blowfish module loaded, every user in the userfile essentially has no password set. This means any other user that matches a hostmask applied to a handle (*!*@*.aol.com, I'm looking at you) could set the password and gain access to that user's Eggdrop account.
```
michaelortmann: typo: configuraion ==> configuration

doc/sphinx_source/mainDocs/pbkdf2.rst
```diff
@@ -0,0 +1,94 @@
...
+
+    The default configuration of Eggdrop 1.9.0 has both the blowfish and pbkdf2 modules enabled (see `Hybrid Configuration`_ below). This will allow users upgrading a seamless transition to the PBKDF2 module. For users starting an Eggdrop for the first time, it is recommended to comment out the 'loadmodule blowfish' line, in order to implement the `Solo Configuration`_.
+
+    Also of note, the maximum password length is increased to 30 with PBKDF2, up from 15 with Blowfish. The automatically-generated botnet passwords are now 30 characters instead of the maximum-allowed 16 used with blowfish the blowfish module, and pull from a larger character set than what was used with the blowfish module. Finally, if you are linking bots, you'll need to ensure the same module is loaded on both bots (ie, if the hub bot is using the pbkdf2 module, the leaf bots must have pbkdf2 loaded as well in order to enable authentication checks).
```
michaelortmann: Word repeat typo: "blowfish the blowfish". Also "pull" should be past tense, shouldnt it?

PR #1064, opened 2020-11-06
PR against feat/pbkdf2: Add enable-pass config option

michaelortmann:
If enable-pass = 0 and encryption2 mod (pbkdf2) is loaded and encryption mod (blowfish) is loaded:

PASS is removed (and not created) when PASS2 is created
(PASS2 is created when new passwords are created or when PASS verification leads to automatic PASS2 creation)

PASS is removed while loading userfile or when enable-pass setting is changed from 1 to 0
(For each user that is not a bot and has PASS2 set)
(As above, only do this, if encryption mod (blowfish) is loaded)
(Bot cleartext passwords are always in PASS, never in PASS2)

PASS is only removed, if encryption mod (blowfish) is loaded as we discussed on IRC

share bots are not informed about removal of PASS, removals are propagated by next userfile sharing (relink) as we discussed on IRC

TODO: Open question: I am unsure, if i can untrace a tcl variable in its trace handler:
https://github.com/eggheads/eggdrop/pull/1064/files#diff-61a0df921bf6976d093fb69bb1e0ca4a626c66177a91a0d42dea41b6d6ea2e78R1131
when there are other code lines that also untrace:
https://github.com/eggheads/eggdrop/pull/1064/files#diff-2d8e3362ec4e40b7a6e562a1c7968e7b67063d05f639d9da1e0fa32b414c8706R527


src/userent.c
```diff
@@ -304,28 +305,35 @@ int pass_set(struct userrec *u, struct user_entry *e, void *buf)
...
     else {
-      if (encrypt_pass)
-        encrypt_pass(pass, new);
+      if (encrypt_pass && (!encrypt_pass2 || !remove_pass))
+          encrypt_pass(pass, new);
       if (encrypt_pass2)
         new2 = encrypt_pass2(pass);
     }
-    if (encrypt_pass) {
+    if ((u->flags & USER_BOT) || (encrypt_pass &&
```
vanosg: @thommey what's the verdict here?

michaelortmann: before and after introduction of pbkdf2: bot passwords are stored in cleartext. The code doing this before pbkdf2 is here:  https://github.com/eggheads/eggdrop/blob/5182a229e7042a75fcf2c3c5bbf9835a90bad90d/src/userent.c#L259. when i added pbkdf2 i did a mistake in my initial PR. If blowfish was loaded,  in my initial pbkd2 PR there was a bug, preventing a bot from saving a bot password if blowfish mod is not laded. This could be the case if only pbkdf2 mod is loaded. The Error line is this one: https://github.com/eggheads/eggdrop/pull/1055/files#diff-dd198bb0cb017512db4c27ee75144ae992e8c4c963b8d3c25b85dbdf785a7263R323.

michaelortmann:
Changed configure / build system:

```
./configure --disable-tls
[...]
./eggdrop [...]
Can't load modules pbkdf2: Initialization failure: configured with --disable-TLS or openssl not found
```

```
./configure --with-sslinc=/home/michael/opt/openssl-0.9.8zh/include/ --with-ssllib=/home/michael/opt/openssl-0.9.8zh/lib/
[...]
./eggdrop [...]
Can't load modules pbkdf2: Initialization failure: compiled with openssl version < 1.0.0
```

michaelortmann:
Status: WIP: pass_set() will be cleaned up and more tests will be carried out. "Soon" it will be ready for review again. If you don't look at pass_set(), and don't mind that the new code isn't well developer tested yet, you could start reviewing it now. changes to the code should be very minimal from now on.

michaelortmann:
Ready for test / review / merge into #1055.

PR #1068, opened 2020-11-10
(Cleanup) Cygwin: Eggdrop does not need FreeConsole() anymore

michaelortmann:
Found by: Travis Howell
Patch by: Travis Howell
Fixes: part of #176

One-line summary:
Eggdrop does not need FreeConsole() anymore

Windrop used that patch. I think they used it, because some time ago, the `FreeConsole()` behaviour in Cygwin was broken. Now it compiles and runs fine, but as i tested below, it works the same with and without `FreeConsole()`. So why not clean that old legacy code up.
Original patch here:
http://lists.eggheads.org/pipermail/patches/2001-March/000897.html
http://windrop.sourceforge.net/patches/eggdrop1.8.0.patch

I tested with and without FreeConsole() executed under eggdrops background mode under Cygwin 2.11.1 under Windows 7 with gcc 7.3.0. No difference. OK.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -35,7 +35,6 @@
  */
 #include <config.h>
 #ifdef CYGWIN_HACKS
-#  include <windows.h>
 #  undef X509_NAME
 #  undef X509_EXTENSIONS
 #  undef X509_CERT_PAIR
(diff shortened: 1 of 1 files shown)


michaelortmann:
cygwin console is fine, but cmd.exe hangs after:
```
eggdrop.exe-1.9.0 eggdrop.conf
exit
```

PR #1070, opened 2020-11-13
Cleanup bg_prepare_split(): we dont need goto here

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Cleanup `bg_prepare_split()`: we dont need `goto` here

There are good use cases for `goto`. But the use of `goto` in `bg_prepare_split()` was not. A simple `break ` has the same control flow. So `gogo` was replaced with `break`. While at it, another unnecessary line was removed. `break` after `exit` was unnecessary.

Compiles cleanly with `-Wimplicit-fallthrough`

### DIFF
--- a/src/bg.c
+++ b/src/bg.c
@@ -160,7 +160,6 @@ void bg_prepare_split(void)
       break;
     case BG_COMM_ABORT:
       exit(1);
-      break;
     case BG_COMM_TRANSFERPF:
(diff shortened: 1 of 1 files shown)


michaelortmann:
nevermind, i was doing a stupid thing and couldnt read the nesting properly. i blame it on missing indentation tho ;)

PR #1073, opened 2020-11-30
Fix server connect for --disable-ipv6

### LINKED ISSUE #1072: (eggdrop git) ./configure --disable-ipv6 -> broken .jump
Current git code 48e6ff154a93da4083db447d3f4b408a2ede662b is broken for `./configure --disable-ipv6`

`.jump <server>` results into:

`[22:09:19] Failed connect to <server> (Vhost set in config is not available on this machine)`

Eggdrop 1.8.4 is fine.

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #1072

One-line summary:
Fix server connect for `--disable-ipv6`

There were 2 bugs in eggdrop git version 48e6ff154a93da4083db447d3f4b408a2ede662b triggered by `./configure --disable-ipv6`. One bug was breaking server connect. The other one broke the logging of a port number. If server connection was refused the port was logged like:
`[23:44:06] net: attempted socket connection refused: 127.0.0.1:6667`
But the port was not converted to the correct byte order, so for example, instead of 6667 port 2842 was logged.
This PR fixes both bugs.

./configure --disable-ipv6
```
Before:
[23:42:16] Trying server 127.0.0.1:6667
[23:42:16] net: open_telnet_raw(): idx 4 host 127.0.0.1 port 6667 ssl 0
[23:42:16] Failed connect to 127.0.0.1 (Vhost set in config is not available on this machine)
```
After:
```
[23:44:29] Trying server 127.0.0.1:6667
[23:44:29] net: open_telnet_raw(): idx 4 host 127.0.0.1 port 6667 ssl 0
[23:44:29] net: connect! sock 9
[23:44:29] Connected to 127.0.0.1
```

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -55,7 +55,7 @@ extern unsigned long otraffic_irc_today, otraffic_bn_today, otraffic_dcc_today,
                      otraffic_unknown_today;
 
 char natip[121] = "";         /* Public IPv4 to report for systems behind NAT */
-char listen_ip[121] = ""; /* IP (or hostname) for listening sockets    */
+char listen_ip[121] = "";     /* IP (or hostname) for listening sockets       */
 char vhost[121] = "";         /* IPv4 vhost for outgoing connections          */
 #ifdef IPV6
 char vhost6[121] = "";        /* IPv6 vhost for outgoing connections          */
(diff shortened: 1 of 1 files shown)


michaelortmann:
The new code in getvhost() still looks unstable. Its not 100% clear to me yet, in which case setsockname() should be called with which parameters. We probably don't want to call it with allowres=1, if we give it an empty string to resolve. But there are more cases to cover.

vanosg:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #1072

PR #1085, opened 2020-12-31
Ignore config comments in listen cmd

vanosg:
Found by: Lord255
Patch by: Geo

Because a "comment" can go over BADRGS limit and makes for a confusing error,
we check for >ARGS at run. We can handle a comment in <ARGS later in the code.

Tested:
listen port all #comment
listen ip port all #comment
listen port script foo pub #comment
listen ip port script foo pub #comment  <- fatals

### DIFF
--- a/language/core.english.lang
+++ b/language/core.english.lang
@@ -429,4 +429,4 @@ Telnet to the bot and enter 'NEW' as your handle.
 0xe32,Refused telnet connection from %s (tried using my botnetnick)
 0xe33,Lost telnet connection from %s while checking for duplicate
 0xe34,TIMEOUT ident connection
-0xe36,Please remove the #comment from your listen setting and try again
+0xe35,Please remove the #comment from your listen setting and try again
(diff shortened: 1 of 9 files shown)


language/core.english.lang
```diff
@@ -429,3 +429,4 @@ Telnet to the bot and enter 'NEW' as your nickname.
 0xe32,Refused telnet connection from %s (tried using my botnetnick)
 0xe33,Lost telnet connection from %s while checking for duplicate
 0xe34,TIMEOUT ident connection
+0xe36,Please remove the #comment from your listen setting and try again
```
Cizzle: Should be 0xe35.

src/tcldcc.c
```diff
@@ -1295,7 +1308,8 @@ static int tcl_listen STDVAR
   }
   strlcpy(type, argv[i], sizeof(type));
 /* Check if mask or proc exists */
-  if (((argc>3) && !strlen(ip)) || ((argc >4) && strlen(ip))) {
+  if ((((argc>3) && !strlen(ip)) || ((argc >4) && strlen(ip))) &&
```
Cizzle: While you're touching this code, add some more improvements: ```!ip[0]``` instead of ```!strlen(ip)``` (and inverse) and use ```char ip[121] = ""``` at declaration to also get rid of the ```else``` clause at line 1287.
Lots of ```strlen```'s down below too.

Cizzle (changes requested):

PR #1087, opened 2021-01-03
Enhance logging - add millisecs (add %f specifier for timestamp-format)

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Add `%f` specifier for `timestamp-format`

I always wanted to have the possibility to let eggdrop log showing milliseconds. Often i found myself adding this while debugging. Last time i missed it was with ident race debugging. This PR adds `%f` as an additional format specifier for `timestamp-format`.

conf:
`set timestamp-format {[%H:%M:%S.%f]}`

```
$ ./eggdrop -t BotA.conf
[...]
*** -HQ joined the party line.
[22:19:02.736] main: entering loop
[...]
[22:19:39.804] #test6888: mode change '+t ' by BotA!michael@127.0.0.1
```

**Test `%f` vs. `%%f`:**

conf:
`set timestamp-format {[%H:%M:%S.%f]}:`
result:
`[12:57:54.481]`

conf:
`set timestamp-format {[%%H:%%M:%%S.%%f]}`
result:
`[%H:%M:%S.%f]`


src/misc.c
```diff
@@ -535,6 +537,16 @@ void putlog EGG_VARARGS_DEF(int, arg1)
   /* Create the timestamp */
   if (shtime) {
     strftime(stamp, sizeof(stamp) - 2, log_ts, t);
+    if ((f = strstr(stamp, "%f"))) { /* handle millisecond specifier %f */
```
thommey: This is not compatible with the normal strftime syntax where "%%f" would be a literal "%f" and should not be replaced

michaelortmann: thank you. implemented. ready for review again.

PR #1116, opened 2021-02-03
Add tcl script to sort .bots

### LINKED ISSUE #743: Feature add: List bots on partyline in Alpha order when doing .bots or .botinfo

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #743

One-line summary:
Add tcl script to sort .bots

### DIFF
--- a/eggdrop.conf
+++ b/eggdrop.conf
@@ -1744,3 +1744,7 @@ if {[info exists net-type]} {
     }
   }
 }
+
+# This script enhances Eggdrop's built-in dcc '.bots' command to output a sorted
+# list.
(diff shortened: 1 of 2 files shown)


vanosg:
If we were to implement this, it would probably be in source, not as a Tcl. This is a handy script though, we could host it on eggheads in the meantime. I'm glad you properly credited this code to the original author :frowning_face:

PR #1134, opened 2021-02-26
Cleanup irc.mod / string copy

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Cleanup irc.mod / string copy

No functional change.


thommey:
This isn't a simple cleanup, it's a change in behavior. newsplit modifies its argument string to split it, removing this buffer makes the functions modify the input.

michaelortmann:
thats fine. the parameter is modifyable. we dont need to copy before newsplit(). other functions in chan.c call newsplit() on the parameter without such copy also. example: https://github.com/eggheads/eggdrop/blob/develop/src/mod/irc.mod/chan.c#L1192

thommey:
It still has side-effects I have seen in the past, e.g. raw console log being mangled.
It might be warranted to always give the bind handler a copy of this string instead of the real raw string.

michaelortmann:
Some bind functions dont copy the origmsg and modify it. Some copy it before modifying. I see no problems in modifying without copy, so would strip all local copies. You seen problems without copy, so you would like to copy for all binds in advance. We could indeed argue to strip all local copies in favor of letting the bind call make the copy before calling the bind. In each case, i would like to see a general solution instead of the current state where sometimes its copied and sometimes its not.

Anyway, for now i opted to simplify this PR, to just replace strncpy value with strlcpy sizeof.

PR #1137, opened 2021-03-02
split of module conf by files.

ZarTek-Creole:
+ Separation of modules by files.
+ New 'loadmoduleconf' procedure to load the configuration file of the specific module.
+ Creation of the 'conf/defaults/modules' folder for the modules configuration files.
+ Rename eggdrop.conf and eggdrop-basic.conf to example.conf / example-basic.conf
+ Verification of configuration files by modules in different folders:
  * installdir (eggdrop install)
  * conf /
  * conf /modules/
  * conf /defaults/modules/ (default path)

Note: The example-basic.conf file is to remain the same. It loads the modules without module configuration.

Reason for changes:
- Load module variables only if the module is loaded (to relieve [info vars ::*])
- Guide the user more easily on the options/variables necessary for his needs.
- Different path for the configuration modules for the use of multi eggdrop.
- ...


vanosg:
At this time, there is not a desire to split the configuration files, but thank you for submitting this as a guide that can help any future work on this topic

Issue #1142, opened 2021-03-11
listen port already in use


t0000:
Hi,
in version 1.8.4 the bot automatically selected a new port when the other was occupied.

Example:

listen 17050 users
listen 14050 bots

if 17050 is already occupied, the bot has automatically changed to 17051

now the bot can no longer be started:

Tcl error in file 'eggdrop.conf':
Couldn't listen on port '17050' on the given address: Address already in use. Please check that the port is not already in use
    while executing
"listen 17050 users"

the same with "bots"

there are always times that an eggdrop restarts and the port is not directly released

vanosg:
Hi @t0000 , thanks for taking the time to register this with us. This was an intended change and functionality as part of 1.9.0; the counter-point to your report was that bots would begin to listen on ports not specified in the config file and thus, not be able to relink when they started back up.

Can you provide a little more info on what scenario you encountered this issue? Generally we've found modern OS's are able to properly handle releasing ports in most cases. Was this after a normal restart? As a result of a script? etc etc. Any info you can provide would give us some good context so we can make sure we're incorporating all viewpoints into our design decisions. Thanks!

michaelortmann:
eggdrop tries to set `SO_REUSEADDR` to the socket before bind()ing the listen port to prevent this issue:
https://github.com/eggheads/eggdrop/blob/9f6bbd60f40f869f46fa4042c40f8ba46c6d2560/src/net.c#L391
Maybe eggdrop could not set `SO_REUSEADDR`?
Unfortunately the debug output for a failed setting of `SO_REUSEADDR` is only visible if eggdrop is started with `-n` and not if it is started with `-t` (#627 is about the missing debug output issue):
https://github.com/eggheads/eggdrop/blob/9f6bbd60f40f869f46fa4042c40f8ba46c6d2560/src/net.c#L392

michaelortmann:
The main reason for "occupied port" aka EADDRINUSE (The specified address is already in use.) is TIME-WAIT.
TIME-WAIT is part of TCP protocol design, see https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Protocol_operation
For example under linux one could get information about connections in time-wait with `ss --numeric -o state time-wait`
Result would look like
```
Netid      Recv-Q      Send-Q            Local Address:Port             Peer Address:Port       Process                        
tcp        0           0                     127.0.0.1:6667                127.0.0.1:59974       timer:(timewait,51sec,0)
```
[...]

PR #1143, opened 2021-03-11
RFC: Add bind-timeout option to let eggdrop wait for useable listen port

michaelortmann:
Found by:
Patch by: michaelortmann
Fixes: 

One-line summary:
Workaround and/or solution to #1142

We removed trying port+1 port+2 port+3 if port is unavailable with #843. This PR adds a configureable timeout for eggdrop attempting to bind to an (still) unavailable port.

For the following test i blocked the listen port for 2 seconds:
```
$ grep bind-timeout BotA.conf 
set bind-timeout 5
```

```
$ ./eggdrop -n BotA.conf
[...]
[22:04:48] The specified address is already in use. Attempting to bind for 5 seconds.
[22:04:50] Listening for telnet connections on 0.0.0.0 port 3333 (all).
```

### DIFF
--- a/eggdrop.conf
+++ b/eggdrop.conf
@@ -277,6 +277,9 @@ set userfile-perm 0600
 # and set it to the nick you would like to use.
 #set botnet-nick "LlamaBot"
 
+# Bind timeout in seconds. Must be set before listen command. Default is 0.
+# set bind-timeout 300
+
 # This opens a telnet port by which you and other bots can interact with the
 # Eggdrop by telneting in. There are more options for the listen command in
 # doc/tcl-commands.doc. Note that if you are running more than one bot on the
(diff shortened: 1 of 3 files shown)


ZarTek-Creole:
Very good idea. The only negative point may be the name *bind-timeout*. because *bind* reminds me of a lot of things in eggdrop except the port listen. this is an unclear name for the listen timeout. I suggest instead *listen-timeout*?

michaelortmann:
Geo, thommey: what do you think about `listen-timeout`?

ZarTek-Creole (changes requested):
https://github.com/eggheads/eggdrop/pull/1143#issuecomment-797539947

Rename bind-timeout to listen-timeout

vanosg:
I don't think this feature would be used much, if at all. If a port isn't available at time X, it likely still wouldn't be available at time X+5 (or whatever the timeout is set to). At this time, I'm not in favor of adding such a feature

thommey:
I don't think this needs to be an option. The only case where the socket is blocked is if you restart eggdrop very quickly and the OS still blocks it, which is really quick and a sleep(1) should cover that, no? Or what problem are we trying to solve? Can it be solved with SO_REUSEADDR or SO_REUSEPORT?

PR #1147, opened 2021-03-18
RFC: Do not send escape codes to ssl connections

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Do not send escape codes to ssl connections for openssl does not interpret but print them.

If one telnets to the bot, the bot will send some telnet codes. For example to suppress echoing the password. If one connects via ssl, one usually uses openssl and openssl does not interpret those telnet codes but prints them to the screen. This looks ugly. I think, openssl is not the only way to connect to eggdrop via ssl, but i also dont know of any telnet-client able to connect to eggdrop via ssl. So, i need more opinions, if the `if (!dcc[i].ssl)` is a good idea.

Before:
```
$ openssl s_client -cert alice.crt -key alice.key -connect 127.0.0.1:3343
[...]
read R BLOCK
ÿû

Nickname.
^[[?1;2c
```
After:
```
$ openssl s_client -cert alice.crt -key alice.key -connect 127.0.0.1:3343
[...]
read R BLOCK


Nickname.
```

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -2344,7 +2344,8 @@ static void dcc_telnet_got_ident(int i, char *host)
   /* Note: we don't really care about telnet status here. We use the
    * STATUS option as a hopefully harmless way to detect if the other
    * side is a telnet client or not. */
-  dprintf(i, TLN_IAC_C TLN_WILL_C TLN_STATUS_C);
+  if (!dcc[i].ssl)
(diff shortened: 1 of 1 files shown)


thommey:
I agree, I could not find a telnet-over-ssl client. You would have to strip the SSL layer and then redirect to telnet specifically

PR #1161, opened 2021-04-29
Tracks away status as documented in the [isaway] tcl function.

DasBrain:
Found by: DasBrain
Patch by: DasBrain
Fixes: Tracks away status as documented in the [isaway] tcl function.

One-line summary: Tracks away status as documented in the [isaway] tcl function.

An AWAY message did not update the flags of the user, making [isaway] reporting the wrong result.


vanosg:
Commit 5b121864, ironically fixing a bad merge, accidentally/incorrectly removed this code. Thanks for the catch!

PR #1165, opened 2021-05-21
Add net-type Libera.Chat

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Add net-type Libera.Chat

For now, eggdrop will use freenode settings for net-type Libera.Chat. But we need this new seeting to make it super easy to setup eggdrops net-type. Also, both networks are already using difference irc servers, so settings may diverge in the future.


vanosg:
I think this was all incorporated as part of #1168 and #1171 , please let me know if that is not the case. Thanks!

PR #1173, opened 2021-06-09
Add MONITOR support

vanosg:
Add MONITOR support
https://ircv3.net/specs/extensions/monitor

Adds Tcl monitor command (add, delete, list, online, offline, clear)
'online' and 'offline' show which nicknames that are tracked are currently online and offline

Adds MONITOR Tcl bind


skralg (commented):
lgtm

PR #1183, opened 2021-08-02
String utilities

crazycatdevs:
Found by: CrazyCat
Patch by: CrazyCat
Fixes: N/A

One-line summary: small strings utilities I often use

* **ucfirst** :	Make a string's first character uppercase
* **ucwords** : Uppercase the first character of each word in a string
* **uclwords** : Uppercase the first character of each word in a string,


scripts/alltools.tcl
```diff
@@ -437,3 +448,24 @@ proc ordnumber {string} {
   }
   return $string
 }
+
+proc ucfirst {string} {
```
thommey: isn't this what [string totitle] in Tcl is for? It also lowercases the rest

crazycatdevs: It is, my bad...
I remove this useless proc

scripts/alltools.tcl
```diff
@@ -437,3 +445,17 @@ proc ordnumber {string} {
   }
   return $string
 }
+
+proc ucwords {string} {
+	list asent
```
thommey: This doesn't do anything, perhaps you want: `set asent [list]`

scripts/alltools.tcl
```diff
@@ -437,3 +445,17 @@ proc ordnumber {string} {
...
+	foreach word [split [join $string]] {
```
thommey: This has to be just `[split $string]` to iterate over the words, otherwise it breaks on special characters. `[split [join $string]]` does nothing because split and join do the exact opposite of each other

scripts/alltools.tcl
```diff
@@ -437,3 +445,17 @@ proc ordnumber {string} {
...
+		lappend asent [string totitle $word]
```
thommey: `string totitle` already lowercases the non-first characters, so ucwords and uclwords are identical right now. if you want to change the case of the first letter but leave uppercase in the rest, you cannot use `[string totitle]` for that

scripts/alltools.tcl
```diff
@@ -437,3 +445,17 @@ proc ordnumber {string} {
...
+	}
+	set string [join $asent]
+	return $string
+}
+
+proc uclwords {string} {
+	set string [ucwords [string tolower [join $string]]]
```
thommey: This `join` isn't right, `[string tolower $string]` is correct on its own. Join requires a list as argument

PR #1197, opened 2021-08-25
Update Readme : Change FreeNode to Libera

ZarTek-Creole:
Change FreeNode to Libera

### DIFF
--- a/README
+++ b/README
@@ -342,8 +342,8 @@ OBTAINING HELP
 
   You can obtain help with Eggdrop in the following IRC channels:
 
-    -   FreeNode - #eggdrop (official channel), #eggheads (development
-        discussion), #egghelp
+    -   Libera - #eggdrop (official channel), #eggheads (development
+        discussion)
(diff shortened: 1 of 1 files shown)


vanosg:
This file is generated by doc/sphinx-source/installAndSetup/readme.rst - please edit that file, and README will be updated from that. So please a) revert this change and b) edit the .rst file instead   and we'll get this merged. Thanks for finding this, and the submission!

ZarTek-Creole:
The change had already been made [here](https://github.com/eggheads/eggdrop/commit/c905f17a1f0afce317234f0c2f5071ca7cb264e9#diff-ccef1b60476c754948449f904dd6f520f83b3c50c9089e7deba91a4659d6cdeaR325)
Sphinx should not have been executed for the automatic update of the docs files
thank you for your reply

vanosg:
OK, then this should be good already- the file will be updated next time we regenerate documentation. Thanks!

PR #1212, opened 2021-10-20
Don't require message-tags to check for message-tags

### LINKED ISSUE #1179: Eggdrop doesn't process account-tag (and maybe others) unless the message-tags capability has been negotiated.
When enabling a capability that uses tags but without negotiating the `message-tags` capability eggdrop won't process any tag message.

Eggdrop should process tag messages regardless if the `message-tags` capability has been negotiated or not.

vanosg:
Found by: DasBrain, Lord255
Patch by: Geo 
Fixes: #1179 

One-line summary: 
Don't require message-tags to check for message-tags

account-tag, server-time and batch can all use message-tags without the message-tags capability actually being enabled. The previous implementation incorrectly assumed any capability requiring message-tags would... well, would require message-tags to be enabled. Eggdrop would then enter a bad state. This corrects the implementation to always check for message-tags present in a message, rather than only if message-tags is enabled.

account-tags works without message-tags enabled
```
.tcl cap enabled
Tcl: account-notify account-tag away-notify chghost extended-join invite-notify sasl
[20:02:18] [@] :Foo!~u@tffc6xfgjae7q.oragono ACCOUNT Bar
[20:02:18] Foo!~u@tffc6xfgjae7q.oragono has logged into account Bar
[20:02:18] triggering bind testproc
[20:02:18] Foo is Bar!
```

### DIFF
--- a/src/mod/server.mod/server.c
+++ b/src/mod/server.mod/server.c
@@ -2188,7 +2188,7 @@ static Function server_table[] = {
   /* 12 - 15 */
   (Function) match_my_nick,
   (Function) check_tcl_flud,
-  /* Empty */
+  (Function) NULL,              /* was msgtag in 1.9.0, 1.9.1           */
   (Function) & answer_ctcp,     /* int                                  */
   /* 16 - 19 */
   (Function) & trigger_on_ignore, /* int                                */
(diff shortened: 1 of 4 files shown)


skralg (commented):
lgtm

PR #1214, opened 2021-10-21
Fix clobbered var

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
fix clobbered var

found with gcc-12 and CFLAGS `-O1 -Wclobbered -ftrivial-auto-var-init=pattern`

Before:
```
/home/michael/opt/gcc-12-20211017/bin/gcc -O1 -Wclobbered -ftrivial-auto-var-init=pattern -pipe -Wall -I.. -I..  -DHAVE_CONFIG_H -I/usr/include -g3 -DDEBUG -DDEBUG_ASSERT -DDEBUG_MEM -DDEBUG_DNS  -c net.c
net.c: In function ‘setsockname’:
net.c:140:7: warning: variable ‘pref’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Wclobbered]
  140 |   int pref;
      |       ^~~~
```
After:
`/home/michael/opt/gcc-12-20211017/bin/gcc -O1 -Wclobbered -ftrivial-auto-var-init=pattern -pipe -Wall -I.. -I..  -DHAVE_CONFIG_H -I/usr/include -g3 -DDEBUG -DDEBUG_ASSERT -DDEBUG_MEM -DDEBUG_DNS  -c net.c`


skralg (commented):
lgtm

src/net.c
```diff
@@ -137,7 +137,7 @@ int setsockname(sockname_t *addr, char *src, int port, int allowres)
   volatile int af = AF_UNSPEC;
 #ifdef IPV6
   char ip2[INET6_ADDRSTRLEN];
-  int pref;
+  volatile int pref;
```
thommey, line 140: I was expecting a rename of the variable here, not "volatile", how does that fix the issue?

michaelortmann: see sigsetjmp(3)
https://www.freebsd.org/cgi/man.cgi?query=sigsetjmp&apropos=0&sektion=0&manpath=FreeBSD+13.1-RELEASE+and+Ports&arch=default&format=html:
```
All accessible objects have values	as of the time longjmp() routine was
     called, except that the values of objects of automatic storage invocation
     duration that do not have the volatile type and have been changed between
     the setjmp() invocation and longjmp() call	are indeterminate.
```
Linux man page:
```
       The compiler may optimize variables into registers, and longjmp() may restore the values of other registers in addi‐
       tion  to the stack pointer and program counter.  Consequently, the values of automatic variables are unspecified af‐
       ter a call to longjmp() if they meet all the following criteria:

       •  they are local to the function that made the corresponding setjmp() call;

       •  their values are changed between the calls to setjmp() and longjmp(); and

       •  they are not declared as volatile.
```

thommey:
thanks for the explanation, makes sense, LGTM

PR #1216, opened 2021-10-25
Fix several net-type bugs

### LINKED ISSUE #1218: net-type bug
i didnt look up when exactly the bug was introduced.
the problem is, when we do set (or change) net-type,
eggdrop has two traces on that variable.
one in irc.c, another one in server.c (2 different modules)
whatever the first trace is, when it exits the trace functionwithout error, then net_type is set.
net_type_int is another variable, set in the server.c trace
usualy, the irc.c trace is executed before the server.c trace
but the irc.c is the one, setting lots of internal variables in respect of the new/current net_type_int
now, if we change net-type, irc.c doest see the new net_type_int
this is a kinda serious bug, breaking our net-type handling completely, if im correct.

demo:

```
.set net-type efnet
Ok, set.

.set net-type efnet
Ok, set.

.set use-354
Currently: 0

.set net-type libera
Ok, set.

.set use-354
Currently: 0

.set net-type libera
Ok, set.

.set use-354
Currently: 1
```

we probably should try harder to only set one single trace function onto a variable to avoid such issues.

my first idea was, removing the trace in irc.c and exporting do_nettype() thru module API and then let the server.c trace call irc.c:do_nettype() instead.

but cant call from server.c into irc.c because irc mod depends on server.mod, thus when server.mod loads and executes its traces first, the irc mod isnt loaded yet.

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #1218

One-line summary:
fix several `net-type` bugs

`net-type` must be traced only in irc.mod and from there server.mod is called
in case of error, the `net_type` and `net_type_int` variables must not be changed nor  `Ok, set.` must be logged
`fatal()` must only be called during bot startup

**before:**
```
.set net-type 6
[09:41:11] tcl: builtin dcc call: *dcc:set -HQ 1 net-type 6
[09:41:11] #-HQ# set net-type 6
[09:41:11] * ERROR: NET-TYPE NOT SET.
 Must be one of DALNet, EFnet, freenode, Libera, IRCnet, Quakenet, Rizon, Undernet, Other.
```
here `fatal()` lets eggdrop die
**after:**
```
.set net-type 6
[12:17:14] tcl: builtin dcc call: *dcc:set -HQ 1 net-type 6
[12:17:14] #-HQ# set net-type 6
Error: can't set "net-type": Must be one of DALNet, EFnet, freenode, Libera, IRCnet, Quakenet, Rizon, Undernet, Other.
```


vanosg:
@thommey comments? Is this still valid?

PR #1217, opened 2021-10-25
fix char array len for listen-addr

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
listen-addr can be ip or hostname with max len `255 + '\0'`


vanosg:
Upon discussion, it looks like this is not required. Closing per request.

PR #1223, opened 2021-11-06
Use nick/altnick/randnick on bad nickname

### LINKED ISSUE #1201: Investigate invalid nick process on registration
If an eggdrop joins a server and attempts to register with an invalid nick, it looks at keep-nick in the config. If keep-nick is enabled, it will not try another nickname and end up disconnecting. If keep-nick is not enabled, it will generate a random nickname to continue registration with.

The question was posed in channel, should Eggdrop try to use the $altnick nickname, instead of a random nickname?

After some thought, I think I would lean towards saying 'no', a random nick is fine in order to ensure you are able to connect to the server. But, I think a log message to the partyline indicating what is going on behind the curtain might help users who are expecting the nickname "##MYBOT##" to join a channel, but end up with the nickname "sifnhjlper" joining instead. And keep-nick controlling this functionality is not intuitive to check, so probably need to mention that somewhere in there, too.

Thoughts?

vanosg:
Fixes:  #1201 

One-line summary:
Use nick/altnick/randnick on bad nickname

See previous discussion in issue- this PR removes reliance on keep-nick, then (if nick is bad) tries altnick, and THEN a random nick.

```
[14:11:33] [m->] NICK Bad???!
[14:11:34] [@] :calcium.libera.chat 432 * Bad???! :Erroneous Nickname
[14:11:34] Server says my nickname is invalid.
[14:11:34] NICK IS INVALID: 'Bad???!' (using 'MoBad!!??' instead)
[14:11:34] [!m] NICK MoBad!!??
[14:11:35] [m->] NICK MoBad!!??
[14:11:35] [@] :calcium.libera.chat 432 * MoBad!!?? :Erroneous Nickname
[14:11:35] Server says my nickname is invalid.
[14:11:35] NICK IS INVALID: 'MoBad!!??' (using 'peaidhidjfqqjhsv' instead)
[14:11:35] [!m] NICK peaidhidjfqqjhsv
```

### DIFF
--- a/src/mod/server.mod/servmsg.c
+++ b/src/mod/server.mod/servmsg.c
@@ -878,7 +878,7 @@ static int got432(char *from, char *msg)
   else {
     putlog(LOG_MISC, "*", IRC_BADBOTNICK);
     if (!strcmp(erroneous, origbotname)) {
-      strlcpy(nick, altnick, sizeof nick);
+      strlcpy(nick, get_altbotnick(), sizeof nick);
     } else {
       make_rand_str_from_chars(nick, sizeof nick - 1, CHARSET_LOWER_ALPHA);
     }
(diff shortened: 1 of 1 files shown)


michaelortmann:
instead of using altnick directly, we must use get_altbotnick() because if the '?'-magick

PR #1228, opened 2021-11-07
fix disable-ipv6 and buf size

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
fix disable-ipv6 and buf size

**1. fix disable-ipv6:**
`INET6_ADDRSTRLEN` was used also when configured with `--disable-ipv6`. This bug was not noticed, because `INET6_ADDRSTRLEN` is defined even when configured with `--disable-ipv6` on ipv6 able systems but it could break ipv4 only systems. While at it i noticed that we have a lot of redundant code like:
```
#ifdef IPV6
  char s[INET6_ADDRSTRLEN];
#else
  char s[INET_ADDRSTRLEN];
```
so i defined `EGG_INET_ADDRSTRLEN` in eggdrop.h depending on IPV6 so we use only that and and never fall for this bug again.
**2. fix buf size:**
There is a string truncation due to a too small buffer:
https://github.com/eggheads/eggdrop/blob/d6f53d9a6a9b07dc92b4beb43bb7baad677b5fb8/src/mod/filesys.mod/filedb3.c#L1086
introduced by:
https://github.com/eggheads/eggdrop/commit/aa8ef9ccb52826f3a1cf626cc25309af08acfa20
or lurking around even longer, if the old hardcoded value was not enough either. Its hard to calculate.
So i didnt waste much time on calculating the correct value but re-hardcoded it, but this time to 128 bytes, more than enough.
This bug was covered by the use of the (dangerous?) `simple_sprintf()`, so i changed it into `snprintf()`
I also took care that `snprintf()`s dst buffer is distinct to its src buffer.

I would like someone else to test this PR, for i am not fluent with the filesys mod
The filesys mod looks like it could need some testing anyway


vanosg:
First, this is a REALLY nice writeup, well done!

While clever, I'm not sure I'm in favor of the new defines you added as part of #1. You cite 'redundant' code, but for me it gives me clarity and confidence in the value being used at any given time and enhances the readability of the code, and I am more concerned with readability and "simple" coding as a general rule. But, I absolutely do see the value of having a single variable defined that 'just works', although it makes it harder for a coder to know whats in it with a quick glance. 

Does this affect every *_ADDRLEN  in the code base, or just some of them that you saw?

src/mod/filesys.mod/filedb3.c
```diff
@@ -1076,18 +1076,11 @@ static void remote_filereq(int idx, char *from, char *file)
...
-  s = nmalloc(INET6_ADDRSTRLEN);
-  getdccaddr(&dcc[i].sockname, s, INET6_ADDRSTRLEN);
-#else
-  s = nmalloc(INET_ADDRSTRLEN);
-  getdccaddr(&dcc[i].sockname, s, INET_ADDRSTRLEN);
-#endif
-  simple_sprintf(s, "%s %u %d", s, dcc[i].port, dcc[i].u.xfer->length);
-  botnet_send_filesend(idx, s1, from, s);
+  getdccaddr(&dcc[i].sockname, s, sizeof s);
+  snprintf(s2, sizeof s2, "%s %u %lu", s, dcc[i].port, dcc[i].u.xfer->length);
```
vanosg, line 1080: 46 + 1 + 4 + 1 + 4 +1= 57   if my public math is right?

michaelortmann:
i hope i covered any last location. that was one of the main ideas.

michaelortmann:
there are probably locations, that are ipv4 only anyway, but thats another story.

thommey:
good change!

src/mod/filesys.mod/filedb3.c
vanosg: I think my question here was, why is s2 128, instead of 57?

PR #1232, opened 2021-11-09
rip out remaining Context from modules

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Remove `Context`s that are artifacts, that dont help us debug eggdrop, instead they are even misleading.

For example, when the twitch mod does `Context;` and (much) later the bot crashes, it will point into twitch mod, where the last `Context;` was set, even if the bug is outside twitch mod. This does mislead debugging. Most `Context`s are already removed from eggdrop, but some remained. This PR clears the most annoying ones left. The twitch mod probably got its `Context` copied over from our example-mod woobie.


skralg:
lgtm

PR #1233, opened 2021-11-09
(Low Prio) Cleanup after we removed inline keywords

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Cleanup after we removed inline keywords


skralg:
looks fine

PR #1234, opened 2021-11-09
Fix possible string truncation

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix possible string truncation for logging TRANSFER_DCC_CONN

### DIFF
--- a/src/mod/transfer.mod/transfer.c
+++ b/src/mod/transfer.mod/transfer.c
@@ -850,7 +850,7 @@ static struct dcc_table DCC_GET_PENDING = {
 
 static void dcc_fork_send(int idx, char *x, int y)
 {
-  char s1[121];
+  char s[NICKMAX + 1 + UHOSTMAX + 1];
 
   if (dcc[idx].type != &DCC_FORK_SEND)
     return;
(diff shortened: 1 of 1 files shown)


src/mod/transfer.mod/transfer.c
```diff
@@ -850,7 +850,7 @@ static struct dcc_table DCC_GET_PENDING = {
 
 static void dcc_fork_send(int idx, char *x, int y)
 {
-  char s1[121];
+  char s[NICKMAX + 1 + UHOSTMAX + 1];
```
skralg, line 853: I would like to see comments that explain the math

michaelortmann: I would ike to explain here and not in code, if youre fine with that.
the math goes like:
`"%s!%s", dcc[idx].nick, dcc[idx].host`
which is:
`maxstrlen(dcc[idx].nick) + strlen("!") + maxstrlen(dcc[idx].host) + strlen(null-terminator)`
which is:
`(sizeof dcc[idx].nick minus null-terminator) + 1 + (sizeof dcc[idx].host minus null-terminator) + 1`

`sizeof dcc[idx].nick` `sizeof dcc[idx].host` are is defined here:
https://github.com/eggheads/eggdrop/blob/d6f53d9a6a9b07dc92b4beb43bb7baad677b5fb8/src/eggdrop.h#L386-L387

minus NULL-terminator is defined here:
https://github.com/eggheads/eggdrop/blob/d6f53d9a6a9b07dc92b4beb43bb7baad677b5fb8/src/eggdrop.h#L83-L84

vanosg: in far simpler words, 

'nick'+'!'+'ident@host.com'+'\0'   

:P

michaelortmann: i want to stuff skralg with so many information, that he soon will get the grip on eggdrops source and its abouts and be a new major contributor to the source code, and helps us solve dozens of our Issues, so that later we can invote him to our secret release parties where we drink all of beerbots beer.

michaelortmann: mistyping "invoke" instead of "invite" would have improved the sentence for my taste ;)

skralg:
I wasn't asking for an explanation for myself, I was asking for an explanation to be put in a comment in the code so others know what is going on there.

PR #1236, opened 2021-11-09
(Low Prio) Cleanup: simplify sprintf()/egg_snprintf() -> strlcpy()

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Cleanup: simplify sprintf()/egg_snprintf() -> strlcpy()

Because there is no string formatting

### DIFF
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -316,7 +316,7 @@ void tell_verbose_status(int idx)
   }
   cputime = getcputime();
   if (cputime < 0)
-    sprintf(s2, "CPU: unknown");
+    strlcpy(s2, "CPU: unknown", sizeof s2);
   else {
     hr = cputime / 60;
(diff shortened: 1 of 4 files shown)


skralg:
Would just changing sprintf to printf do the same thing? Or am I mixing up languages again?

michaelortmann:
no, that would not do.

printf() writes output to stdout

sprintf() writes output to a character string

see:

https://www.freebsd.org/cgi/man.cgi?query=printf&apropos=0&sektion=3&manpath=FreeBSD+13.0-RELEASE+and+Ports&arch=default&format=html

skralg:
Oh, right. I was thinking there was a variant that didn't do interpolations, but that's really what strlcpy does.
That was a derp moment.

PR #1238, opened 2021-11-12
Replace a hardcoded 81 with NICKLEN

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
At fist, this PR replaced a hardcoded 81 with NICKLEN
But when it turned out, it can also be a hostname:
Now this PR will set it to MAX(NICK/HOST) which is max host len = 256
we dont have / need a define for that, its hardcoded to 256 everywhere, so lets keep it simple.
This PR may fix a hidded bug with chan flood detection when hosts were longer than 80 chars.


skralg:
lgtm

thommey:
This can be a host, not just a nickname, see usage in src/mod/irc.mod/chan.c:detect_chan_flood()

michaelortmann:
Thanks thommey. Fixed. Ready for review again.

PR #1257, opened 2022-01-25
Added -n option explanation

crazycatdevs:
-t and -c reference -n but it wasn't explained

Found by:
Patch by:
Fixes: 

One-line summary:

### DIFF
--- a/README
+++ b/README
@@ -131,6 +131,8 @@ COMMAND LINE
 
   The options available are:
 
+    -n: Don't background, send all log entries to console.
+    
     -t: Don't background, use terminal. This is just like -n, except that
 
         instead of seeing log entries, your console will simulate a DCC


vanosg:
The -n option has been removed; it is no longer a valid option (and never really did anything). We don't error on using it because some people have 20 years of habit typing it in :)

```
./eggdrop -nt patchtest.conf 

Eggdrop v1.9.2+alpha (C) 1997 Robey Pointer (C) 1999-2021 Eggheads
--- Loading eggdrop v1.9.2+alpha (Wed Jan 26 2022)        
<snip>
Writing channel file...
Userfile loaded, unpacking...
=== testbot: 1 channels, 1 users.

NOTE: The -n flag is no longer used, it is as effective as Han
      without Chewie

### ENTERING DCC CHAT SIMULATION ###
You can use the .su command to log into your Eggdrop account.
```

PR #1273, opened 2022-03-07
Proofread certificate in documentation

dereckson:
Found by: dereckson
Patch by: dereckson
Fixes: 

One-line summary:
s/cerfificate/certificate/


dereckson:
Context: I was proofreading a infrastructure as code repository and I noticed this in an eggdrop configuration file, so I've checked the upstream code base too. It's been fixed in config comment for ssl-verify-clients, but it's still alive elsewhere in the doc.

PR #1286, opened 2022-06-01
Add timers with names

### LINKED ISSUE #1285: (Improvement timer system) : Naming of utimers and timers
Currently the syntax is like this:
```
timer <minutes> <tcl-command> [count]
utimer <seconds> <tcl-command> [count]
killtimer <timerID>
killutimer <timerID>
```

I propose to have the possibility of naming them in an optional way:

```
timer [name] <minutes> <tcl-command> [count]
utimer [name] <seconds> <tcl-command> [count]
killtimer <timerID/timerName/partialname*>
killutimer <timerID/timerName/partialname*>
```

What is the point of naming the timers?
The first, it would be easier to kill them if we know the name.

The second, for a reason X or Y we could replace the timer. if we use the same name the eggdrop will kill the old one and recreate the new one.

In some cases this is very useful. In mrc scripting the timers are nameable and this opens up a greater number of possibilities.

A third thing with naming timers:
Imagine that my script is called ZarTeK and in it I create 3 timers for example:
`ZarTeK_1, ZarTeK_2 and ZarTeK_3`

I want to kill the 3 when unloading my script or at the end of the game if it's a game..
I just need to do a `killtimer ZarTeK_*`

if the first argument to timer is a digit it is an unnamed user; if the first is a digit value then it is a timer with a name.

vanosg:
Adds the ability to set a custom name when creating timers/utimers. Names are displayed when the eggdrop Tcl 'timers' command is run with the -names flag. Multiple timers with the same name cannot be added. Tcl commands killtimername and killtimeruname are added to kill timers with a provided name. Using -name requires the count positional variable to be specified. #1285 

Test cases:

.tcl timer 10 foo 1
Tcl: timer1
.tcl timer 11 goo 2 myTimer
Tcl: timer2

.tcl timers
Tcl: {11 goo timer2 2} {10 foo timer1 1} 

.tcl timers -names
Tcl: {11 goo timer2 2 myTimer} {10 foo timer1 1 {}}

.tcl killtimername mytimer
Tcl: 
.tcl timers -names
Tcl: {10 foo timer1 1 {}}

### DIFF
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -616,7 +616,7 @@ int remove_timer_name(tcl_timer_t **stack, char *name)
   int ok = 0;
 
   while (*stack) {
-    if (!strcasecmp((*stack)->name, name)) {
+    if (((*stack)->name) && (!strcasecmp((*stack)->name, name))) {
       ok++;
       remove_timer_from_list(stack);
     } else {
@@ -711,7 +711,7 @@ void list_timers(Tcl_Interp *irp, tcl_timer_t *stack, int names)
   }
 }
 
-/* Find a timer by name
+/* Find a timer by name. Returns 1 if found, 0 if not
  */
 int find_timer(tcl_timer_t *stack, char *name)
 {
@@ -720,11 +720,11 @@ int find_timer(tcl_timer_t *stack, char *name)
   for (mark = stack; mark; mark = mark->next) {
     if (mark->name) {
       if (!strcasecmp(mark->name, name)) {
-        return 0;
+        return 1;
       }
     }
   }
-  return 1;
+  return 0;
 }
 
 int isowner(char *name) {
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -676,7 +676,7 @@ void wipe_timers(Tcl_Interp *irp, tcl_timer_t **stack)
 
 /* Return list of timers
  */
-void list_timers(Tcl_Interp *irp, tcl_timer_t *stack, int names)
+void list_timers(Tcl_Interp *irp, tcl_timer_t *stack)
 {
   char mins[11], count[11], id[26], *x;
   EGG_CONST char *argv[5];
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -554,7 +554,7 @@ void rehash()
 /* Add a timer
  */
 unsigned long add_timer(tcl_timer_t ** stack, int elapse, int count,
-                        char *cmd, unsigned long prev_id)
+                        char *cmd, char *name, unsigned long prev_id)
 {
   tcl_timer_t *old = (*stack);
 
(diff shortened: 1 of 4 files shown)


thommey:
Would it be possible to use the name as its ID? So that "timer123" is simply called what the user specified. That would simplify working with these timers a lot because you could just do something like:
```tcl
timer -name test [list putmsg #test hi]
killtimer test
```
I mean, there is no real value in keeping "timer123" and it could simply be replaced with the user-supplied identifier.

vanosg:
Are there any backwards-compatibility issues that could arise? We can hack at the code more, but right now the timer ID is hardcoded to return timer%s as a the ID string.. I can't think of any, but I didn't mess with that because I'm sure review wasn't exhaustive

thommey:
What the actual timer names are was never part of the API specification and I doubt anyone relies on that. I believe Tcl made a similar change with their "after" IDs. If it's possible to simply change all user-facing "timerXXX" by the name if they have a name (they can still also keep their ID, that becomes an internal detail then) that sounds like a good change to me.

ZarTek-Creole:
I think thommey is right. This will reduce the number of commands. In my ticket I proposed to modify the existing utimer and timer commands.
```
timer [name] <minutes> <tcl-command> [count]
utimer [name] <seconds> <tcl-command> [count]
```
if the argument after (u)timer is a number, then it is an unnamed (u)timer. If it contains other things than [0-9] it is a (u)timer name.


> Multiple timers with the same name cannot be added. 

I propose that it replaces the previous (u)timer with the same name. for the (u)timer names, the return could be 1, -1
1 = timer name added
-1 = timer name replace.

For the issue of backwards compatibility. imagine that a script relies on "timerXXX"; if we stay logical, they will continue to work normally, because they don't use named timers, and unnamed timers can remain in "timerXXX" format without any problem

vanosg:
ok, I looked at it a little more and yes, just changing the string from timer%s  to a user-defined one looks like it won't cause any global issues and allow us to overwrite the [timers] name value. For now (and likely forever because no one will ever look at this again) I'm going to leave the -> id value as part of the struct and, if no name is specified, write timer%s to ->name, and output $name in the timers command. This should also allow us to *sniff* not need the kill(u)timername commands.

vanosg:
current functionality:
```
.tcl timer 10 foo 1
Tcl: timer1
.tcl timer 11 goo 1
Tcl: timer2
.tcl timer 12 moo 1 myTimer
Tcl: myTimer
.tcl timers
Tcl: {12 moo myTimer 1 } {11 goo timer2 1 } {10 foo timer1 1 }

.tcl killtimer timer2
Tcl: 
.tcl timers
Tcl: {12 moo myTimer 1 } {10 foo timer1 1 }
.tcl killtimer mytimer
Tcl: 
.tcl timers
Tcl: {9 foo timer1 1 }

.tcl timer 10 foo 3 timer2
Tcl error: timer name may not begin with "timer"

.tcl timer 10 foo 1 mytimer
Tcl: mytimer
.tcl timer 10 goo 2 mytimer
Tcl error: timer already exists by that name
```

src/tclmisc.c
```diff
@@ -248,48 +248,72 @@ static int tcl_binds STDVAR
...
+        Tcl_AppendResult(irp, "timer name may not begin with \"timer\"", NULL);
+        return TCL_ERROR;
+      }
+      /* Check for existing timers by same name */
+      if (find_timer(timer, argv[4])) {
+        Tcl_AppendResult(irp, "timer already exists by that name", NULL);
+        return TCL_ERROR;
+      }
+    }
+    x = add_timer(&timer, atoi(argv[1]), (argc >= 4 ? atoi(argv[3]) : 1),
```
thommey: While rewriting this, it would be good to do some error checking there on that atoi(). count 0 is invalid, and atoi("test") is also 0, so 0 should be rejected

src/tclmisc.c
```diff
@@ -248,48 +248,72 @@ static int tcl_binds STDVAR
...
+                  argv[2], (argc == 5 ? argv[4] : '\0'), 0L);
```
thommey: argv[4] is char* but '\0' is char, which should be an error. you can use "" for the empty string or pass NULL if that works

src/chanprog.c
```diff
@@ -571,25 +574,48 @@ unsigned long add_timer(tcl_timer_t ** stack, int elapse, int count,
...
+
+/* Remove a timer (via name, not ID)
+ */
+int remove_timer(tcl_timer_t **stack, char *name)
+{
   int ok = 0;
 
   while (*stack) {
-    if ((*stack)->id == id) {
+    if (((*stack)->name) && (!strcasecmp((*stack)->name, name))) {
```
thommey: would prefer with fewer ():
if ((*stack)->name && !strcasecmp())

src/tclmisc.c
vanosg: 0 is valid input (infinite counter). So [timer 1 foo test] would create an infinite counter. I don't know that we can error check enough to distinguish between user intents

thommey: strtol instead of atoi, the man page for strtol describes how exactly to determine if something is a valid number, we might have examples of that in our code

vanosg:
Added a check for count as an integer
```
.tcl timer 1 foo goo
Tcl error: count value must be >=0
.tcl timer 1 foo 1
Tcl: timer3
.tcl timer 1 goo 0
Tcl: timer4
```

PR #1298, opened 2022-06-16
Update port tls to 6697 - rfc7194

ZarTek-Creole:
Found by: ZarTeK
Patch by: ZarTeK
Fixes: port tls per default is 6697 and not 7000 in rfc7194

One-line summary:


vanosg:
Thanks; this is just an example and not an rfc definition so I think we're OK with how it is. Thanks for submitting though

PR #1302, opened 2022-06-23
(WIP) Enhance account tracking abilities

### LINKED ISSUE #1369: Fully implement lookup_by_account functionality

vanosg:
a PR that tracks adding get_user_by_account() where appropriate, to complement get_user_by_host(). This will allow Eggdrop to track a user by their account name (first), and then check a hostname. The assumption here is that account names are more precise and less prone to collisions/spoofing than a hostmask.


vanosg:
update for all channels- remove strlcpy
Does this break Tcl binds (the userhost arg passed to a function?)

vanosg:
#1369 tracking

PR #1308, opened 2022-07-19
Update ssl.conf

TehPeGaSuS:
Most of the generators I know, use 3650 days (~10 years) for self-signed certificates. IMHO, eggdrop could do the same.
 As someone stated on IRC: 
  - 10 year is the usual "next employee's problem" duration

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -313,10 +313,10 @@ sslstart:
 	fi
 
 sslcert: sslstart
-	openssl req -new -x509 -nodes -days 365 -keyout $(DEST)/eggdrop.key -out $(DEST)/eggdrop.crt -config ssl.conf
+	openssl req -new -x509 -nodes -days 3650 -keyout $(DEST)/eggdrop.key -out $(DEST)/eggdrop.crt -config ssl.conf
 
 sslsilent: sslstart
-	openssl req -new -x509 -nodes -days 365 -keyout $(DEST)/eggdrop.key -out $(DEST)/eggdrop.crt -config ssl.conf \
+	openssl req -new -x509 -nodes -days 3650 -keyout $(DEST)/eggdrop.key -out $(DEST)/eggdrop.crt -config ssl.conf \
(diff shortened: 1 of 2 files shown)


ZarTek-Creole (changes requested):
`-days` is useless if already set in ssl.conf file

vanosg:
This change moves in the opposite direction of the security community at large. Granted, the use cases of Eggdrop are different from webpages, but the three-year expiry matched the community standard when it was authored. Since then (2018), certificate authorities have changed that standard from three years to two years (with some bodies pushing to change to just one year), largely in part to lessons learned from the deprecation of SHA-1 as a signing algorithm and, iirc, moving away from 1024-bit RSA keys. There are pros and cons to this; and while I don't think we necessarily need to change from 3 years to 2 years either, I think inserting a practice that arguably has a negative impact on long-term security and goes against the larger community as  default setting is not a step that needs to be taken. If a user wishes to change it on their own, that is certainly their choice and can easily be done (and heck, this would be a perfect subject to add to the wiki if one so chose!)

EDIT: ok, so I looked up the actual date right as I posted this- turns out the industry again shortened the expiration period to 1 year from 2 years in 2020

TehPeGaSuS:
> EDIT: ok, so I looked up the actual date right as I posted this- turns out the industry again shortened the expiration period to 1 year from 2 years in 2020

Since  I'm not a fan of issues `lingering around` and understanding the security concerns, I deem this closed without further changes.

ZarTek-Creole:
> `-days` is useless if already set in ssl.conf file

Anyway, on the duration of the certificates, I think the [-days in makefile.in](https://github.com/eggheads/eggdrop/pull/1308/commits/ced615957aa7fb3cc4c12aea032a1464ab08c6f3) argument can be removed and one can keep/set the value directly in ssl.conf, right?

TehPeGaSuS:
Feel free to open a new PR with your thoughts to be reviewed by the eggdrop team.

Cheers

PR #1312, opened 2022-07-26
Fix plain SASL auth when TLS is unavailable

eryg-kai:
Found by: @eryg-kai
Patch by: @eryg-kai
Fixes: n/a

One-line summary: 

Fix plain SASL auth when TLS is unavailable

There is common code after the if/else that was being omitted by an
ifdef that should only be affecting the last else branch.

This change brings this ifdef in line with the previous ifdef (utilizing
an early `return 1`).

I have built and ran this change without TLS to verify it works but 
please let me know if there is more I should do.

### DIFF
--- a/src/mod/server.mod/servmsg.c
+++ b/src/mod/server.mod/servmsg.c
@@ -1383,8 +1383,8 @@ static int tryauthenticate(char *from, char *msg)
       }
     }
 #else
-      putlog(LOG_DEBUG, "*", "SASL: TLS libs missing EC support, try PLAIN or "
-                "EXTERNAL method");
+      putlog(LOG_DEBUG, "*", "SASL: TLS libs not present or missing EC support."
+                " Try the PLAIN or EXTERNAL method instead");
       return 1;
     }
 #endif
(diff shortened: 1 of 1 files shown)


vanosg:
Hi @eryg-kai , thanks so much for not only finding this bug, but providing a PR for it! First (quick) glance, it looks good. We'll look at it a little more closely soon and determine if it is too late to slide this in to 1.9.3 before it goes stable or not. Feel free to join us on #eggdrop on Libera if you want to talk more about it, we love having contributors around to bounce ideas off of and provide feedback. Thanks again!

vanosg:
Testing:
Confirmed that a 1.9.2 bot using PLAIN will not finish the SASL authentication process (timeout) when compiled with --distable-tls. Applying this patch allowed the bot to authenticate with PLAIN both with and without --disable-tls.

vanosg:
And just to make it crystal clear why this is allowed, from RFC 4616 (PLAIN SASL):

"   The PLAIN SASL mechanism does not provide a security layer."

So there is no reason I can see that we would need to have TLS libs installed in order to use PLAIN SASL authentication.

vanosg:
@michaelortmann can I get a thumbs up from you that this looks good before we merge it?

PR #1330, opened 2022-09-12
Fix more format specifiers

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix format specifier for intptr_t with c99 PRIdPTR that were shadowed by the use of egg_snprintf() and simple_sprintf() / use snprintf() for all getudef()

This PR is also a continuation of fe00c1e285b9399c68f1192e3bbb1402cf790de7
This PR will also solve part of #1028


thommey:
lgtm

PR #1332, opened 2022-09-12
Task/finishaccountbind

thommey:
- Central setaccount() function to update account and call binds
- Normalized m->account values (empty string = unknown, "*" = not logged in, otherwise accountname)
- raw and rawt accept wildcards
- moved most of account functionality to irc.mod because it's only relevant in the context of channels
- improved .status reporting for account tracking
- request account-notify and extended-join by default

### DIFF
--- a/doc/sphinx_source/using/accounts.rst
+++ b/doc/sphinx_source/using/accounts.rst
@@ -1,0 +1,81 @@
+Account tracking in Eggdrop
+===========================
+
+In Eggdrop 1.9.3, Eggdrop added the ability to associate nicknames with the service accounts they are logged into. It is IMPORTANT to note that Eggdrop's ability to do this is dependent on an IRC server's implementation of three features- the IRCv3 extended-join capability, the IRCv3 account-notify capability, and WHOX support. All three of these features must be supported by the server and, in the case of extended-join and account-notify, requested by Eggdrop in order for Eggdrop to maintain "perfect" association between nicknames and account statuses.
+
+Required Server Capabilities
+----------------------------
+You're going to see this repeated a lot- the IRC server must support three features in order for Eggdrop to accurately associate accounts with nicknames. These three features allow Eggdrop to always know the current association between an account and a nickname by getting account statuses of users already on a channel when it joins, new users joining a channel, and users who authenticate while on a channel.
+
+extended-join
+^^^^^^^^^^^^^
+`extended-join <https://ircv3.net/specs/extensions/extended-join>`_ is an IRCv3-defined capability that adds the account name of a user to the JOIN message sent by the IRC server, alerting clients that a new member has joined a channel. Enabling this capability allows Eggdrop to immediately determine the account name associated with a user joining a channel
+
+account-notify
+^^^^^^^^^^^^^^
+`account-notify <https://ircv3.net/specs/extensions/account-notify>`_ is an IRCv3-defined capability that sends a message to a channel when a member of the channel either authenticates or deauthenticates from their account. Enabling this capability allows Eggdrop to immediately associate an account to a channel member when they authenticate or deauthenticate.
+
+WHOX
+^^^^
+'WHOX <https://ircv3.net/specs/extensions/whox>`_ is a server feature that allows a client to request custom fields to be returned in a WHO response. If a server supports this capability, Eggdrop sends a WHOX query to the server when it joins a channel, allowing it to immediately determine accounts associated with channel members when Eggdrop joins a channel.
+
+Enabling Eggdrop Account Tracking
+---------------------------------
+By default, the Eggdrop config file will attempt to enable all the capabilities required for account tracking. There are two settings to pay attention to
+::
+
+  # To request the account-notify feature via CAP, set this to 1
+  set account-notify 1
+
+  # To request the extended-join feature via CAP, set this to 1
+  set extended-join 1
+
+The ability of a server to support WHOX queries is determined via a server's ISUPPORT (005) reply. If a server supports WHOX queries, Eggdrop will automatically enable this feature.
+
+Checking Account-tracking Status
+--------------------------------
+While Eggdrop is running, join the partyline and type `.status`. If account-tracking is enabled (both the server supports and Eggdrop has requested), you'll see this line
+::
+
+  Loaded module information:
+    #eggdroptest        : (not on channel)
+    Channels: #eggdroptest (trying)
+    Account tracking: Enabled           <--- This line
+    Online as: BeerBot (BeerBot)
+
+Otherwise, the prompt will tell you which required capability is missing/not enabled
+::
+
+  Loaded module information:
+    #eggdroptest        :   2 members, enforcing "+tn" (greet)
+    Channels: #eggdroptest (need ops)
+    Account tracking: Best-effort (Missing capabilities: extended-join, see .status all for details)      <---- This line
+    Online as: Eggdrop (Eggdrop)
+
+Determining if a Server Supports Account Capabilities
+-----------------------------------------------------
+A server announces the capabilities it supports via a CAP request. If you have Tcl enabled on the partyline (or via a raw message from a client), you can send `.tcl cap ls` and see if the extended-join and account-notify capabilities are supported by the server. If they are not listed, the server does not support it.
+
+A server announces if it supports WHOX via its ISUPPORT (005) announcement. If you have Tcl enabled on the partyline, you can send `.tcl issupport isset WHOX` and if it returns '1', WHOX is supported by the server.
+
+Best-Effort Account Tracking
+----------------------------
+If a server only supports some, but not all, of the required capabilities, Eggdrop will switch to 'best effort' account tracking. This means Eggdrop will update account statuses whenever it sees account information, but in this mode Eggdrop cannnot guarantee that all account associations are up to date.
+
+If a server does not support extended-join, Eggdrop will not be able to determine the account associated with a user when they join. Eggdrop can update this information by sending a WHOX to the server.
+
+If a server does not support account-notify, Eggdrop will not be able to determine the account assoicated with a user if they authenticate/deauthenticate from their account after joining a channel. Eggdrop can update this information by sending a WHOX to the server.
+
+If a server does not support WHOX, Eggdrop will not be able to determine the accounts associated with users already on a channel before Eggdrop joined. There is no reliable way to update this information.
+
+One workaround to significantly increase the accuracy of account tracking for scripts in a 'best effort' scenario would be to issue a WHOX query (assuming the server supports it), wait for the reply from the server, and then query for the account information.
+
+account-tag
+^^^^^^^^^^^
+One supplementary capability that can assist a best-effort account tracking scenario is the IRCv3-defined `account-tag capability <https://ircv3.net/specs/extensions/account-tag>`_. The account-tag capability attaches a tag with the account name associated with the user sending a command. Enabling this capability allows Eggdrop to update its account tracking every time a user talks in channel, sets a mode, sends a kick, etc. While still not able to offer the same level of accuracy as enabling the "main three" account tracking features, it can increase the overall accuracy of the account list. Additionally, binds that react to user activity (pub, kick, mode, etc) containing account-tag will update the internal account list prior to executing the associated callback, so looking up the account name in the callback can be considered accurate.
+
+Using Accounts with Tcl Scripts
+-------------------------------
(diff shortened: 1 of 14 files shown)


src/mod/irc.mod/chan.c
```diff
@@ -212,6 +212,33 @@ static void update_idle(char *chname, char *nick)
   }
 }
 
+/* set user account on all members on all channels,
+ * trigger account bind if account state was not "unknown" (empty string)
+ */
+// TODO should be in irc.mod
```
vanosg: It is!

src/mod/irc.mod/chan.c
```diff
@@ -212,6 +212,33 @@ static void update_idle(char *chname, char *nick)
...
+
+  for (chan = chanset; chan; chan = chan->next) {
+    if ((m = ismember(chan, nick))) {
+      if (rfc_casecmp(m->account, account)) {
+        /* account was known */
+        if (m->account[0]) {
+          if (!strcmp(account, "*")) {
+            putlog(LOG_JOIN, chan->dname, "%s!%s has logged out of their account", nick, m->userhost);
+          } else {
+            putlog(LOG_JOIN, chan->dname, "%s!%s logged in to their account %s", nick, m->userhost, account);
```
vanosg: I can see LOG_JOIN makes sense particularly when a user is joining a channel with this information, but an idle user could be present in a channel for a week and then authenticate, so I would tend to view an account as a server-related activity moreso than a join activity. My first instinct was to go either DEBUG or SERVER... but to be fair, 'aways' are listed under join. Hmm hmm...

And separately, pure aesthetics, do we want to log the entire usermask? Or just the nick

src/mod/irc.mod/chan.c
```diff
@@ -1216,25 +1241,11 @@ static int got352or4(struct chanset_t *chan, char *user, char *host,
   /* Update accountname in channel records, 0 means logged out */
   /* A 0 is not a change from "" */
   if (account) {
```
vanosg, line 1242: This gets tricky.... this function is triggered for every WHOX, but the account field we're looking for depends on the WHOX format field being "c%%chnufat,222". If a Tcl script sends "%cuhnar,42" as the format field, Eggdrop does this:
```
[22:04:05] BeerBot!user@b0rg-C0E59E12.foo.net logged in to their account :realname
```
And .channel shows the account name as ":realname"

eggdrop.conf
```diff
@@ -1158,17 +1158,21 @@ server add ssl.example.net +7000
...
 
-# To request the message-tags feature via CAP, set this to 1. NOTE: Enabling
-# this feature may interfere with RAW binds in scripts.
-#set message-tags 0
+# To request the message-tags feature via CAP, set this to 1
+set message-tags 1
+
+# To request the account-tag feature via CAP, set this to 1
+# This should only be enabled if necessary for imperfect account tracking
+# see doc/ACCOUNT for details
```
vanosg: A user may have other reasons they want to enable this than account-tracking, we may want to soften this language

eggdrop.conf
```diff
@@ -1158,17 +1158,21 @@ server add ssl.example.net +7000
...
 # To request the account-notify feature via CAP, set this to 1
-#set account-notify 0
+set account-notify 1
 
 # To request the extended-join feature via CAP, set this to 1
-#set extended-join 0
+set extended-join 1
 
 # To request the invite-notify feature via CAP, set this to 1
 #set invite-notify 0
```
vanosg, line 1167: I was going to say I was wary of enabling this by default, but I talked myself out of it, this is a good idea. But that inner-monologue left me thinking we need to add a big warning above this section saying "just because you enable it here doesn't mean the server supports it" or some variation. I definitely see people thinking these are enabled simply because it is set here.

eggdrop.conf
```diff
@@ -1158,17 +1158,21 @@ server add ssl.example.net +7000
...
 
-# To request the message-tags feature via CAP, set this to 1. NOTE: Enabling
-# this feature may interfere with RAW binds in scripts.
-#set message-tags 0
+# To request the message-tags feature via CAP, set this to 1
```
vanosg, line 1169: Does this no really longer affect RAW binds after these changes? IIRC, a rawt bind will stop, and the raw bind table will not be checked?

Also, do we really want this enabled by default in a general sense? I suspect you did it for account-tag, but ... enh. Change.

vanosg (commented):
Awesome!

src/mod/irc.mod/irc.c
```diff
@@ -1045,13 +1061,59 @@ static void flush_modes()
...
+    } else if (!strcasecmp("account-notify", current->name) && current->enabled) {
+      notify = 1;
+    } else if (!strcasecmp("account-tag", current->name) && current->enabled) {
+      tag = 1;
+    }
+    current = current->next;
+  }
+
+  if (whox && notify && extjoin) {
+    dprintf(idx, "%s", "    Account tracking: Perfect\n");
```
vanosg: Can we go back to "enabled" or some other word here? Supported? Fully enabled? I know we've been using this word to describe the scenario, but I think the intended meaning without additional context would be lost on a novice user.

src/mod/irc.mod/irc.c
```diff
@@ -1045,13 +1061,59 @@ static void flush_modes()
...
+      if (!notify) {
+        dprintf(idx, "%s", "      - account-notify missing => Accounts will not update immediately when users log in or out\n");
+      }
+      if (!extjoin) {
+        dprintf(idx, "%s", "      - extended-join missing  => Accounts will not be known immediately when a user joins\n");
+      }
+      if (tag) {
+        dprintf(idx, "%s", "      - account-tag present    => Accounts will update whenever someone messages a channel or this bot\n");
+      }
+      dprintf(idx, "%s", "      See doc/ACCOUNT for more details");
```
vanosg: Need a \n

src/mod/irc.mod/irc.c
```diff
@@ -1045,13 +1061,59 @@ static void flush_modes()
...
+        dprintf(idx, "%s", "      - WHOX missing           => Accounts will not be known after Eggdrop joins a channel\n");
+      }
+      if (!notify) {
+        dprintf(idx, "%s", "      - account-notify missing => Accounts will not update immediately when users log in or out\n");
+      }
+      if (!extjoin) {
+        dprintf(idx, "%s", "      - extended-join missing  => Accounts will not be known immediately when a user joins\n");
+      }
+      if (tag) {
+        dprintf(idx, "%s", "      - account-tag present    => Accounts will update whenever someone messages a channel or this bot\n");
```
vanosg: I like this idea... maybe add the inverse comment as well, what *does* work? 
```
      if (!whox) {
        dprintf(idx, "%s", "      - WHOX missing           => Accounts will not be known after Eggdrop joins a channel\n");
      } else {
        dprintf(idx, "%s", "      - WHOX present           => Accounts will update when Eggdrop joins a channel\n");
      }
```
to give the user a full idea of what it does/doesn't do?

src/mod/irc.mod/chan.c
thommey: gotaway logs exactly the same way including the full hostmask, would be a discussion about that as well.
Channel related I believe we only have JOIN and MODES but those changes are more subtle than that, I'll leave this for now but make a mental note to discuss both before release (look at all LOG_JOIN uses and stuff)

src/mod/irc.mod/chan.c
thommey: we ignore the response in got354 if the query type isn't 222 so mismatch isn't possible unless scripts are malicious, the colon must be from gotaccount

eggdrop.conf
thommey: I think your warning on top of the CAP block in the configuration should be enough

eggdrop.conf
thommey: can you find out if we need it or if it works separately from account-tag? I disabled it for now and only set account-tag

src/mod/irc.mod/irc.c
thommey: perfect idea

thommey:
ready for review again I think

thommey:
docs need to be generated after merge

eggdrop.conf
vanosg: account-tag will insert a tag to a message, even if message-tags is not enabled.

src/mod/irc.mod/chan.c
vanosg: nuh-uh!

doc/sphinx_source/using/accounts.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Determining if a Server Supports Account Capabilities
+-----------------------------------------------------
+A server announces the capabilities it supports via a CAP request. If you have Tcl enabled on the partyline (or via a raw message from a client), you can send `.tcl cap ls` and see if the extended-join and account-notify capabilities are supported by the server. If they are not listed, the server does not support it.
+
+A server announces if it supports WHOX via its ISUPPORT (005) announcement. If you have Tcl enabled on the partyline, you can send `.tcl issupport isset WHOX` and if it returns '1', WHOX is supported by the server.
+
+Best-Effort Account Tracking
+----------------------------
+If a server only supports some, but not all, of the required capabilities, Eggdrop will switch to 'best effort' account tracking. This means Eggdrop will update account statuses whenever it sees account information, but Eggdrop cannnot guarantee the accuracy of account association at any given time.
```
thommey: "Eggdrop cannnot guarantee the accuracy of account association at any given time" I would say "Eggdrop cannot guarantee that the account association is up to date"

doc/sphinx_source/using/accounts.rst
```diff
@@ -0,0 +1,81 @@
...
+
+If a server does not support extended-join, Eggdrop will not be able to determine the account associated with a user when they join. Eggdrop can update this information by sending a WHOX to the server.
+
+If a server does not support account-notify, Eggdrop will not be able to determine the account assoicated with a user if they authenticate/deauthenticate from their account after joining a channel. Eggdrop can update this information by sending a WHOX to the server.
+
+If a server does not support WHOX, Eggdrop will not be able to determine the accounts associated with users already on a channel before Eggdrop joined. There is no reliable way to update this information.
+
+One workaround to significantly increase the accuracy of account tracking for scripts in a 'best effort' scenario would be to issue a WHOX query (assuming the server supports it), wait for the reply from the eserver, and then query for the account information.
```
thommey: typo "eserver" and I would mention an explicit code example to do a minutely /who on all channels:
```tcl
bind cron - * updatewho
proc updatewho {args} {
  foreach chan [channels] {
    if {[botonchan $chan]} {
      # The exact same WHO query eggdrop sends on-join
      putserv "WHO $chan c%chnufat,222"
    }
  }
}
```

doc/sphinx_source/using/accounts.rst
```diff
@@ -0,0 +1,81 @@
...
+
+account-tag
+^^^^^^^^^^^
+One supplementary capability that can assist a best-effort account tracking scenario is the IRCv3-defined `account-tag capability <https://ircv3.net/specs/extensions/account-tag>`_. The account-tag capability attaches a tag with the account name associated with the user sending a command. Enabling this capability allows Eggdrop to update its account tracking every time a user talks in channel, sets a mode, sends a kick, etc. While still not able to offer the same level of accuracy as enabling the "main three" account tracking features, it can increase the accuracy and can help with scripts that react to user commands/messages.
```
thommey: Maybe mention `[getaccount $nick $chan]` specifically working in bind callback procs

doc/sphinx_source/using/accounts.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Using Accounts with Tcl Scripts
+-------------------------------
+The Eggdrop Tcl ACCOUNT bind is triggered whenver an existing account record stored by Eggdrop is modified, such as a user de/authenticating to their account while in a channel, or information such as an account-tag being seen that updates an existing user. However, the ACCOUNT bind will NOT be triggered for the creation of a new user record, such as a user joining a channel. The bind is triggered for every channel the user is seen on- this means if a user is present with Eggdrop on four channels, the bind will be executed four times, each time with a different channel variable being passed to the associated Tcl procedure. Additionally, in a best-effort account tracking situation, Eggdrop will update the account associated with a user on all channels, not just the channel the event is seen on (and thus resulting in a bind trigger for each channel the user is on).
+
+In order to trigger Tcl script events to cover all instances where a user logs in, you need to pair an ACCOUNT bind with a JOIN bind. This will allow you to execute account-based events when a user joins as well as if they authenticate after joining.
```
thommey, line 81: ... "If you want to execute an action when Eggdrop first joins a channel and gets to know account names you can trigger on raw 315 (End of /WHO).
```tcl
bind rawt - 315 endofwho
proc endofwho {from key text tags} {
  set chan [lindex [split $text] 0]
  foreach nick [chanlist $chan] {
    if {[getaccount $nick $chan] != "*"} {
      putlog "$nick on $chan is authenticated as [getaccount $nick $chan]"
    } else {
      putlog "$nick on $chan is not authenticated with services"
    }
  }
  # make sure there is no implicit return 1 which would stop further processing
  return 0
}

doc/sphinx_source/using/accounts.rst
vanosg: Do you want this in docs, or can we put this on the wiki? In my head, code samples could/should go there, a section has already been started for this type of thing

doc/sphinx_source/using/accounts.rst
vanosg: same, doc or wiki?

PR #1333, opened 2022-09-13
Avoid freeaddrinfo(NULL) and enhance error logging

### LINKED ISSUE #1329: setlisten can call freeaddrinfo with NULL leading to crash on openbsd
getaddrinfo() can fail and it is unclear from the specification whether the addresslist is set to NULL then or if it still needs to be freed.
On OpenBSD it sets it to NULL and calling freeaddrinfo() with NULL is not allowed.
That code snippet also lacks the gai_strerror() reporting suggestion from the man page of getaddrinfo - it can and does fail.
Our tcl listen command does not support hostnames - maybe it should?

Found by: Release_

michaelortmann:
Found by: Release_
Patch by: michaelortmann
Fixes: #1329

One-line summary:
Avoid `freeaddrinfo(NULL)` and enhance error logging

The behavior of `freeadrinfo(NULL)` is left unspecified by RFC 3493
Some Operating Systems like FreeBSD handle this for compatibility / portability but others like OpenBSD do not (as reported in #1329), so let eggdrop avoid calling `freeadrinfo(NULL)`.
Log `gai_strerror()` for every single case of error of `getaddrinfo()` and `getnameinfo()`

```
$ uname -a
OpenBSD openbsd71.home.arpa 7.1 GENERIC.MP#3 amd64
```
Config file:
```
set listen-addr test.com
listen 3333 all
```
Before:
```
* Last context: tclhash.c/247 []
* Please REPORT this BUG!
* Check doc/BUG-REPORT on how to do so.
* Wrote DEBUG
* SEGMENT VIOLATION -- CRASHING!
Segmentation fault (core dumped) 
```
After:
```
tcldcc: setlisten(): getaddrinfo(): hostname 'test.com' not known
Tcl error in file 'BotA.conf':
Couldn't listen on port '3333' on the given address: Cannot assign requested address. Please check that the port is not already in use
    while executing
"listen 3333 all"
    (file "BotA.conf" line 1668)
* CONFIG FILE NOT LOADED (NOT FOUND, OR ERROR)
```

### DIFF
--- a/src/dns.c
+++ b/src/dns.c
@@ -555,7 +555,7 @@ void *thread_dns_ipbyhost(void *arg)
       freeaddrinfo(res0);
   }
   else if (error == EAI_NONAME)
-    debug1("dns: thread_dns_ipbyhost(): getaddrinfo(): hostname '%s' not known", dtn->host);
+    debug1("dns: thread_dns_ipbyhost(): getaddrinfo(): hostname %s not known", dtn->host);
   else
     debug1("dns: thread_dns_ipbyhost(): getaddrinfo(): error = %s", gai_strerror(error));
   pthread_mutex_lock(&dtn->mutex);
--- a/src/dns.c
+++ b/src/dns.c
@@ -550,8 +550,8 @@ void *thread_dns_ipbyhost(void *arg)
       }
     }
 #endif
-    if (res0) /* The behavior of freeadrinfo(NULL) is left unspecified by RFC
-               * 3493 */
+    if (res0) /* The behavior of freeadrinfo(NULL) is left unspecified by RFCs
+               * 2553 and 3493. Avoid to be compatible with all OSes. */
       freeaddrinfo(res0);
   }
   else if (i == EAI_NONAME)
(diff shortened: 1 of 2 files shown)


thommey:
lgtm

vanosg:
Putting aside the "should we allow hostnames" (probably yes, just not for this version)- can we change the error message to something more like "unknown IP address" instead of "unknown hostname"? I think an error like this would lead the user to believe there is something wrong with the hostname provided, instead of telling them hostnames are not allowed.

(that is, if I'm reading that error case for EAI_NONAME correctly, I can't fully tell from a quick glance if it is also for an unresolvable hostname)

michaelortmann:
EAI_NONAME is not exactly identical for unresolveable i suppose. but the man pages combined with apriori knowledge about our usage of the function getaddrinfo() we are fine logging like this. the alternative would be to not handle EAI_NONAME as any special case and print system default error for it, which is more generic but has less power to help the user for our use case.

vanosg:
```
.tcl listen foo.com 4444
Tcl error: invalid ip address
```

vanosg:
What do you think of that?

Issue #815, opened 2019-06-19
[1.8.4] matchattr is not backward compatible


wilkowy:
1.6.21:
```
.tcl matchattr "handle" "-|-" "#channel"
Tcl: 1
```

1.8.4:
```
.tcl matchattr "handle" "-|-" "#channel"
Tcl error: Unknown flag specified for matching
```

Bug or change?

The workaround for now is to check flags for "-" & "-|-" before.

vanosg:
Yep, definitely a bug, from what I can see. Yikes!

vanosg:
Hi @wilkowy, looking in to this a little more, I want to walk back my last comment of 'bug'. In 1.6.21, the code you wrote would use the - as a 'not' prefix to a flag, not a 'any' flag the way they are used in bind masks (I'm assuming, perhaps incorrectly, that the bind mask format is how you thought it would work?), and didn't provide error checking if no flags were specified after the -. Now, it does.

Is there a specific use case you have, where you anticipate code as you wrote out being used like this? As it was implemented in 1.6.21, I can't think of any scenario where something other than a '1' would be returned when passing - or -|-. Essentially you're not providing Eggdrop any flags to match against, this change brings it more in line and, correctly in my opinion, tells you that no flags were provided (or in the exact terms of the error message, an unknown flag).

But hey, I could easily be wrong here- let me know your thoughts and we'll work through it. I think at minimum I'll add an additional phrase to tcl-commands.doc making explicit the functionality of '-' before closing out this issue. Thanks!

Test examples from 1.6.21 always returning 1:
```
.+user test1
Added test1 (no host) with no password and no flags.
.whois test1       
HANDLE    PASS NOTES FLAGS           LAST
test1     no       0 -               never (nowhere)
.tcl matchattr test1 -
Tcl: 1
.tcl matchattr test1 -|-
Tcl: 1

.chattr test1 +f #eggheads
No global flags for test1.
Channel flags for test1 on #eggheads are now +f.
.whois test1
HANDLE    PASS NOTES FLAGS           LAST
test1     no       0 -               never (nowhere)
   #eggheads          f               never
.tcl matchattr test1 -
Tcl: 1
.tcl matchattr test1 -|- #eggheads
Tcl: 1
.tcl matchattr test1 |- #eggheads
Tcl: 1

.chattr test1 -f #eggheads
No global flags for test1.
No flags for test1 on #eggheads.
.chattr test1 +f          
Global flags for test1 are now +f.
[...]

PR #1334, opened 2022-09-18
Fix matchattr without flags to always match again, e.g. - or -|-

### LINKED ISSUE #1267: matchattr throws an error when testing -|-
A long time before, with eggdrop 1.6.21, it was possible to do `matchattr $nick -|- $chan` and it returned 1 when user was unknow or haven't any flag in eggdrop.
Actually (since 1.8.x), it throws an error:

> Tcl error: Unknown flag specified for matching

This also happen if you want to check if user has no global flag but a channel flag, or if user has global flag but no chan flag.

So, the only way to now check that is to do `$handle != "*" && [chattr $handle $chan]=="-"`
Old scripts must be modified, which could be a potential danger: some users don't update their eggdrops because scripts doesn't work anymore

thommey:
Fixes: #1267

The matchattr/bind flag string "-" or "-|-" is commonly used to match "anyone". It isn't a real match because there are no flags specified so there is nothing to match. - and | are just meta-characters that don't parse to modes that can match.

Found by: crazycatdevs
Patch by: thommey

### DIFF
--- a/src/tcluser.c
+++ b/src/tcluser.c
@@ -270,6 +270,7 @@ static int tcl_matchattr STDVAR
       if (!plus.global && !plus.udef_global && !plus.chan &&
           !plus.udef_chan && !plus.bot) {
         /* No flags (e.g. "-" or "+" or "-|-" matches anyone */
+        Tcl_AppendResult(irp, "1", NULL);
         return TCL_OK;
       }
     }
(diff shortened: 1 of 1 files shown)


wilkowy:
Found by: #815 :P

vanosg:
@wilkowy yeah... you too :) Thanks!

PR #1340, opened 2022-10-07
redo matching docs

vanosg:


vanosg:
@thommey : based on my tests, we have no way to change matching for global flags from 'or' to 'and'. I updated the doc examples around that to match for now. Please add more-better words to it

For everyone, please let me know if this revised documentation makes sense- the goal was to finally actually document how flag masks are used by Eggdrop.

vanosg:
And I think I figured out why people started using "-" as a 'skip' flag- looking at .binds, it presents an "everything" bind as -|-  . So I bet people just started copying that.

doc/sphinx_source/using/tcl-commands.rst
```diff
@@ -2922,10 +2884,55 @@ Removing a bind
...
+
+The easiest way to explain how to build a flag mask this is by demonstration. A flag mask of "v" by itself means "has a global v flag". To also check for a channel flag, you would use the flag mask "v|v". This checks if the user has a global "v" flag, OR a channel "v" flag (again, the | means "OR" and ties the two types of flags together). You could change this mask to be "v&v", which would check if the user has a global "v" flag AND a channel "v" flag. Lastly, to check if a user ONLY has a channel flag, you would use "\*|v" as a mask, which would not check global flags but does check if the user had a channel "v" flag.
```
skralg: 'this' is an extra word that needs removed

doc/sphinx_source/using/tcl-commands.rst
```diff
@@ -2922,10 +2884,55 @@ Removing a bind
...
+In the next section, you will see several references to "flags". The "flags" argument takes a flag mask, which is a value that represents the type of user that is allowed to trigger the procedure associated to that bind. The flags can be any of the standard Eggdrop flags (o, m, v, etc). Additionally, when used by itself, a "-" or "*" can be used to skip processing for a flag type. A flag mask has three sections to it- global, channel, and bot flag sections. Each section is separated by the | or & logical operators ( the | means "OR" and the & means "AND; if nothing proceeds the flag then Eggdrop assumes it to be an OR). Additionally, a '+' and '-' can be used in front of a flag to check if the user does (+) have it, or does not (-) have it.
```
thommey: `"flags" argument` does not make sense in this context, maybe `"flags" argument of [bind] or [matchattr]`?

doc/sphinx_source/using/tcl-commands.rst
```diff
@@ -2922,10 +2884,55 @@ Removing a bind
...
+
+The easiest way to explain how to build a flag mask is by demonstration. A flag mask of "v" by itself means "has a global v flag". To also check for a channel flag, you would use the flag mask "v|v". This checks if the user has a global "v" flag, OR a channel "v" flag (again, the | means "OR" and ties the two types of flags together). You could change this mask to be "v&v", which would check if the user has a global "v" flag AND a channel "v" flag. Lastly, to check if a user ONLY has a channel flag, you would use "\*|v" as a mask, which would not check global flags but does check if the user had a channel "v" flag.
+
+You will commonly see flag masks for global flags written "ov"; this is the same as "\|ov".
```
thommey: You could add the also very commonly used variant `-|ov` and/or your desired `*|ov`.
I would also change one of the examples in the table (with no global flags) to it because it is so common.

doc/sphinx_source/using/tcl-commands.rst
```diff
@@ -2922,10 +2884,55 @@ Removing a bind
...
+
+Some additional examples:
+
++------------+-----------------------------------------------------------------+
+| Flag Mask  | Action                                                          |
++============+=================================================================+
+| m, +m      | Checks if the user has the m global flag                        |
```
thommey: `m|-` is also a version of this or `+m|-` if you want to include one of those so that it's obvious when reading existing scripts what it means without having to go to the side note after the table.

src/flags.c
```diff
@@ -1133,7 +1133,7 @@ static int flag2str(char *string, int bot, int udef)
     x++;
   }
   if (string == old)
-    *string++ = '-';
```
thommey, line 1136: Can we postpone this one to next release? I have no idea what breaking effects this might have and flags are hairy.

PR #1349, opened 2022-11-15
typo fix : evnt -> event 

ZarTek-Creole:


vanosg:
The "evnt" bind is actually "evnt", so not a typo, just a funnily-spelled bind resulting from an earlier 4-char name length limit

ZarTek-Creole:
strang because : `wrong # args: should be "callevent event" ` say too  `event`

For example: callevent rehash, or callevent myownevent123


I don't tell you the reason for limitation...

PR #1350, opened 2022-11-15
Update eggdrop.conf with Predefined events

ZarTek-Creole:
Predefined events triggered by Eggdrop as an example

Found by: @[ZarTek-Creole](https://github.com/ZarTek-Creole)

Patch by: @[ZarTek-Creole](https://github.com/ZarTek-Creole)

Fixes: 

One-line summary:
Predefined events triggered by Eggdrop as an example


vanosg:
These are great examples, but probably belong in the wiki more than the config file. The existing example is provided for the common use case of logging into services upon join.

Issue #1352, opened 2022-11-16
Eggdrop crashes when launching into background on Mac OS


thommey:
Works with -n, crashes without -n on macOS Monterey with Intel CPU

Tcl8.6 non-threaded or threaded (does not affect it), issue in 1.8 and 1.9 (so it's not tdns)

Found by: MacinMan

thommey:
Quoting MacinMan:

```
--disable-corefoundation Disable use of CoreFoundation API and revert
                         to standard select based notifier, required
                         when using naked fork (i.e. not followed by
                         execve).
```
To Tcl when self-compiling it seems to fix the problem also also adding --enable-64bit because 10.15 (Catalina) drops 32 bit support, so, `./configure --disable-corefoundation --enable-64bit`

thommey:
We do "naked fork" so this is absolutely required, needs to be documented

PR #1115, opened 2021-02-03
Fix crash - Add an AlertNotifier callback function

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #515 and #1031

One-line summary:
Add an AlertNotifier callback function.

Additional description (if needed):
The line `putlog(LOG_MISC, "*", "stub tickle_AlertNotifier");` shouldn't be triggered. Test case is `cd == NULL`. The `putlog` is to show, if we have more work to do, if we have to replace that putlog stub. We could then borrow from https://github.com/tcltk/tcl/blob/645d300beada575f29f390a13f578f469a61b4fa/unix/tclUnixNotfy.c#L102. It's not that complicated. A simple `tsdPtr->eventReady = 1;` and some boilerplate would probably do. Having said that, the fix works for #515 and #1031 and if no problem / side effect is found, i propose we merge it in.

Test cases demonstrating functionality (if applicable):
**1. Test for #515:**
Before:
```
.tcl package require Thread
[20:07:46] tcl: builtin dcc call: *dcc:tcl -HQ 1 package require Thread
[20:07:46] tcl: evaluate (.tcl): package require Thread
Tcl: 2.8.6
.tcl thread::send -async [set tid [thread::create]] { after 10000; puts test }
[20:07:52] tcl: builtin dcc call: *dcc:tcl -HQ 1 thread::send -async [set tid [thread::create]] { after 10000; puts test }
[20:07:52] tcl: evaluate (.tcl): thread::send -async [set tid [thread::create]] { after 10000; puts test }
[20:07:52] * Last context: tclhash.c/734 [Tcl proc: *dcc:tcl, param:  $_dcc1 $_dcc2 $_dcc3]
[20:07:52] * Please REPORT this BUG!
[20:07:52] * Check doc/BUG-REPORT on how to do so.
[20:07:52] * Wrote DEBUG
[20:07:52] * SEGMENT VIOLATION -- CRASHING!
Segmentation fault (core dumped)
```
After:
```
.tcl package require Thread
[20:05:37] tcl: builtin dcc call: *dcc:tcl -HQ 1 package require Thread
[20:05:37] tcl: evaluate (.tcl): package require Thread
Tcl: 2.8.6
.tcl thread::send -async [set tid [thread::create]] { after 10000; puts test }
[20:05:46] tcl: builtin dcc call: *dcc:tcl -HQ 1 thread::send -async [set tid [thread::create]] { after 10000; puts test }
[...]

PR #1353, opened 2022-11-16
Fix crash - add tcl service mode hook stub / enhance doc

### LINKED ISSUE #1351: Replacement Tcl Notifier fails on Mac OS
`Tcl_ServiceModeHook: Notifier not initialized` on startup with both threaded and unthreaded Tcl 8.6.12
Disabling REPLACE_NOTIFIER is a workaround (./configure, make config, edit config.h to remove HAVE_NOTIFIER_INIT, make, make install).

macOS Monterey with Intel CPU

Found by: MacinMan

michaelortmann:
Found by: MacinMan
Patch by: michaelortmann
Fixes: #1351 and #1352

One-line summary:
Fix crash - add tcl service mode hook stub / enhance doc

This fix is similar to #1115

Test under macOS Big Sur 11.7.1.
With default tcl 8.5.9 (MacOSX sdk) under macOS eggdrop has got no problem:
`./configure --with-tcl=/System/Volumes/Data/Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/System/Library/Frameworks/Tcl.framework/tclConfig.sh`
With user compiled tcl 8.6.12:
`./configure --with-tcl=/Users/michael/opt/tcl8.6.12/lib/tclConfig.sh`
**Before:**
```
michael@macos eggdrop % ./eggdrop -t BotA.conf 

Eggdrop v1.9.4+alpha (C) 1997 Robey Pointer (C) 2010-2022 Eggheads
Tcl_ServiceModeHook: Notifier not initialized
zsh: abort      ./eggdrop -t BotA.conf
```
**After:**
No problem.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -19,6 +19,8 @@ Eggdrop uses the GNU autoconfigure scripts to make things easier.
         correctly compile Eggdrop. It will also try to find Tcl, which
         is required to compile.
 
+        For macOS tcl and openssl see doc/COMPILE-GUIDE.
+
 2.  
 
     Type either 'make config' or 'make iconfig' to determine which
@@ -154,5 +156,5 @@ the README file. If not, then READ IT!&@#%@!
 
 Have fun with Eggdrop!
 
-  Copyright (C) 1997 Robey Pointer Copyright (C) 1999 - 2022 Eggheads
-  Development Team
+  Copyright (C) 1997 Robey Pointer
+  Copyright (C) 1999 - 2022 Eggheads Development Team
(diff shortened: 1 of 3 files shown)


thommey:
LGTM

vanosg:
Does the log msg need to be misc? That looks more like it should be a debug message, no?

michaelortmann:
i modeled it after:
https://github.com/eggheads/eggdrop/blob/db30b022c53c3e5e278b21c09154713d21968416/src/tcl.c#L591-L595

so, we should change both log types, should we not?

vanosg:
Yeah... you added that line too :P 

Please change to debug and we should be good to go on this

thommey:
That message should never appear unless something's weird, we only set the service mode to SERVICE_ALL and nothing else should modify it, I would leave it at MISC so we can pay attention if it comes up. LGTM

PR #1360, opened 2022-11-29
Update autobotchk

TehPeGaSuS:
Let's enable the systemd unit so it actually restart the bot on crash/machine reboot.

Found by: PeGaSuS
Patch by: PeGaSuS
Fixes:

One-line summary:
Currently the user unit isn't enabled which will cause eggdrop to not restart on crash/machine reboot. This PR should fix it.


TehPeGaSuS:
I'll do a new PR with some new findings

PR #1364, opened 2022-12-14
Fix usage of language file for IRC_CHANGINGSERV

### LINKED ISSUE #1327: jump reason should be translated, perhaps
maybe the "changing servers" reason given to QUIT should be in the language file instead of a fixed string

Found by: primalz

michaelortmann:
Found by: primalz
Patch by: michaelortmann
Fixes: #1327

One-line summary:

`.jump` and `.tcl jump` still works


thommey (approved):
LGTM

PR #1370, opened 2022-12-19
Update bothost only on own chghost

vanosg:
Found by: Geo
Patch by: Geo

One-line summary:
Eggdrop incorrectly updated its own internally-tracked hostname with any host in a CHGHOST, whether it was intended for Eggdrop or not.

```
[02:35:21] [@] :geo!foo@foo.bar.mooz CHGHOST foo foo.bar.moo

(gdb) p botuserhost
$2 = "Eggdrop@b0rg-6B5420BE.aol.com", '\000' <repeats 72 times>
```


thommey:
LGTM on its own, but we should probably re-check bans and autoop/voice/... on a user host change. I assume we already do for nick change because that also adjusts the mask

PR #1372, opened 2022-12-20
Add CHGHOST bind

### LINKED ISSUE #1371: CHGHOST improvements
Add chghost bind; check ban/etc lists for matches after a chghost

vanosg:
Found by: thommey
Patch by: Geo
Fixes: #1371 

Questions for discussion:
- Should this trigger on chghost only, or chghost AND 396 (Undernet +x)?
- For the proc, should the last 'arg' value be returned in nick!ident@host or just ident@host ? ident and host are the only two values that will change AND nick is provided earlier, but I think the userbase is used to getting a full hostmask back

### DIFF
--- a/doc/sphinx_source/using/tcl-commands.rst
+++ b/doc/sphinx_source/using/tcl-commands.rst
@@ -3571,7 +3571,6 @@ The following is a list of bind types and how they work. Below each bind type is
 
   Description: triggered when a server sends either an IRCv3 spec CHGHOST message to change a user's hostmask. The mask is matched against "#channel nick!user\@host" and can contain wildcards. The triggered proc will return the nick of the user who's hostmask changed; the hostmask the affected user had before the change, the handle of the affected user (or * if no handle is present), the channel the user was on when the bind triggered, and the new hostmask of the affected user. This bind will trigger once for each channel the user is on.
 
-
 ^^^^^^^^^^^^^
 Return Values
 ^^^^^^^^^^^^^
--- a/doc/sphinx_source/using/tcl-commands.rst
+++ b/doc/sphinx_source/using/tcl-commands.rst
@@ -3432,6 +3432,7 @@ The following is a list of bind types and how they work. Below each bind type is
           init-server       - called when we actually get on our IRC server
           disconnect-server - called when we disconnect from our IRC server
           fail-server       - called when an IRC server fails to respond 
+          hidden-host       - called when the bot's host is hidden by the server
 
   Note that Tcl scripts can trigger arbitrary events, including ones that are not pre-defined or used by Eggdrop.
 
--- a/doc/sphinx_source/using/tcl-commands.rst
+++ b/doc/sphinx_source/using/tcl-commands.rst
@@ -3570,7 +3570,7 @@ The following is a list of bind types and how they work. Below each bind type is
 
   procname <nick> <old user@host> <handle> <channel> <new user@host>
 
-  Description: triggered when a server sends an IRCv3 spec CHGHOST message to change a user's hostmask. The new host is matched against mask in the form of "#channel nick!user\@host" and can contain wildcards. The specified proc will be called with the nick of the user who's hostmask changed; the hostmask the affected user had before the change, the handle of the affected user (or * if no handle is present), the channel the user was on when the bind triggered, and the new hostmask of the affected user. This bind will trigger once for each channel the user is on.
+  Description: triggered when a server sends an IRCv3 spec CHGHOST message to change a user's hostmask. The new host is matched against mask in the form of "#channel nick!user\@host" and can contain wildcards. The specified proc will be called with the nick of the user whose hostmask changed; the hostmask the affected user had before the change, the handle of the affected user (or * if no handle is present), the channel the user was on when the bind triggered, and the new hostmask of the affected user. This bind will trigger once for each channel the user is on.
 
 ^^^^^^^^^^^^^
 Return Values
--- a/doc/sphinx_source/using/tcl-commands.rst
+++ b/doc/sphinx_source/using/tcl-commands.rst
@@ -3569,7 +3569,7 @@ The following is a list of bind types and how they work. Below each bind type is
 
   procname <nick> <old user@host> <handle> <channel> <new user@host>
 
-  Description: triggered when a server sends either an IRCv3 spec CHGHOST message to change a user's hostmask. The mask is matched against "#channel nick!user\@host" and can contain wildcards. The triggered proc will return the nick of the user who's hostmask changed; the hostmask the affected user had before the change, the handle of the affected user (or * if no handle is present), the channel the user was on when the bind triggered, and the new hostmask of the affected user. This bind will trigger once for each channel the user is on.
+  Description: triggered when a server sends an IRCv3 spec CHGHOST message to change a user's hostmask. The mask is matched against "#channel nick!user\@host" and can contain wildcards. The triggered proc will return the nick of the user who's hostmask changed; the hostmask the affected user had before the change, the handle of the affected user (or * if no handle is present), the channel the user was on when the bind triggered, and the new hostmask of the affected user. This bind will trigger once for each channel the user is on.
 
 ^^^^^^^^^^^^^
 Return Values
--- a/doc/sphinx_source/using/tcl-commands.rst
+++ b/doc/sphinx_source/using/tcl-commands.rst
@@ -3432,7 +3432,7 @@ The following is a list of bind types and how they work. Below each bind type is
           init-server       - called when we actually get on our IRC server
           disconnect-server - called when we disconnect from our IRC server
           fail-server       - called when an IRC server fails to respond 
-          hidden-host       - called when the bot's host is hidden by the server
+          hidden-host       - called after the bot's host is hidden by the server
 
   Note that Tcl scripts can trigger arbitrary events, including ones that are not pre-defined or used by Eggdrop.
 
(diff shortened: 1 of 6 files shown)


TehPeGaSuS:
> Should this trigger on chghost only, or chghost AND 396 (Undernet +x)?

Definitely both since, after all, your host is changing.

> For the proc, should the last 'arg' value be returned in nick!ident@host or just ident@host ? ident and host are the only two values that will change AND nick is provided earlier, but I think the userbase is used to getting a full hostmask back

I'd go for the `n!u@h` style since then it's easier to match against it inside procs, IMHO.

vanosg:
I take that back; other binds return uhost as user@host, not nick!user@host

vanosg:
@thommey when you're around... this should *not* be stackable, right?

thommey:
Binds that take a mask must be stackable because multiple could match, not sure how that's handled in the code but non-stackable with a wildcard mask doesn't make sense

thommey:
Consistent would be: match against "#chan nick!oldident@oldhost nick!newident@newhost" and pass `nick olduser@oldhost hand chan newuser@newhost` I would say. The bind mask can be simplified though, it's probably not useful to match so thoroughly

vanosg:
A 396 or CHGHOST can be received when a bot first connects to the server (ie, the server enforces +x and changes hosts automatically) OR later in the session, if the bot were to authenticate with services. This leads to inconsistent behavior if the bot isn't on a channel when this happens because if the bot isn't on a channel, the bind won't ever be triggered. (CHGHOSTs for other users are only ever received if the bot is on a channel with them, 396s are only for the client, not other users). 

Would it make sense to make a 396/CHGHOST targeted at Eggdrop to be treated as a global/one-time event (differently than how we do it for another user)? Or maybe not include 396 in this bind at all, since it is only ever received for Eggdrop itself and never for another user (maybe leave this to be implemented as a rawt bind?).

thommey:
I am also leaning towards having +x being a special event because it can have significant meaning. Eggdrop could wait for its host to be hidden to join channels, for example (by setting them all to +inactive on init-server and reversing that on 396 for example).
Does it fit into bind evnt?

src/mod/irc.mod/chan.c
```diff
@@ -1335,6 +1335,59 @@ static int got354(char *from, char *msg)
...
+ */
+static int got396(char *from, char *msg)
+{
+  char *nick, *ident, userbuf[UHOSTLEN];
+
+  nick = newsplit(&msg);
+  if (match_my_nick(nick)) {  /* Double check this really is for me */
+    strlcpy(userbuf, botuserhost, sizeof userbuf);
+    ident = strtok(userbuf, "@");
+    egg_snprintf(botuserhost, sizeof botuserhost, "%s@%s", ident, msg);
```
michaelortmann: Can you change this to snprintf() pls? egg_snprintf() is not checked by the compiler, and should be phased out / changed into snprintf() everywhere.

src/mod/irc.mod/irc.c
```diff
@@ -776,6 +776,27 @@ static int invite_4char STDVAR
...
 
+static int check_tcl_chghost(char *nick, char *from, char *mask, struct userrec *u,
+                             char *chan, char *ident, char * host)
+{
+  struct flag_record fr = { FR_GLOBAL | FR_CHAN | FR_ANYWH, 0, 0, 0, 0, 0 };
+  char usermask[UHOSTMAX];
+  int x;
+
+  get_user_flagrec(u, &fr, NULL);
+  simple_sprintf(usermask, "%s!%s@%s", nick, ident, host);
```
michaelortmann: also pls. use snprintf() instead. if compiler throws warning, resize the array. i know its often a mess with UHOSTMAX stuff, but can we have snprintf() everywhere from now on?

vanosg:
Test:
With Tcl script:

.tcl bind chghost * * foo
.tcl proc foo {nick user hand chan arg} {putlog "nick user hand chan arg"}

```
[14:24:29] [@] :geo!case@foo.bar.moo CHGHOST case different.host.now
Tcl:
[14:24:29] geo geo!case@foo.bar.moo * #eggdroptest geo!case@different.host.now
[14:24:29] geo geo!case@foo.bar.moo * #test2 geo!case@different.host.now

.channel
@geo       handle   *          ---   m 10h         case@different.host.now

```
and for a self-chghost (396):
```
[15:07:13] [@] :irc.argon.b0rg.eu 396 BeerBot a.b.c.d :is now your displayed host

.status
<...>
Online as: BeerBot!LeafBlower@a.b.c.d (BeerBot)
<...>
```

vanosg:
hidden-host evnt bind tests:
```
.tcl bind evnt * hidden-host foo
Tcl: hidden-host
.tcl proc foo {type} {putlog "host changed! $type"}
Tcl: 

[15:56:28] [@] :irc.foo.com 396 BeerBot a.b.c.d :is now your displayed host
[15:56:28] host changed! hidden-host
```

vanosg:
Summary of feature:
- Adds CHGHOST bind to respond to CHGHOST message
- Updates internal user records with new hosts after CHGHOST
- Adds hidden-host type to EVNT bind
- Updates internal tracking of own host after a 396

doc/sphinx_source/using/tcl-commands.rst
```diff
@@ -3563,6 +3564,13 @@ The following is a list of bind types and how they work. Below each bind type is
 
   Module: irc
 
+(56) CHGHOST
+
+  bind chghost <flags> <mask> <proc>
+
+  procname <nick> <old user@host> <handle> <channel> <new user@host>
+
+  Description: triggered when a server sends an IRCv3 spec CHGHOST message to change a user's hostmask. The mask is matched against "#channel nick!user\@host" and can contain wildcards. The triggered proc will return the nick of the user who's hostmask changed; the hostmask the affected user had before the change, the handle of the affected user (or * if no handle is present), the channel the user was on when the bind triggered, and the new hostmask of the affected user. This bind will trigger once for each channel the user is on.
```
thommey: "proc will return" -> proc will be passed?
It is unclear from this paragraph whether the mask is matched against the old nick!user@host or the new one

src/mod/irc.mod/chan.c
```diff
@@ -1335,6 +1335,60 @@ static int got354(char *from, char *msg)
...
+  struct chanset_t *chan;
+  memberlist *m;
+  char mask[1024], *nick, *ident, buf[MSGMAX], *s1=buf, *chname;
+
+  strlcpy(s1, from, sizeof buf);
+  nick = splitnick(&s1);
+  ident = newsplit(&msg);  /* Get the ident */
+  /* Update my own internal hostmask */
+  if (match_my_nick(nick)) {
+    strlcpy(botuserhost, from, sizeof botuserhost);
```
thommey: wouldn't that be the old userhost? don't you have to construct the new one from ident and msg?

src/mod/irc.mod/chan.c
```diff
@@ -1335,6 +1335,60 @@ static int got354(char *from, char *msg)
...
+  u = get_user_by_host(from);
+  /* Run the bind for each channel the user is on */
+  for (chan = chanset; chan; chan = chan->next) {
+    chname = chan->dname;
+    m = ismember(chan, nick);
+    if (m) {
+      snprintf(m->userhost, sizeof m->userhost, "%s@%s", ident, msg);
+      snprintf(mask, sizeof mask, "%s %s", chname, from);
+      check_tcl_chghost(nick, from, mask, u, chname, ident, msg);
+      get_user_flagrec(m->user ? m->user : get_user_by_host(s1), &fr,
```
thommey, line 1366: why is the flag getting after calling the bind, shouldn't it be before so it filters?

src/mod/irc.mod/chan.c
```diff
@@ -1335,6 +1335,60 @@ static int got354(char *from, char *msg)
...
+ * :barjavel.freenode.net 396 BeerBot unaffiliated/geo/bot/beerbot :is now your hidden host (set by services.)
+ */
+static int got396(char *from, char *msg)
+{
+  char *nick, *ident, *uhost, userbuf[UHOSTLEN];
+
+  nick = newsplit(&msg);
+  if (match_my_nick(nick)) {  /* Double check this really is for me */
+    strlcpy(userbuf, botuserhost, sizeof userbuf);
+    ident = strtok(userbuf, "@");
```
thommey, line 1385: strtok can return NULL and we shouldn't crash then

src/mod/irc.mod/chan.c
```diff
@@ -1335,6 +1335,60 @@ static int got354(char *from, char *msg)
...
+    uhost = newsplit(&msg);
+    snprintf(botuserhost, UHOSTMAX, "%s@%s", ident, uhost);
```
thommey: sizeof botuserhost is better than UHOSTMAX

doc/sphinx_source/using/tcl-commands.rst
```diff
@@ -3432,6 +3432,7 @@ The following is a list of bind types and how they work. Below each bind type is
           init-server       - called when we actually get on our IRC server
           disconnect-server - called when we disconnect from our IRC server
           fail-server       - called when an IRC server fails to respond 
+          hidden-host       - called when the bot's host is hidden by the server
```
thommey: I think it would be worth documenting that $::botname is updated before the bind is called

src/mod/irc.mod/chan.c
vanosg: Pulling the flag record for the bind triggering is currently done inside check_tcl_chghost. But to your point, the flagrec is then pulled again here to be passed to check_this_member, which rechecks what status flags the user should receive on IRC, if any. I guess we could consolidate and call it before check_tcl_chghost, and then pass it to the function?

doc/sphinx_source/using/tcl-commands.rst
vanosg: Good catch- while the code current matches against the old user@host, I think this instead needs to match against the new user@host, similar to how a NICK bind matches on the new nickname

vanosg:
Updated test cases:

```
.tcl bind chghost n * owner
Tcl: *
.tcl proc owner {nick old hand chan new} {putlog "Owner $nick changed host to $new"}
Tcl: 
.tcl bind chghost * *!*@*.foo.com host
Tcl: *!*@*.foo.com
.tcl proc host {nick old hand chan new} {putlog "User $nick changed host to $new"}
Tcl: 

<User with no flags changes host to f.foo.com>
[14:44:41] User Geo changed host to Geo!scase@f.foo.com

<User with owner flags changes host to doo.com>
[14:45:51] Owner Bob changed host to Bob!scase@doo.com


.tcl putlog $botname
[14:46:48] BeerBot!bbot@grbgr-BE9345D7.aol.com
Tcl: 
<Bot changes hostmask to woo.com>
.tcl putlog $botname
[14:47:10] BeerBot!bbot@woo.com
Tcl: 

```

PR #1374, opened 2022-12-24
Add userhost-in-names capability

### LINKED ISSUE #1188: New features: userhost-in-names for IRCv3 capabilities
I am using an IRC server which uses [userhost-in-names](https://ircv3.net/specs/extensions/userhost-in-names) IRCv3 capabilities. Eggdropt does not yet support this.

**Description**
The userhost-in-names capability extends the NAMES reply messages (RPL_NAMREPLY) to contain the full hostmask (nick!user@host) of every user listed.

**Example:**

```
--> NAMES #atheme
:irc.bnc.im 353 guest = #atheme :Rylee!rylai@localhost somasonic!andrew@somasonic.org
:irc.bnc.im 366 guest #atheme :End of /NAMES list
```
Historical note: Some older (pre-IRCv3) software implements the same extension as PROTOCTL UHNAMES and advertises it using the UHNAMES token in RPL_ISUPPORT; some implementations support enabling the same capability using either CAP or PROTOCTL. New servers or clients SHOULD NOT implement the PROTOCTL method, although they may still do so for compatibility with older software.

For information, here is a table with the functions that are not yet implemented in eggdrop to compare to other irc bots :
[IRCv3 bots Client Support](https://ircv3.net/software/clients#bots)

### LINKED ISSUE #1368: Investigate/add userhost-in-names

vanosg:
Fixes:  #1368 
Fixes #1188 

One-line summary:
Add IRCv3 userhost-in-names capability

This updates channel records with hostmasks received from the userhost-in-names capability, and nothing more. Useful for cases where WHO queries are disabled.\


vanosg:
Current implementation questions:
* Should enabling userhost-in-names *not* send a WHO?
* Is it OK to have "competing" userhost requests? ie, a WHO and NAMES request toggling CHAN_PENDING flags; will this cause a desynch? (I don't think so)

thommey:
It should still send WHO because we are also interested in account info.
Where does CHAN_PENDING get set on a names request? the names response is sent without a request after a JOIN and afaik eggdrop never requests names

thommey:
Still need to add users, not just update hosts (and on updating hosts we might need to call chghost)

vanosg:
> It should still send WHO because we are also interested in account info. 

We don't get a WHO response at all (is my understanding)- so we won't get any account info. 

> Where does CHAN_PENDING get set on a names request? 

Eggdrop gets a JOIN for itself (gotjoin), and sends reset_chan_info. reset_chan_info sets the channel into a CHAN_PENDING status until reset_chan_info calls refresh_who_chan, which sends a WHO #channel to the server. Once the 315 is received, CHAN_PENDING is removed and CHAN_ACTIVE is set.

More specifically to the question, it's not ever set right now as a response to NAMES, but after the gotjoin().

> the names response is sent without a request after a JOIN and afaik eggdrop never requests names

From RFC, a 353 (names reply) is sent automatically after a JOIN. ( "If a JOIN is successful, the user is then sent the channel's topic (using RPL_TOPIC) and the list of users who are on the channel (using RPL_NAMREPLY), which must include the user joining") Without Eggdrop sending a NAMES, we're getting a 353 right now "for free".

vanosg:
Putting this here for probable later discussion:
* got352or4 adds the WHO_SYNCED flag. 
* got315 calls sync_members()
* sync_members clears users from the userlist that don't have the WHO_SYNCED flag set.

... because right now I'm debating running NAMES through got352or4 (hello got352or3or4!)

vanosg:
We're going to pause development on this feature until we can find a server that disables WHO. There are several implementation questions we have that can't be answered by our speculation, we need to see this "in the wild".

ZarTek-Creole:
I provided two IRC servers to @vanosg

src/mod/irc.mod/chan.c
```diff
@@ -1335,6 +1335,61 @@ static int got354(char *from, char *msg)
   return 0;
 }
 
+/* got 353: NAMES
+ * <server> 353 <chan> :[+/@]nick [+/@]nick ....
```
thommey: where is the "=/@" prefix in the example?

src/mod/irc.mod/chan.c
```diff
@@ -1335,6 +1335,61 @@ static int got354(char *from, char *msg)
...
+  if (find_capability("userhost-in-names")) {
+    newsplit(&msg);
+    newsplit(&msg); /* Get rid of =/@ prefix */
+    chname = newsplit(&msg);
+    nameptr = newsplit(&msg);
+    while ((uhost = newsplit(&nameptr))) {
+      if (!strcmp(uhost, "")) {
+        break;
+      }
+      fixcolon(uhost);
```
thommey, line 1279: I suggest moving fixcolon to before the loop as `fixcolon(nameptr);` - it only needs to remove the colon once at the beginning of the string if present ("x", ":x", ":x x x" are the options)

src/mod/irc.mod/chan.c
```diff
@@ -1335,6 +1335,61 @@ static int got354(char *from, char *msg)
...
+      nick = splitnick(&uhost);
+      /* Strip @, +, etc chars prefixed to nicks in NAMES */
```
thommey, line 1281: I would probably go with a more flexible removal routine, suspecting that "%+" can also happen with multi-prefix cap, right?

```c
char prefixchars[64];
snprintf(prefixchars, "+%%@%s", opchars);
```
and then
```c
nick += strcspn(nick, prefixchars);
```
to strip them all off in any order.

src/mod/irc.mod/chan.c
```diff
@@ -1335,6 +1335,61 @@ static int got354(char *from, char *msg)
...
+        nick=nick+1;
+      }
+      p = strchr(uhost, '@');
+      if (p) {
+        *p = 0;
+        host = p+1;
+      }
+      chan = findchan(chname);      /* See if I'm on channel */
+      if (chan && uhost && host) {
+        got352or4(chan, uhost, host, nick, "", NULL);
```
thommey, line 1298: a comment above to explain that this is a fake WHO response would be nice

vanosg: added

src/mod/irc.mod/chan.c
vanosg: grabbed prefix chars from 005 and used those

src/mod/irc.mod/chan.c
vanosg: done

src/mod/irc.mod/chan.c
vanosg: modified example

PR #1387, opened 2023-01-18
Add ssl header version to .status

### LINKED ISSUE #1322: OpenSSL header/library mismatch possible
With Tcl, we output both header version and library version. We should do the same with OpenSSL because a version mismatch between the found header and the found library is possible.

https://www.openssl.org/docs/manmaster/man3/OPENSSL_VERSION_NUMBER.html

In case of a mismatch we could warn the user to provide --with-sslinc and --with-ssllib to fix this (and set LD_LIBRARY_PATH or we set -Wl,-rpath with the new rpath semantic that makes it actually set runpath ...)

Found by: _wheel_

michaelortmann:
Found by: wheel
Patch by: michaelortmann
Fixes: #1322

One-line summary:
Add ssl header version to .status

This is not exactly, what #1322 asked for, so i would like to leave #1322 open.
But its a useful addition and could help debugging.

```
$ LD_LIBRARY_PATH=/home/michael/opt/openssl-3.1.0-beta1/lib64 ./eggdrop -t BotA.conf
[...]
.status
[...]
TLS library: OpenSSL 3.1.0-beta1 21 Dec 2022 (header version OpenSSL 3.0.7 1 Nov 2022)
```

### DIFF
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -336,22 +336,22 @@ void tell_verbose_status(int idx)
   dprintf(idx, "%s %s (%s %s)\n", MISC_TCLVERSION,
           ((interp) && (Tcl_Eval(interp, "info patchlevel") == TCL_OK)) ?
           tcl_resultstring() : (Tcl_Eval(interp, "info tclversion") == TCL_OK) ?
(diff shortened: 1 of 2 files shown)


vanosg:
@thommey re-open #1322 if you want?

PR #1395, opened 2023-02-02
Cleanup: unused but set variable

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Cleanup: unused but set variable

No functional change

Still compiles and runs


vanosg (approved):
Interesting, it's not even that it was used and then removed, but committed like that. Nice catch

PR #1398, opened 2023-02-06
Autoscript

vanosg:
One-line summary: Add autoscripts feature

The autoscripts system will allow Eggdrop to download, install, and configure scripts via the partyline. Since it's inception, users have needed to load a Tcl script into Eggdrop by downloading a Tcl file, editing the file to customize settings, and then sourcing that file in the config file. The autoscripts system helps by:

-   Centralizing commonly-used scripts in a single location
-   Allowing scripts to be downloaded via the partyline
-   Allowing script settings to be configured via the partyline
-   Allowing user-written scripts to be managed by the autoscripts system
-   Providing a documented API to write autoscripts-compatible scripts

To take advantage of this, scripts will have to be modified to adhere to a published API and be accompanied by a manifest.json file. Eggheads will host commonly used scripts so that users can download known "good" scripts. Additionally, the API will allow scripts to  be packaged and distributed outside of being downloaded from eggheads.

And obviously the traditional 'source' method of loading scripts is not going away!

### DIFF
--- a/doc/html/using/autoscripts.html
+++ b/doc/html/using/autoscripts.html
@@ -1,0 +1,341 @@
+
+<!DOCTYPE html>
+
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" /><meta name="generator" content="Docutils 0.19: https://docutils.sourceforge.io/" />
+
+    <title>Eggdrop Autoscripts &#8212; Eggdrop 1.9.4 documentation</title>
+    <link rel="stylesheet" type="text/css" href="../_static/pygments.css" />
+    <link rel="stylesheet" type="text/css" href="../_static/eggdrop.css" />
+    <script data-url_root="../" id="documentation_options" src="../_static/documentation_options.js"></script>
+    <script src="../_static/doctools.js"></script>
+    <script src="../_static/sphinx_highlight.js"></script>
+    <link rel="search" title="Search" href="../search.html" />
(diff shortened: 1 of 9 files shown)


ZarTek-Creole:
First of all thank you for this realization, very good work. I'm glad the idea was taken for a scripting system that we had talked about with crazycat.

I see that in the current state, several points:
- ".egg list" does not work. "Missing or unknown subcommand"
- The name autoscript, I don't find it great as a name for a script manager
- Having an .egg and .scripts alias isn't a good idea. A single command should avoid future confusion.
- It would be wise to use namespaces for the code. to avoid conflict with other scripts
- It's good to centralize scripts from one place, but it would be better if the user could add additional sources
- Maybe it would also be better to do ".egg fetch author/script" I imagine that has a conflict of interest if we put only one "youtube" script out of the 20 that exist of which there are easily 2-3 often use.
- Wouldn't it take the bot's master rights to be able to install scripts?
- When we run ".egg" without arguments, don't display "Missing or unknown subcommand"
- Have the possibility to choose the version of the scripts. For various reasons it may be useful to use a specific version
- Have an info command ".egg info <script>" that gives information, like the web address of the project. Versions, bug reports, tags, etc.
- an ".egg search <word>" to search for scripts
- Tags for scripts. to search, classify. For example "youtube" "http" "music" .. for youtube script

If you want me to work on code snippets or points, let me know.

If it can help here is my debut work from 8 months ago in the same idea
https://github.com/ZarTek-Creole/TCL-SCRIPT-MANAGER

vanosg:
Add base eggdrop version update check?

Issue #809, opened 2019-03-22
eof on ident should not terminate dcc connection


thommey:
as far as I know, a timeout on the ident lookup for an incoming telnet connection does not result in it being terminated. eof_dcc_identwait() does however terminate the underlying incoming telnet connection.

Found by: alizilla

PR #1406, opened 2023-02-09
Fix more ident lookup bugs

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix more ident lookup bugs

If a socket is set `O_NONBLOCK` and `connect()` returns `EINPROGRESS`, it must be checked for `ECONNREFUSED`.
Here, this was done properly:
https://github.com/eggheads/eggdrop/blob/2a6a36888f5aa2204d84a9e6282d35e5421c2c8a/src/net.c#L560-L580
But here it was not and a write was committed:
https://github.com/eggheads/eggdrop/blob/2a6a36888f5aa2204d84a9e6282d35e5421c2c8a/src/dcc.c#L1383-L1404

Additionally `O_NONBLOCK` is POSIX 2001 and old quirk code is removed, see: https://www.gaertner.de/~neitzel/susv3/basedefs/fcntl.h.html

Additionally also the proper code mentioned above had an issue, it did `select()` with an timeout of 1 sec, when it should be nonblocking. to have it fully nonblocking, the main loop `select()` would have to handle the econnrefused. to keep this PR small, this PR lowers the timeout to 0.5 sec, to mitigate long blocking but still detect econnrefused.

Additionally there was a bug here:
https://github.com/eggheads/eggdrop/blob/2a6a36888f5aa2204d84a9e6282d35e5421c2c8a/src/dcc.c#L1388-L1403
Were this line `j = 0` ever executed, the gates of tartaros would have opened for `dcc[0]`.

**Without this fix:**
When telnetting to the bot and the user has no identd running, the bot would log like:
```
[21:01:18] Telnet connection: localhost.home.arpa/45052
[21:01:18] net: eof!(write) socket 13 (Broken pipe,32)
```
**With this fix:**
Test without identd:
```
[12:24:02] Telnet connection: localhost.home.arpa/52216
[12:24:02] net: attempted socket connection refused: 127.0.0.1:113
[12:24:02] Ident failed for localhost.home.arpa: Connection refused
[12:24:09] pbkdf2 method SHA256 rounds 16000, user 12.493ms sys 0.416ms
[12:24:09] Logged in: testuser (telnet@localhost.home.arpa/52216)
*** testuser joined the party line.
```
Test wih identd:
```
[12:24:25] Telnet connection: localhost.home.arpa/42944
[12:24:28] pbkdf2 method SHA256 rounds 16000, user 16.223ms sys 6.315ms
[12:24:28] Logged in: testuser (michael@localhost.home.arpa/42944)
*** testuser joined the party line.
```
Additionally i applied this patch to windrop and it also works under cygwin.


michaelortmann:
Maybe related/fixing #809?

vanosg:
Please resolve conflicts, thanks

michaelortmann:
> Please resolve conflicts, thanks

done. ready for review again.

src/net.c
```diff
@@ -533,6 +533,54 @@ static int get_port_from_addr(const sockname_t *addr)
...
+
+  rc = connect(s, &addr->addr.sa, addr->addrlen);
+  /* To minimize a proven race condition, call ident here (especially when
+   * rc < 0 and errno == EINPROGRESS)
+   */
+  if (check_tcl_event_ident) {
+    /* push/pop errno, better safe than sorry */
+    e = errno;
+    check_tcl_event("ident");
+    errno = e;
```
thommey, line 550: use e later then for strerror? I am not sure setting errno is legal

src/net.c
```diff
@@ -533,6 +533,54 @@ static int get_port_from_addr(const sockname_t *addr)
...
+  }
+  if (rc < 0) {
+    if (errno == EINPROGRESS) {
+      /* Async connection... don't return socket descriptor
+       * until after we confirm if it was successful or not */
+      tv.tv_sec = 0; /* dont block */
+      tv.tv_usec = 0;
+      FD_ZERO(&sockset);
+      FD_SET(s, &sockset);
+      select(s + 1, NULL, &sockset, NULL, &tv);
```
thommey, line 561: are you assuming select with TV=0 cannot be interrupted?otherwise I am missing at least EINTR handling - I think other select errors (invalid FD, ...) should also be handled, just in case

src/net.c
```diff
@@ -533,6 +533,54 @@ static int get_port_from_addr(const sockname_t *addr)
 #endif
 }
 
+/* Check for O_NONBLOCK connect() EINPROGRESS could be ECONNREFUSED */
+int connect_nonblock(int s, sockname_t *addr, int check_tcl_event_ident) {
```
thommey, line 536: where does the async/nonblock come from? Already set on the socket elsewhere? Could document that requirement as comment to make it more obvious

src/net.c
michaelortmann: errno is defined by the ISO C standard to be a modifiable lvalue of type int
https://man7.org/linux/man-pages/man3/errno.3.html

src/net.c
michaelortmann: i added a src comment for connect_nonblock()

src/net.c
michaelortmann: this code block was copied from https://github.com/eggheads/eggdrop/blob/develop/src/net.c#L584-L603
and i assumed this is working in current "production" code
now i can see your concern for not checking the return value of select and EINTR
but this is another issue (is it)?
if you are still concerned about this piece of code, please open a new separate issue about our "production" code.

thommey:
Overall very good change, but this does change the behavior of open_telnet_raw which previously had a select timeout of 1s which would detect ECONNREFUSED within that timeframe, now the timeout is 0 (return instantly) which never detects that and always returns a valid socket (unless it's localhost I guess).
That makes every invocation rely on the mainloop properly detecting and handling that, not sure if that's a regression risk or not, any opinion?

michaelortmann:
eggdrop did select() with a timeout of 1 sec, when it should be nonblocking. to have it fully nonblocking, the main loop select() would have to handle the econnrefused. to keep this PR small, this PR lowers the timeout to 0.5 sec, to mitigate long blocking but still detect econnrefused.

PR #1441, opened 2023-03-14
Fix putlog timestamp: localtime() -> localtime_r()

michaelortmann:
Found by: Arno
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix putlog timestamp: localtime() -> localtime_r()

No, eggdrop didnt invent the time machine, it just suffered from thread-unsafe localtime():
[21:55:36] Writing user file...
[21:55:35] Writing channel file...
This bug was introduced in 2019 by my very self:
5b395f2b7cd572afb0e58ee221e3ad86c9b94881

Also this patch replaces ctime() with ctime_r()


src/main.c
```diff
@@ -599,10 +599,11 @@ static void core_secondly()
       check_botnet_pings();
 
       if (!miltime) {           /* At midnight */
-        char s[25];
+        char s[26];
         int j;
 
-        strlcpy(s, ctime(&now), sizeof s);
+        ctime_r(&now, s);
+        s[24] = 0;
```
thommey, line 606: why is it an array of 26 if you cut it off at index 24?

thommey (commented):
do we need a feature test for ctime_r in autoconf or are we expecting thread-safe functions from posix 2001 to exist already?

src/main.c
michaelortmann: because ctime_r will write 26 bytes. if buffer smaller, then bufferoverflow.

see also:

https://manpages.debian.org/bookworm/manpages-dev/ctime_r.3.en.html

char *ctime_r(const time_t *restrict timep,
                    char buf[restrict 26]);

michaelortmann:
> do we need a feature test for ctime_r in autoconf or are we expecting thread-safe functions from posix 2001 to exist already?

thanks for asking!

`ctime_r()` is posix 2001. well. mostly:

we never (or did we) define, how to handle posix 2001 options.

if we say we depend on posix 2001, do we mean, we depend on posix 2001 options also?

i am open to whatever you decide. but a decision has to be made, on which options (or all) eggy should be able to depend on.

`ctime_r()` for example falls under the `TSF` (Thread-Safe Functions) extension.

should we decide to not depend on posix 2001 TSF option (or any option) then we have to implement check and fallback for ALL such functions, especially all `_r()` functions that crept in over time.

should we decide to depend on this, then we dont need to configure check for those.

i would vote to explicitly state which posix 2001 option(s) eggdrop depends on. keeping the list as short as possible. 

not allowing to depend on ANY option sounds nice. but then we have to make sure we have fallbacks for any `_r()` func and probably tdns? my gut says threading without `_r()` funcs doesnt make sense, but i am not sure.

thommey:
Agreed, I think we can rely on TSF.

thommey (approved):

src/main.c
michaelortmann: truncating it with `s[24] = 0` is to strip the trailing `\n`

PR #1459, opened 2023-07-11
Modifications for autoscripts

crazycatdevs:
- scripts variables are now in a <script> namespace rather than global
- modified udefflag to allow several flags
- added udefstr and udefint


crazycatdevs:
Proposal:
pehaps that udef variables can be more globaly defined in manifest.json. I didn't do it for retro-compatibility but something like:
```
{
   "config": {
      "udef": {
         "<varname>": {
            "type": "<flag/int/str",
            "description": "description of the variable",
            "value": (only if needed)
         }
      }
   }
}
```
And then the egg_config proc could be simplified with only one loop on udef and a switch on type.

vanosg:
Working through namespaces, we couldn't find a 'good' way to do this without requiring more metadata or other complex Tcl-foo. The intent/hope was this would be a lightweight solution as we're not prepared to maintain such a heavy lift the extra complexity may bring. Not to say we won't/can't go that way in the future, but I think this is the extent to which we're able to commit to at this time

This issue being, we want it to be easy to bring in scripts, but a script that already has a namespace makes it difficult to handle compared to one without, it would require us to heavily modify one or the other based on the namespace. This approach takes the 'dumb pipe' approach and lets the script handle it.

vanosg:
So at this time, we're taking the great udef suggestion, and updating the way we handle global namespace issues rather than placing every script in its own namespace

PR #1461, opened 2023-07-25
inet_aton() -> inet_pton()

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
inet_aton() -> inet_pton()

inet_pton() is POSIX 2001 and additionally supports IPV6
other egg_inet_aton() / inet_aton() calls within eggdrop were already replaced with inet_pton
This PR fixes the rest / module API


src/modules.c
```diff
@@ -526,7 +526,7 @@ Function global_table[] = {
   (Function) sock_has_data,
   (Function) bots_in_subtree,
   (Function) users_in_subtree,
-  (Function) egg_inet_aton,
+  (Function) 0,                   /* was egg_inet_aton -- use inet_pton() instead */
```
thommey, line 518: can't we just export inet_aton here then to avoid breaking modules?

michaelortmann: inet_aton is not posix. inet_aton(y,z) could be mapped to inet_pton(AF_INET,y,z), but thats not equivalent. inet_aton accepts numbers-and-dots notation, something inet_pton doesnt, so it could potentially break modules anyway. given the module api breakage, i propose we milestone this PR for eggdrop 1.10 pr newer.

michaelortmann: if we can risk theoretical breakage, we could of course export inet_aton if available, and export 0 otherwise. if that is what you had in mind.

michaelortmann: how to proceed?

PR #1474, opened 2023-10-03
Cleanup dead increment

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Cleanup dead increment

Value stored to 'len' is never read
Also fix format  specifier

No functional change, tested server connect still works

### DIFF
--- a/src/mod/server.mod/servmsg.c
+++ b/src/mod/server.mod/servmsg.c
@@ -2199,20 +2199,20 @@ static void connect_server(void)
 
 #ifdef IPV6
     if (inet_pton(AF_INET6, botserver, buf)) {
-      len += egg_snprintf(s, sizeof s, "%s [%s]", IRC_SERVERTRY, botserver);
+      len += snprintf(s, sizeof s, "%s [%s]", IRC_SERVERTRY, botserver);
     } else {
 #endif
-     len += egg_snprintf(s, sizeof s, "%s %s", IRC_SERVERTRY, botserver);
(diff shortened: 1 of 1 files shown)


thommey:
I know where this is coming from and I hate fixing this.

The idea is to enforce a memory safe pattern:
```c
size_t written = 0;
char buf[2048];

written += snprintf(buf + written, sizeof buf - written, "%d", test); // <----- (1)
written += snprintf(buf + written, sizeof buf - written, "%s%d", test2, test3);
written += snprintf(buf + written, sizeof buf - written, "%d%s", test4, test5);
written += snprintf(buf + written, sizeof buf - written, "%d", test6);
written += snprintf(buf + written, sizeof buf - written, "%f", test7); // <----- (2)
```

Technically, (1) can be simplified to `written = snprintf(buf, sizeof buf, "%d", test)` and (2) can be simplified to `snprintf(buf + written, sizeof buf - written, "%f", test7);`.

cppcheck really hates the dead store for the last line and this is where your fix is coming from.
I don't like either simplification because it reduces maintainability. If you add stuff later, you HAVE TO REMEMBER to add "written +=" to the last line instead of just copy & pasting the memory safe pattern. This has bitten people a lot and it's hard enough to write safe C code.

If this bugs static analyzers please add overrides/suppressions as this is a limitation of the analyzer throwing a false positive in your face when you have a perfectly good reason to do this - readability and maintainability are a lot more important than "dead stores" that the compiler throws away anyway.

PR #1475, opened 2023-10-04
Enhance error msg and remove superfluous semicolon

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Enhance error msg and remove superfluous semicolon

`.tcl listen 1.2.3.4 2345 users`
Before:
```
.tcl listen 1.2.3.4 2345 users
Tcl error: Couldn't listen on port '2345' on the given address: Cannot assign requested address. Please check that the port is not already in use
```
After:
```
Tcl error: Couldn't listen on port 2345 on the given address '1.2.3.4': Cannot assign requested address. Please check that the port is not already in use
```

### DIFF
--- a/src/tcldcc.c
+++ b/src/tcldcc.c
@@ -1176,15 +1176,22 @@ static int setlisten(Tcl_Interp *irp, char *ip, char *portp, char *type, char *m
     if (strlen(newip)) {
       setsockname(&name, newip, port, 1);
       i = open_address_listen(&name);
+      if (i < 0) {
+        snprintf(msg, sizeof msg, "Couldn't listen on port %d on the given "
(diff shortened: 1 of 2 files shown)


src/tcldcc.c
```diff
@@ -1176,15 +1176,22 @@ static int setlisten(Tcl_Interp *irp, char *ip, char *portp, char *type, char *m
     if (strlen(newip)) {
```
vanosg, line 1176: Looking at line 1039, I actually think this branch logic can be removed, as we always set newip now? If you can double-check me on that, I'll clean this section up appropriately.

thommey (approved):

PR #1477, opened 2023-10-06
remove quick-logs, line buffer instead

### LINKED ISSUE #1449: Remove quick-logs, write to logfiles instantly
quick-logs might have had a purpose in the past to protect disks against frequent writes, but I don't think it's relevant anymore and it would be much more beneficial to have logs be live. At least in my opinion.
This proposes to eliminate the quick-logs setting entirely and always instantly write to the logfiles that are held open anyway. It should be made sure they're line-buffered.

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: #1449

One-line summary:
remove quick-logs, line buffer instead

remove quick-logs
quick logs disabled meant file logfile was buffered with `flush()` every 5 minues
quick logs enabled meant logfile was buffered  `flush()` every minute
now, instead, we dont explicitely flush, but set logfile to line buffered
`setvbuf(_IOLBF)` is posix 2001
**Please regenerate docs via sphinx after merge**

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -88,7 +88,7 @@
 #endif
 
 extern char origbotname[], botnetnick[]; 
-extern int dcc_total, conmask, cache_hit, cache_miss, max_logs, quick_logs,
+extern int dcc_total, conmask, cache_hit, cache_miss, max_logs, log_interval,
            quiet_save;
 extern struct dcc_t *dcc;
 extern struct userrec *userlist;
@@ -141,6 +141,7 @@ char ver[41];        /* Version info (short form) */
 volatile sig_atomic_t do_restart = 0; /* .restart has been called, restart ASAP */
 int resolve_timeout = RES_TIMEOUT;    /* Hostname/address lookup timeout        */
 char quit_msg[1024];                  /* Quit message                           */
+int log_elapsed = 1;
 
 /* Moved here for n flag warning, put back in do_arg if removed */
 unsigned char cliflags = 0;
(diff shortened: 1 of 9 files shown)


src/main.c
```diff
@@ -662,10 +659,7 @@ static void core_secondly()
     if (((int) (nowtm.tm_min / 5) * 5) == (nowtm.tm_min)) {     /* 5 min */
       call_hook(HOOK_5MINUTELY);
       check_botnet_pings();
-      if (!quick_logs) {
-        flushlogs();
-        check_logsize();
-      }
+
```
thommey, line 662: why is check_logsize not necessary here but left in core_minutely()?

src/misc.c
```diff
@@ -580,8 +579,10 @@ void putlog (int type, char *chname, const char *format, ...)
           if (keep_all_logs) {
             snprintf(path, sizeof path, "%s%s", logs[i].filename, ct);
             logs[i].f = fopen(path, "a");
+            setvbuf(logs[i].f, NULL, _IOLBF, 0); /* line buffered */
```
thommey: missing NULL pointer check, opening could fail

thommey (changes requested):

src/misc.c
michaelortmann: Added NULL pointer check

src/main.c
michaelortmann: `check_logsize()` is no rocket science, so its no called for every log to make it absolutely precise, instead it is called minutely.

and thats enough imo.

if it is already called minutely, i see no reason, to additionally call it every 5 minutes, making it kinda called twice every 5 minutes.

could there be some log bloat hooked for the 5 minute hook? yes, but it will then happen anyway. and it will be fixed in next minute anyway.

the difference in calling or not calling it additionally every 5 minutes is so small, that i took the freedom to remove it here.

thommey (approved):
LGTM

PR #1480, opened 2023-10-08
Update autoconf to 2.71

vanosg:
Found by: Geo
Patch by: Geo, thommey


michaelortmann:
Good modernization, but do we really require autoconf version 2.71?
Shouldnt  AC_PREREQ([x.yz]) be set to the **minimum** required version.

thommey:
misc/runautotools now checks for the exact version, bumping the minimum isn't enough or we generate pointless changes back and forth if we use different versions.

A notable change is that we now use the compiler in its default mode (C11, C99, whichever newest it supports), because the macro to generate CFLAGS and force C99 is deprecated.

PR #1488, opened 2023-10-14
Fix logging of QUIT msg

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Log QUIT msg like we already do for any other msg sent to the server.

Messages from eggdrop to irc server are logged when raw-log is set. depending on the queue the message was put in, it gets logged like for example `[m->]` or `[s->]` The Quit message was sent to server directly without using those queues and thus was not logged.

in eggdrop.conf `set raw-log 1`
```
.die
[23:29:24] tcl: builtin dcc call: *dcc:die -HQ 1 
[23:29:24] #-HQ# die 
[23:29:24] [->] QUIT :-HQ
```

### DIFF
--- a/src/mod/server.mod/server.c
+++ b/src/mod/server.mod/server.c
@@ -2094,9 +2094,13 @@ static void server_postrehash()
 
 static void server_die()
 {
+  char msg[MSGMAX];
   cycle_time = 100;
   if (server_online) {
-    dprintf(-serv, "QUIT :%s\n", quit_msg[0] ? quit_msg : "");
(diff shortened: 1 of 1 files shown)


thommey (changes requested):
I know it was there before but `quit_msg[0] ? quit_msg : ""` seems pointless, `quit_msg` is the same.
Otherwise LGTM

michaelortmann:
> I know it was there before but `quit_msg[0] ? quit_msg : ""` seems pointless, `quit_msg` is the same. Otherwise LGTM

good find, cleanup done.

vanosg (approved):

vanosg (approved):

PR #1514, opened 2023-11-30
Fix status report for realname

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix status report for realname

`.status` shows the current content of variable `realname`, but should show the content of the realname used at the time eggdrop connected to the server or the last `311` it got.
Additionally max len for realname was upped from 80 to 130. Many servers had max len 50. But servers like inspircd have 130 (and are configurable up to 500). So i raised eggdrops limit from 80 to 130.
Additionally handing of unset realname was not consistent. if unset, 2 different defaults where used depending if unset in config file or unset through `.tcl set realname ""` during runtime. For the 2nd case default was `/msg LamestBot hello`. This PR changes both cases to default `/msg <nick> hello`.

`.tcl set realname testrealname`
connect to irc server
```
.status
.tcl set realname ""
.status
```
**Before:**
```
.status
    Online as: BotA!eggdrop@Clk-8553FB0D (testrealname)
.tcl set realname ""
.status
    Online as: BotA!eggdrop@Clk-8553FB0D ()
```
**After:**
```
.status
    Online as: BotA!eggdrop@Clk-8553FB0D (testrealname)
.tcl set realname ""
.status
    Online as: BotA!eggdrop@Clk-8553FB0D (testrealname)
```


src/eggdrop.h
```diff
@@ -49,10 +49,11 @@
...
  */
-#define CHANNELLEN 80 /* FIXME see issue #3 and issue #38 and rfc1459 <= 200 */
-#define HANDLEN    32 /* valid values 9->NICKMAX                             */
-#define NICKMAX    32 /* valid values HANDLEN->32                            */
-#define USERLEN    10
+#define CHANNELLEN  80  /* FIXME see issue #3 and issue #38 and rfc1459 <= 200 */
+#define HANDLEN     32  /* valid values 9->NICKMAX                             */
+#define NICKMAX     32  /* valid values HANDLEN->32                            */
+#define USERLEN     10
+#define REALNAMELEN 120
```
thommey: Where does the 120 limit come from?

vanosg:
Can we simplify (and unfortunately, maybe re-write) this process and just store the realname value when the Eggdrop does a WHOIS on itself upon connect and if/when it receives the CAP setrealname? Then have .status display that variable instead of botname? I think that is cleaner overall. 

Also, please don't mess with the default botnames, that is, like, history!

src/eggdrop.h
michaelortmann: god gave it to me in my sleep. like he gave it to the coder that wrote the former value of 80 down. What limit do you propose?

michaelortmann:
with the code in this PR, we will set variable used for `.status display` when WHOIS reply comes in.

ive got no idea about CAP setrealname. do you want to code the CAP stuff based in this PR? it could be added independently later also.

i didnt wanna mess with it. the problem was, it would be inconsistent, to have 2 default strings for the realname, but ive got an idea, we could leave the history as a comment.

src/eggdrop.h
michaelortmann: inspircd v3 has default 128:
https://docs.inspircd.org/3/configuration/
maxreal 	Number 	128 	The maximum length of a real name (gecos).

inspircd v4 has default 130:
https://docs.inspircd.org/4/configuration/
maxreal 	Number 	130 	The maximum length of a real name.

so lets change it to 130 assuming if it works even on inspircd, it will work on all.

PR #1523, opened 2024-01-11
Remove compat vsnprintf() (eggdrop requires posix 2001)

michaelortmann:
Found by: vanosg
Patch by: michaelortmann
Fixes: #226 

One-line summary:
Eggdrop depends on POSIX.1-2001, which defines `vsnprintf()`. 

This PR keeps `egg_vsnprintf()` for backward compatibility to (older) modules but redefines it to use `vsnprintf()` directly with a one-liner define in `eggdrop.h`.

All egg_vsnprintf() were replaced by vsnprintf():
`find ./ -type f -exec sed -i 's/egg_vsnprintf(/vsnprintf(/g' {} \;`
**Note: don't do this in git directory. cd src/ first!**

No functional change. I tested 2 share bots, which successfully started and shared userfile.


src/modules.c
```diff
@@ -521,7 +521,7 @@ Function global_table[] = {
   (Function) egg_inet_aton,
   /* 252 - 255 */
   (Function) egg_snprintf,
-  (Function) egg_vsnprintf,
```
thommey, line 524: Why not just export the real vsnprintf here for compatibility?

michaelortmann: because we already do like:
src/mod/module.h includes src/main.h
src/main.h includes eggdrop.h
and eggdrop.h has it defined for compatibility:
https://github.com/eggheads/eggdrop/pull/1523/files#diff-e717df86c10ed7c4ff2d25b284e548a7054e43d63d74dad305d7ecd5318ac1feR284

michaelortmann: but you made me thinking, we could as well define some symbols in modules.h instad of eggdrop.h when it is only for mod compatibility.

also the strcpyz() define could be streamlined.

just tell me if i shall do all this in this PR, or create a new Issue as a reminder for later

michaelortmann: ok, done,
now we keep binary mod compatibility
ready for review again.

PR #1547, opened 2024-03-16
Crashfix bot_away()

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Check return value of `partyidx`
Also treat a char as unsigned to avoid buffer[negative index]

Enhance error check and log for base64_to_int()
Found by botnet fuzzing combined with `-fsanitize=address`


vanosg:
Can we address the FIXMEs before we merge?

michaelortmann:
First FIXME got fixed.

Second FIXME was separated / replaced by the more general #1583

src/botcmd.c
```diff
@@ -62,11 +62,13 @@ static char base64to[256] = {
...
 int base64_to_int(char *buf)
 {
-  int i = 0;
+  int i = 0, j;
 
   while (*buf) {
-    i = i << 6;
-    i += base64to[(int) *buf];
+    j = base64to[(uint8_t) *buf];
+    if (i > ((INT_MAX >> 6) - j)) /* if overflow return -1 */
```
thommey, line 69: INT_MAX >> 6 smells like UD because it overflows int, no?, the old way was better if that's the case

thommey (changes requested):

src/mod/assoc.mod/assoc.c
```diff
@@ -336,9 +336,8 @@ static void zapf_assoc(char *botnick, char *code, char *par)
         add_assoc(par, chan);
         chanout_but(-1, chan, ASSOC_CHNAME_NAMED2, botnick, nick, par);
       }
-    } else {
-      dprintf(LOG_DEBUG, "ASSOC: Channel number outside bounds\n");
-    }
+    } else
+      dprintf(LOG_DEBUG, "ASSOC: Channel number outside bounds 0..%i\n", GLOBAL_CHANS);
```
thommey, line 340: undo this one change pls, then it's good (removing braces is the opposite of code style guidelines)

thommey (changes requested):

src/mod/assoc.mod/assoc.c
michaelortmann: absolutely right. done.

PR #1549, opened 2024-03-17
Fix SERVER_NICKLEN off-by-one error

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix SERVER_NICKLEN off-by-one error

https://github.com/eggheads/eggdrop/blob/4703f71067437c361ca8d5c8c4a09d41561a7a20/src/mod/filesys.mod/files.c#L433
https://github.com/eggheads/eggdrop/blob/4703f71067437c361ca8d5c8c4a09d41561a7a20/src/mod/modvals.h#L75
https://github.com/eggheads/eggdrop/blob/4703f71067437c361ca8d5c8c4a09d41561a7a20/src/mod/server.mod/server.h#L83-L84
Bug can be triggert by loading server mod and filesys mod, starting eggdrop and issuing `.files` and `get`
I didnt analyze the implication any further, but this has crash potential, because currently it would not only set nicklen to a random value (in my test 1447122753) but also trigger a load of misaligned address

### DIFF
--- a/src/mod/filesys.mod/files.c
+++ b/src/mod/filesys.mod/files.c
@@ -425,6 +425,12 @@ static void cmd_reget_get(int idx, char *par, int resend)
   long where = 0;
   int nicklen = NICKLEN;
 
+  if (!par[0]) {
+    dprintf(idx, "%s: %sget <file(s)> [nickname]\n", MISC_USAGE,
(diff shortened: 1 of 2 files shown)


thommey:
nice find!

PR #1550, opened 2024-03-17
Dont realloc (and fix use after free in 3rd party modules)

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Dont realloc

No need to realloc. Besides the cleaning up, it will help with real world problem. Modules like starts.mod copy channel member pointers and suffer from use after free. A problem that is very hard to fix (and find) in those modules.

No functional change for eggdrop core / channel mod.


thommey:
comment

thommey:
where does it do that?

thommey:
Closed for now, I don't think keeping the first pointer alive is helping much unless we observe a 3rd party module doing that.

PR #1559, opened 2024-04-07
Only import eggdrop python example

crazycatdevs:
Better way to import eggdrop, idea from @vanosg tested by @crazycatdevs


vanosg:
As discussed in channel, we're going to leave this as it is and document this specific use-case elsewhere.  The approach taken in the example is more 'pythonic' and we'll stick to that. We'll also discuss the core issue of duplicating the commands and see if we want to rename something. Thanks for testing this and identifying this issue!

PR #1568, opened 2024-04-26
Add logging of user and sys times for evaluated .tcl

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Add logging of user and sys times for evaluated .tcl commands

Additional description (if needed):

Test cases demonstrating functionality (if applicable):
```
.tcl source scripts/g_pbkdf2.tcl
[09:44:11] tcl: builtin dcc call: *dcc:tcl -HQ 1 source scripts/g_pbkdf2.tcl
[09:44:11] tcl: evaluating .tcl source scripts/g_pbkdf2.tcl
[09:44:11] tcl: evaluated .tcl source scripts/g_pbkdf2.tcl, user 64.542ms sys 64.414ms
Tcl: 
.tcl pbkdf2::pbkdf2 sha256 hunter 42 1500
[09:44:15] tcl: builtin dcc call: *dcc:tcl -HQ 1 pbkdf2::pbkdf2 sha256 hunter 42 1500
[09:44:15] tcl: evaluating .tcl pbkdf2::pbkdf2 sha256 hunter 42 1500
[09:44:15] tcl: evaluated .tcl pbkdf2::pbkdf2 sha256 hunter 42 1500, user 538.320ms sys 0.000ms
3V_nep¡Ð¤òüÌê »^Ã¡ÅÊT~3ö,
```

PR #1569, opened 2024-04-27
Add tcl pbkdf2 function

michaelortmann:
Found by: grawity
Patch by: michaelortmann
Fixes: 

One-line summary:
Add tcl pbkdf2 function

Useful for tcl scripts like https://github.com/grawity/eggdrop-sasl/blob/master/README.md#scram-sha-support that add SASL SCRAM mechanism, where the author said "which is very slow in Tcl so the server may time out". So this PR will help the author out with a fast replacement function exported by eggdrop. hope that helps and cheerz to grawity for his great work :)

The new tcl function pbkdf2() returns as hexadecimal string by default and -bin by option, which is similar, to what tcllibs sha256() does (older tcllibs md5 had it the other way around), see https://core.tcl-lang.org/tcllib/doc/trunk/embedded/md/tcllib/files/modules/sha1/sha256.md

Benchmark times were measured with #1568 applied
**Benchmark and Result of external tcl script providing pbkdf2():**
```
.tcl source scripts/g_pbkdf2.tcl
[03:05:34] tcl: builtin dcc call: *dcc:tcl -HQ 1 source scripts/g_pbkdf2.tcl
[03:05:34] tcl: evaluate (.tcl): source scripts/g_pbkdf2.tcl
Tcl: 
.tcl pbkdf2::pbkdf2 sha256 hunter 42 1500
[03:05:45] tcl: builtin dcc call: *dcc:tcl -HQ 1 pbkdf2::pbkdf2 sha256 hunter 42 1500
[03:05:45] tcl: evaluate (.tcl): pbkdf2::pbkdf2 sha256 hunter 42 1500
3V_nep¡Ð¤òüÌê »^Ã¡ÅÊT~3ö,
```
user 514.968ms sys 0.000ms
**Benchmark and result of new pmkdf2 mod function providing pbkdf2():**
```
.tcl pbkdf2 hunter 42 1500 sha256
[08:27:07] tcl: builtin dcc call: *dcc:tcl -HQ 1 pbkdf2 hunter 42 1500 sha256
[08:27:07] tcl: evaluate (.tcl): pbkdf2 hunter 42 1500 sha256
Tcl: 150D9333565F6E6570A193D0A4F2FC97CCEAA0BB5EC3A1C5CA9454067E33F62C
.tcl pbkdf2 -bin hunter 42 1500 sha256
[08:27:17] tcl: builtin dcc call: *dcc:tcl -HQ 1 pbkdf2 -bin hunter 42 1500 sha256
[08:27:17] tcl: evaluate (.tcl): pbkdf2 -bin hunter 42 1500 sha256
3V_nep¡Ð¤òüÌê »^Ã¡ÅÊT~3ö,
```
user 0.510ms sys 0.000ms
**Test and demo of replacement function for grawities script:**
```
.tcl source scripts/g_base64.tcl
.tcl source scripts/g_pbkdf2.tcl
```
orig func:
```
.tcl set saltedPassword [::pbkdf2::pbkdf2 sha256 hunter mysalt 1500]
.tcl b64:encode $saltedPassword
Tcl: rNb2VlyBIeNaSMc9YtVcthpvUoi1fPvAcN1PcHlc1J0=
```
replacement func:
```
.tcl set saltedPassword2 [pbkdf2 -bin hunter mysalt 1500 sha256]
.tcl b64:encode $saltedPassword2
Tcl: rNb2VlyBIeNaSMc9YtVcthpvUoi1fPvAcN1PcHlc1J0=
```
see also:
https://github.com/grawity/eggdrop-sasl/blob/master/g_scram.tcl#L156

### DIFF
--- a/src/mod/pbkdf2.mod/pbkdf2.c
+++ b/src/mod/pbkdf2.mod/pbkdf2.c
@@ -8,6 +8,7 @@
  */
 
 #include "src/mod/module.h"
+#include "src/mod/pbkdf2.mod/tclpbkdf2.c"
 
 #if OPENSSL_VERSION_NUMBER >= 0x1000000fL /* 1.0.0 */
 #define MODULE_NAME "encryption2"
@@ -222,45 +223,6 @@ static char *pbkdf2_verify(const char *pass, const char *encrypted)
     return pbkdf2_encrypt(pass);
   return (char *) encrypted;
 }
-
-static int tcl_encpass2 STDVAR
-{
-  BADARGS(2, 2, " string");
-  Tcl_SetResult(irp, pbkdf2_encrypt(argv[1]), TCL_STATIC);
-  return TCL_OK;
-}
-
-static int tcl_pbkdf2 STDVAR
-{
-  unsigned int rounds;
-  const EVP_MD *digest;
-  int digestlen;
-  unsigned char buf[256];
-
-  BADARGS(5, 5, " pass salt rounds digest");
-  rounds = atoi(argv[3]);
-  digest = EVP_get_digestbyname(argv[4]);
-  if (!digest) {
-    Tcl_AppendResult(irp, "PBKDF2 error: Unknown message digest '", argv[4], "'.", NULL);
-    return TCL_ERROR;
-  }
-  digestlen = EVP_MD_size(digest);
-  if (!PKCS5_PBKDF2_HMAC(argv[1], strlen(argv[1]), (const unsigned char *) argv[2], strlen(argv[2]), rounds, digest, digestlen, buf)) {
-    Tcl_AppendResult(irp, "PBKDF2 key derivation error: ", ERR_error_string(ERR_get_error(), NULL), ".", NULL);
-    return TCL_ERROR;
-  }
-  buf[digestlen] = 0;
-  Tcl_Obj *result = Tcl_NewByteArrayObj(buf, digestlen);
-  explicit_bzero(buf, digestlen);
-  Tcl_SetObjResult(irp, result);
-  return TCL_OK;
-}
-
-static tcl_cmds my_tcl_cmds[] = {
-  {"encpass2", tcl_encpass2},
-  {"pbkdf2",   tcl_pbkdf2},
-  {NULL,       NULL}
-};
 
 static tcl_ints my_tcl_ints[] = {
   {"pbkdf2-re-encode", &pbkdf2_re_encode, 0},
--- a/src/mod/pbkdf2.mod/pbkdf2.c
+++ b/src/mod/pbkdf2.mod/pbkdf2.c
@@ -226,15 +226,40 @@ static char *pbkdf2_verify(const char *pass, const char *encrypted)
 static int tcl_encpass2 STDVAR
 {
   BADARGS(2, 2, " string");
-  if (strlen(argv[1]) > 0)
-    Tcl_AppendResult(irp, pbkdf2_encrypt(argv[1]), NULL);
-  else
-    Tcl_AppendResult(irp, "", NULL);
+  Tcl_AppendResult(irp, pbkdf2_encrypt(argv[1]), NULL);
+  return TCL_OK;
+}
+
+static int tcl_pbkdf2 STDVAR
+{
+  unsigned int rounds;
+  const EVP_MD *digest;
+  int digestlen;
+  unsigned char buf[256];
+
+  BADARGS(5, 5, " pass salt rounds digest");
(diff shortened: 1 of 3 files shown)


vanosg:
I'm starting to dig into this in-depth, but a few initial suggestions. Could you please...
- Add this command to doc/tcl-commands.doc
- Add encpass2 to tcl-commands.doc ;) 
- Move (and create) these Tcl comamnds into tclpbkdf2.c

src/mod/pbkdf2.mod/pbkdf2.c
```diff
@@ -226,15 +226,39 @@ static char *pbkdf2_verify(const char *pass, const char *encrypted)
...
+}
+
+static int tcl_pbkdf2 STDVAR
+{
+  unsigned int rounds;
+  const EVP_MD *digest;
+  int digestlen;
+  unsigned char buf[256];
+
+  BADARGS(5, 5, " pass salt rounds digest");
```
vanosg: change 'pass' to 'string', since this could be used for anything?

src/mod/pbkdf2.mod/pbkdf2.c
```diff
@@ -226,15 +226,39 @@ static char *pbkdf2_verify(const char *pass, const char *encrypted)
...
+  rounds = atoi(argv[3]);
+  digest = EVP_get_digestbyname(argv[4]);
+  if (!digest) {
+    Tcl_AppendResult(irp, "PBKDF2 error: Unknown message digest '", argv[4], "'.", NULL);
+    return TCL_ERROR;
+  }
+  digestlen = EVP_MD_size(digest);
+  if (!PKCS5_PBKDF2_HMAC(argv[1], strlen(argv[1]), (const unsigned char *) argv[2], strlen(argv[2]), rounds, digest, digestlen, buf)) {
+    Tcl_AppendResult(irp, "PBKDF2 error: PKCS5_PBKDF2_HMAC(): ", ERR_error_string(ERR_get_error(), NULL), ".", NULL);
```
vanosg: Can we change this to something less technical instead of just PKCS5_PBKDF2_HMAC()? Maybe "PBKDF2: Hash generation error: ...."?

src/mod/pbkdf2.mod/pbkdf2.c
```diff
@@ -226,15 +226,39 @@ static char *pbkdf2_verify(const char *pass, const char *encrypted)
...
+    return TCL_ERROR;
+  }
+  buf[digestlen] = 0;
+  Tcl_Obj *result = Tcl_NewByteArrayObj(buf, digestlen);
```
vanosg: LOVE that you're using Objects! In my experience, I'm used to seeing hashes as a hex or base64 string instead of a byte array, and for encpass2 we're storing it as a base64 string... Can we do that here as well (base64 string) instead of returning this as a byte array, or does it need to be in byte format for a specific purpose? I'm trying to think through use cases and I think base64 will be more useful (and easier to handle/store) than a byte array

vanosg (commented):
This is a great feature to add!

src/mod/pbkdf2.mod/pbkdf2.c
michaelortmann: this function was about replacing the original tcl function, hence the same return type.

src/mod/pbkdf2.mod/pbkdf2.c
michaelortmann: pbkdf2 is really about passwords and nothing else. see: https://en.wikipedia.org/wiki/PBKDF2

michaelortmann:
>     * Add this command to doc/tcl-commands.doc

done

>     * Add encpass2 to tcl-commands.doc ;)

done

>     * Move (and create) these Tcl comamnds into tclpbkdf2.c

not sure, thats a good idea, because then pbkdf2.c and tclpbkdf2.c would mutually depend on each other. If you still want, i could add a shared header file to make such "separation".

michaelortmann:
done, tcl funcs separated into tclpbkdf2.c

src/mod/pbkdf2.mod/pbkdf2.c
michaelortmann: meanwhile the code was changed to output hex str by default, bin optional.

vanosg:
note for merge- run makedepend

vanosg:
As a real world test, can you hack this into grawity's script and confirm it works as expected?

michaelortmann:
> note for merge- run makedepend

this PR does not need makedepend

PR #1467, opened 2023-08-31
(RFC) WebUI mod

michaelortmann:
Found by: [michaelortmann](https://github.com/michaelortmann/)
Patch by: [michaelortmann](https://github.com/michaelortmann/)
Fixes: 

One-line summary:
WebUI mod adds webinterface to eggdrop

Additional description (if needed):
Screenshot desktop firefox: https://imgur.com/a/WONc3dR
Screenshot iOS safari: https://imgur.com/a/AQwxNgs (the imgur content detection/warning, is ridiculous ;)
Reuses eggdrops telnet interface
Adds minimal http/websocket
The html and [vanilla js](http://vanilla-js.com/) (best js framework ;) code is handcrafted
It sends our eggman as favicon :)
Work in progress, but already working for me
Eggdrops `make install` doesnt automatically copy from module to text directory, so you have to `cp src/mod/webui.mod/webui.html ~/eggdrop/text/` manually for now
Add to config:
```
loadmodule webui
listen 8080 webui
listen +8443 webui
```
Yes, the `+` means it already supports https

Test cases demonstrating functionality (if applicable):
Tested with destop firefox, desktop brave (based on chromium) and iphone iOS safari

PR #1581, opened 2024-05-07
Copy module text files

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Copy module text files

Tested with #1467, were we copy `src/mod/webui.mod/text/webui.html` to `$DEST/text`


vanosg:
Can you provide context on why we are copying these files?

michaelortmann:
> Can you provide context on why we are copying these files?

#1467 (webui.mod) will make use of it. but i think its a good idea in general, to copy module text/ files (if they dont exist yet).

michaelortmann:
currently there is no way for any module to create directories or copy files via module Makefile because module Makefiles do not have install targets defined

we cannot add/run a new install target to the module Makefile(s) because it would break any other module that doesnt define such target when the core Makefile tries to run such new target for each module

but how does eggdrop handle cases like filesys.mod? it does so in core Makefile via install target install-filesys

so the current state of webui.mod (#1467) will do like filesys.mod, see: https://github.com/eggheads/eggdrop/pull/1467/commits/7680021567b1265e5671e34e6aab3d4a7112e5e0#diff-b48650cabc0f502b52865f525765f2d688198f06e0ffb5456c76fabab7f2bfb2

PR #1585, opened 2024-05-14
Full account tacking

vanosg:
a PR that tracks adds get_user_from_channel, that calls both get_user_by_account() and get_user_by_host(). This will allow Eggdrop to track a user by their account name (first), and then check a hostname. The assumption here is that account names are more precise and less prone to collisions/spoofing than a hostmask. All the user protections and checks associated with handles can now be used based on account names instead of hosts.

### DIFF
--- a/doc/sphinx_source/using/tcl-commands.rst
+++ b/doc/sphinx_source/using/tcl-commands.rst
@@ -1136,7 +1136,7 @@ nick2hand <nickname> [channel]
   Module: irc
 
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-account2nicks <handle> [channel]
+account2nicks <account> [channel]
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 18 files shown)


vanosg:
Test setup- user Geo has a matching host and +o flag but no account, jack3 has a matching account and +o flag but no host, thommey has no matching info.

Testing Tcl chanlist, detects Geo by host and jack3 by account:
```
.tcl chanlist #eggheads +o
Tcl: Geo jack3
```
Testing Tcl handonchan, detects Geo and jack3 but not thommey
```
.tcl handonchan Geo
Tcl: 1
.tcl handonchan jack3
Tcl: 1
.tcl handonchan thommey
Tcl: 0
```

Testing Tcl hand2nicks:
```
.tcl hand2nicks geo
Tcl: Geo
.tcl hand2nicks jack3 
Tcl: jack3
.tcl hand2nicks thommey
Tcl: 
```

Testing Tcl nick2hand:
```
.tcl nick2hand geo
Tcl: Geo
.tcl nick2hand Jack3
Tcl: jack3
.tcl nick2hand thommey
Tcl: *
```

vanosg:
Testing with and without extended-join enabled; with extended-join Eggdrop properly op'd a user that only had a matching account (no hostmask) added.

Topic bind worked

Pub bind worked

Kick bind worked

PR #1589, opened 2024-05-22
Proposal for #1588

### LINKED ISSUE #1588: matchattr returns 0 with unknown handle and negative flag
In a script, I want to act only if user is not known as halfop or more.
Logical test:
`if {[matchattr $handle -l|-l $chan]} { ... do stuff }`

But if user is not known, matchattr returns 0, as if the user has the +l flag.
Culprits:
https://github.com/eggheads/eggdrop/blob/e28a0a023fe077fbec2343f8eb494371138fcd44/src/tcluser.c#L254 and https://github.com/eggheads/eggdrop/blob/e28a0a023fe077fbec2343f8eb494371138fcd44/src/tcluser.c#L258
If the user is not known, it returns default value which is 0.
Peharps could it be better to initialize it to -1 (user unknown), sets it to 0 if entering the condition and then do the stuff.

crazycatdevs:
Found by: CrazyCat
Patch by: CrazyCat
Fixes: #1588 

One-line summary:
initialize ok to -1, sets it to 0 if user is known

### DIFF
--- a/src/tcluser.c
+++ b/src/tcluser.c
@@ -251,11 +251,13 @@ static int tcl_matchattr STDVAR
 {
   struct userrec *u;
   struct flag_record plus = {0}, minus = {0}, user = {0};
-  int ok = 0, nom = 0;
+  int ok = -1;
+  int nom = 0;
 
   BADARGS(3, 4, " handle flags ?channel?");
 
   if ((u = get_user_by_handle(userlist, argv[1]))) {
+    ok = 0;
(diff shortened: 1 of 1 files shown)


thommey:
Unfortunately we can't do that, because you break code of the inverse logic:
```tcl
if {[matchattr $hand o|o $chan]} { # privileged code }
```
would suddenly give access to the "*" handle.

PR #1591, opened 2024-05-26
Fix python configure

### LINKED ISSUE #1587: Python issues
Documenting a few things here from a fresh install

Without python installed, this appears in ./configure
```
./configure: line 10630: EGG_PYTHON_ENABLE: command not found
```

With python3 installed but not the dev files:
```
./configure: line 10630: EGG_PYTHON_ENABLE: command not found
```
and
```
checking  version is >= 3.8.0... no ()
configure: WARNING: Eggdrop requires python version 3.8.0 or higher

  There was an issue with your python installation.
  Please read python.mod/config.log for more details.
  The python module will be disabled.
```

and still the same error as above once python3-dev is installed.

In addition to fixing this, we probably need to document somewhere what libraries are required to run python; similar to what we have done for Tcl

vanosg:
Found by: Csibesz`
Patch by: Geo
Fixes: #1587 

One-line summary:
Remove a leftover autoconf macro call, and also move around documentation so it is easier to find


vanosg:
@thommey we still need an unbind command

PR #1402, opened 2023-02-06
Workaround for usage of Emoji with default TCL_UTF_MAX==3

thommey:
Ugly workaround that does not break current usage of emoji in the Tcl API. Our usage of the Tcl API is not correct because Eggdrop is encoding agnostic, and processes byte arrays only. Those however get passed to Tcl as strings, when they are not. The real solution is fairly complex (decode all bytes into strings with the proper encoding guess from/to the Tcl API).

If users are unable to recompile Tcl with increased TCL_UTF_MAX and are stuck with 8.6 stable (8.7 is not released as stable at the time of this PR), this should be a suitable, albeit ugly and complex, workaround that keeps emoji support alive.

We do not wish to break the currently working emoji usage however, which mostly works in the current state, except for a crash if passed to [string tolower/toupper/totitle], other problematic call sites aren't known.

So we overwrite Tcl's string tolower/toupper/totitle with our own version that:
- detects utf-8 sequences that are 4 bytes
- converts them into utf-8 encoded surrogate pairs before Tcl's string functions are called
- those remain untouched by case changes in Tcl
- convert the surrogate pairs back in the result to the original utf-8 4-byte code sequences

Side effect is that we also convert surrogate pairs if they were sent as such, which should not have a negative impact on existing scripts.

Single-byte encodings being valid 4-byte emoji sequences or 6-byte surrogate pairs is hopefully highly unlikely.

Found by: Empus
Patch by: thommey
Fixes: #1163

PR #1597, opened 2024-05-31
Fix memleak in egg_string_unicodesup_desurrogate()

### LINKED ISSUE #1545: Eggdrop 1.9.5 - Possible memory leak in botnet communication?
I've seen a steady increase in memory usages of my eggdrops since i've migrated from 1.8.4 to 1.9.5.
Example of a bot : yesterday mem usage 143M today same bot 228M.
From what I can see in the memory dump it looks like the bot is having things in it's memory that passed yesterday over it's botnet.

I've tried .rehash and .restart from the bots commandline bot but the memory usage stays the same.
Only killing the bot from linux shell clears the memory.

michaelortmann:
Found by: michaelortmann (or Mystery-X if this fixes #1545, but unlikely, see https://github.com/eggheads/eggdrop/issues/1545#issuecomment-2041046997)
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix memleak in `egg_string_unicodesup_desurrogate()`

This bug affects only eggdrop 1.9.5 RC 1 (since #1402)

### DIFF
--- a/src/tcl.c
+++ b/src/tcl.c
@@ -767,6 +767,7 @@ Tcl_Obj *egg_string_unicodesup_desurrogate(const char *oldstr, int len)
 {
   int stridx = 0, bufidx = 0;
   char *buf = nmalloc(len);
+  Tcl_Obj *o;
 
   while (stridx < len) {
     uint32_t low, high;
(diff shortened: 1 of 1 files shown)


thommey:
nice find!

PR #1603, opened 2024-06-05
Update gseen.mod info

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Update gseen.mod info

We still want to worship / give credit to G'Quann by linking to his site,
but at the same time, we need to provide a link to an updated version that will work with newer / current eggdrop
We probably also need / want to change that info in doc/
 

vanosg:
Closed by 74c786165272a9a7c7a7374d1906d2b30201d862

PR #1618, opened 2024-06-15
add lookup_user_record()

vanosg:
Found by: Mystery-X
Patch by: Geo

It was discovered after the latest rewrite lookup up user records based on account that there are situations where m might not be known, and m being NULL can cause a crash if not properly checked. This creates a wrapper function to a) check which lookup to use based on the provided input, b) fallback to a different lookup in case something returns NULL.

Wrapper function to find an Eggdrop user record based on either a provided
channel memberlist record, host, or account. This function will first check
a provided memberlist and return the result. If no user record is found (or
the memberlist itself was NULL), this function will try again based on a
provided account, and then again on a provided host.

 When calling this function it is best to provide all available independent
variables- ie, if you provide 'm' for the memberlist, don't provide
'm->account' for the account, use the independent source variable 'account'
if available. This allows redundant checking in case of unexpected NULLs

### DIFF
--- a/src/cmds.c
+++ b/src/cmds.c
@@ -70,7 +70,7 @@ static int add_bot_hostmask(int idx, char *nick)
         struct userrec *u;
 
         egg_snprintf(s, sizeof s, "%s!%s", m->nick, m->userhost);
-        u = lookup_user_record(m, NULL, NULL);
+        u = get_user_from_member(m);
         if (u) {
           dprintf(idx, "(Can't add hostmask for %s because it matches %s)\n",
                   nick, u->handle);
@@ -1913,6 +1913,7 @@ static int add_to_handle(struct userrec *u, int idx, char *handle, char *host, i
     return 1;
   }
   if (type) {
+    // account
     u2 = lookup_user_record(NULL, host, NULL);
     if (u2) {
       dprintf(idx, "That account already exists for user %s\n", u2->handle);
@@ -1920,6 +1921,7 @@ static int add_to_handle(struct userrec *u, int idx, char *handle, char *host, i
     }
     addaccount_by_handle(handle, host);
   } else {
+    // host
     for (q = get_user(&USERENTRY_HOSTS, u); q; q = q->next) {
       if (!strcasecmp(q->extra, host)) {
         dprintf(idx, "That %s is already there.\n",
--- a/src/cmds.c
+++ b/src/cmds.c
@@ -70,7 +70,7 @@ static int add_bot_hostmask(int idx, char *nick)
         struct userrec *u;
 
         egg_snprintf(s, sizeof s, "%s!%s", m->nick, m->userhost);
-        u = get_user_by_host(s);
+        u = lookup_user_record(m, NULL, NULL);
         if (u) {
           dprintf(idx, "(Can't add hostmask for %s because it matches %s)\n",
                   nick, u->handle);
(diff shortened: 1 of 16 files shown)


src/mod/module.h
```diff
@@ -524,8 +524,9 @@ typedef void (*chanout_butfunc)(int, int, const char *, ...) ATTRIBUTE_FORMAT(pr
 #define bind_bind_entry ((int(*)(tcl_bind_list_t *, const char *, const char *, const char *))global[320])
 #define unbind_bind_entry ((int(*)(tcl_bind_list_t *, const char *, const char *, const char *))global[321])
 #define argv0 ((char *)global[322])
-#define get_user_from_member ((struct userrec * (*)(memberlist *))global[323])
+#define lookup_user_record ((struct userrec * (*)(memberlist *, char *, char *))global[323])
```
thommey, line 527: const pls

```suggestion
#define lookup_user_record ((struct userrec * (*)(memberlist *, const char *, const char *))global[323])
```

src/cmds.c
```diff
@@ -1907,13 +1907,13 @@ static int add_to_handle(struct userrec *u, int idx, char *handle, char *host, i
       return 1;
     }
   }
-  if ( !type && !glob_botmast(fr) && !chan_master(fr) && get_user_by_host(host)) {
+  if ( !type && !glob_botmast(fr) && !chan_master(fr) && lookup_user_record(NULL, NULL, host)) {
     dprintf(idx, "You cannot add %s matching another user!\n",
```
thommey: the type condition on that dprintf is unnecessary, the if-condition already includes it, we're guaranteed to be dealing with  hostmask

src/mod/irc.mod/chan.c
```diff
@@ -333,7 +325,7 @@ static int detect_chan_flood(char *floodnick, char *floodhost, char *from,
     chan->floodwho[which][0] = 0;
     if (which == FLOOD_DEOP)
       chan->deopd[0] = 0;
-    u = get_user_from_member(m);
+    u = lookup_user_record(m, NULL, from);
```
thommey: missing the account name we might have, m is always true for everything but FLOOD_JOIN so I'd try to make that clear in the code:

```suggestion
    if (which == FLOOD_JOIN) {
      u = lookup_user_record(m, victim_or_account, from);
    } else {
      u = get_user_from_member(m);
    }
```

src/mod/server.mod/servmsg.c
```diff
@@ -493,7 +493,7 @@ static int detect_flood(char *floodnick, char *floodhost, char *from, int which)
   if (!strcasecmp(floodhost, botuserhost))
     return 0;
 
-  u = get_user_by_host(from);
```
thommey, line 496: can't we get an account into this function, same as the other flood function?

thommey (changes requested):

thommey:
More generally I think I disagree with changing `get_user_from_member(m)` to `lookup_user_record(m, NULL, NULL)` or even `lookup_user_record(m, NULL, from)`.
`lookup_user_record(m, NULL, from)` suggests m could be NULL when in almost all of those cases it can't be.
Using `get_user_from_member(m)` is less confusing in my opinion when we know m isn't NULL.

The nickname search combination `lookup_user_record & check_all_chan_records` should be folded into a single thing in my opinion, because it is always used together and is an unnecessary extra step. It could just get a nickname and call get_user_from_member(m) on the first one it finds and return that result.

src/mod/server.mod/servmsg.c
vanosg: Not without making big changes to the larger codebase

vanosg:
The (m, NULL, NULL) is necessary and only used in cases where host is derived from m. If m doesn't exist, m-> doesn't exist. These cases don't have a 'from' or similar variable to pull the host from.

vanosg (commented):
should be ```    u2 = lookup_user_record(NULL,  NULL, host);  ``` ?

PR #1622, opened 2024-06-16
Update configure.ac

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Update configure.ac

`isascii()` is POSIX 2001

There is no actual known bug, but good configure diagnostics are helpful in catching any, and this PR enhances those.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -91,6 +91,7 @@ EGG_CHECK_OS
 # Checks for header files.
 EGG_HEADER_STDC
 AC_HEADER_DIRENT
+AC_HEADER_RESOLV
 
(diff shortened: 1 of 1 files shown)


thommey:
What good does AC_HEADER_RESOLV do if we don't use it? At least this PR doesn't add using it

michaelortmann:
> What good does AC_HEADER_RESOLV do if we don't use it? At least this PR doesn't add using it

We want good information from configure run, to be able to debug reports from users. So checking for the more "exotic" futures.

michaelortmann:
Alright, AC_HEADER_RESOLV doesnt really add anything.

PR #1623, opened 2024-06-16
Race fix: tdns: strerror() -> strerror_r()

michaelortmann:
Found by: thommey and michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix locking  / race for tdns error logging

Non main threads like tdns must use `strerror_r()` instead of `strerror()`.

```
valgrind --tool=helgrind ./eggdrop -nt BotA.conf
[...]
.tcl dnslookup google.de foo
```

Without this patch applied, valgrind will report race


thommey:
```
       int strerror_r(int errnum, char buf[.buflen], size_t buflen);
                      /* XSI-compliant */
       char *strerror_r(int errnum, char buf[.buflen], size_t buflen);
                      /* GNU-specific */
       strerror_r():
           The XSI-compliant version is provided if:
               (_POSIX_C_SOURCE >= 200112L) && ! _GNU_SOURCE
           Otherwise, the GNU-specific version is provided.
```

I have no words, but both work in our context, LGTM

thommey (approved):

PR #1632, opened 2024-06-29
Allow python threads and async io

thommey:
Found by:
Patch by: thommey
Fixes: 

One-line summary: Allows python to use threads and async io
Python threads were not given a chance to run in the background so far because we held onto the GIL.
Now we release it around our select() which should give Python threads or asyncio runners (in a separate thread) enough room to work.

The concern of having multithreaded code call into our Tcl interpreter against its will turns out to be unwarranted - the GIL takes care of this.


thommey:
Test case 1 - Threads:
```py
from eggdrop.tcl import putlog
from threading import Thread
from time import sleep

def tick():
    while True:
        putlog("thread running")
        sleep(.2)

thread = Thread(target = tick)
thread.start()
```

Test case 2 - Async IO with runner in separate Thread
```py
from eggdrop.tcl import putlog
import asyncio
from threading import Thread

async def main():
    while True:
        putlog("asyncio running")
        await asyncio.sleep(.3)

def async_run():
    asyncio.run(main())

thread = Thread(target = async_run)
thread.start()
```

PR #1411, opened 2023-02-13
Warn user of expired cert

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Warn user of expired cert

Additional description (if needed):
This is about the cert loaded from config variable `ssl-certificate`
The check is done when
- the cert is loaded in `ssl_init()`
- daily
- when a +n owner logs in

**Please run `misc/runautotools` when testing / merging this PR**

Test cases demonstrating functionality (if applicable):
```
$ ./eggdrop -t BotA.conf 
[...]
WARNING: certificate expired: BotA.crt

### ENTERING DCC CHAT SIMULATION ###
[...]
```

PR #1641, opened 2024-07-09
Document minimum library versions

vanosg:
@thommey any disagreements with what you see here?

Once we figure out what it is, need to update doc/TLS to reflect minimum TLS version as well


michaelortmann:
We dont need to raise openssl req to 1.0.2. #1411 now supports 0.9.8 also.

thommey:
good idea, Tcl is >= 8.5, otherwise fine

PR #1650, opened 2024-07-13
Python unbind functionality

thommey:
```python
from eggdrop import bind
from eggdrop.tcl import putmsg

def test(nick, host, hand, chan, text):
    putmsg(chan, "it worked once and once only")
    xxx.unbind()

xxx = bind("pub", "-", "!test", test)
```


thommey:
Not happy with the static buffers yet

skralg (approved):

skralg (approved):

PR #1660, opened 2024-07-29
Github action for autoconf

thommey:
Patch by: Geo


skralg (approved):

vanosg:
Do we want to add a makedepend option as well?

thommey:
The end goal is to have both automated as necessary, at least on develop that should be straight forward (just need to find a way to introduce "quiet" time so it doesn't run it while we merge multiple PRs).

autotools has a specific autoconf version requirement so not anyone can run it - that's why the action is important right now.

I would defer makedepend to when we automate both (and docs generation etc.).

PR #1695, opened 2024-10-07
Fix things not working when there is a bind log

DasBrain:
When there is a `bind log`, a few things stop working, as they do not expect logging functions to change the Tcl interpreter result.

This fixes those issues by storing the result before triggering the log binds and restoring it after the call.

Things found to be broken:

* `.tcl` dcc command (empty result is shown)
* Tcl errors reported in bind.

And maybe a few more.

Found by: DasBrain
Patch by: DasBrain
Fixes: bind log, hopefully.

One-line summary:
Fix wrong Tcl results when there is a bind log.

.tcl string cat foo
.tcl bind log * +|+ {string cat}
.tcl string cat foo

### DIFF
--- a/src/tclhash.c
+++ b/src/tclhash.c
@@ -1240,13 +1240,19 @@ void check_tcl_die(char *reason)
 void check_tcl_log(int lv, char *chan, char *msg)
 {
   char mask[512];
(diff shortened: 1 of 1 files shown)


thommey (approved):
LGTM

PR #1696, opened 2024-10-07
Log account changes as join, not mode console level.

DasBrain:
(Only channel ops can change modes, users can change their own account)

Found by: DasBrain
Patch by: DasBrain
Fixes: 

One-line summary:
Account changes are logged using +j (join, parts, quits, aways and netsplits) not +k (kick, bans and mode changes) console level.

### DIFF
--- a/src/mod/irc.mod/chan.c
+++ b/src/mod/irc.mod/chan.c
@@ -106,9 +106,9 @@ static void setaccount(char *nick, char *account)
         /* account was known */
         if (m->account[0]) {
           if (!strcmp(account, "*")) {
-            putlog(LOG_MODES, chan->dname, "%s!%s has logged out of their account", nick, m->userhost);
+            putlog(LOG_JOIN, chan->dname, "%s!%s has logged out of their account", nick, m->userhost);
           } else {
-            putlog(LOG_MODES, chan->dname, "%s!%s logged in to their account %s", nick, m->userhost, account);
+            putlog(LOG_JOIN, chan->dname, "%s!%s logged in to their account %s", nick, m->userhost, account);
           }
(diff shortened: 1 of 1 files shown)


thommey:
what's the rationale for changing it? we deliberately thought JOIN is suitable

DasBrain:
Currently account changes are logged as mode changes, not join.

A channel member doing a `/msg NickServ identify` should is not a mode change, which may for example be audited.
(A log with mode changes/kicks can be used as audit log for operator actions - an ordinary channel member logging some stuff in there doesn't fit it in my opinion.) 

So I agree:

> we deliberately thought JOIN is suitable

vanosg:
I think @thommey misspoke on this one- I think we deliberately made it a MODE flag, not JOIN flag, because join is associated with a single channel and an account change is more global in nature. I do recall a very large discussion we had on this topic at the time but can't remember details any more... While I was initially in favor of JOIN, I think my mind was changed, but I'm not sure why it is MODE now. Maybe SERV?

DasBrain:
Hard to argue with that reasoning. So, yeah...

I view ACCOUNT messages similar to NICK changes.  
Both apply to the user in all common channels.

NICK is also logged in +j, if I remember correctly.

And I think I already explained why I prefer account changes on +j: To not clutter up some "operator audit log" - anything else in +k requires channel operator status to trigger.

vanosg:
Just to document here- if you set console to * and are in two channels, you'll see this:
```
[20:59:15] Nick change: Geo -> Geo_
[20:59:15] Nick change: Geo -> Geo_
```
on a nick change. 
So I guess no reason to not add account there as well.... last chance for opposing views!

Robby-:
Also not opposed to this, having read the reasoning to change this, it kinda makes more sense for this to indeed be in `LOG_JOIN`.

PR #1709, opened 2024-10-27
Crashfix: PART

### LINKED ISSUE #1708: SEGMENT VIOLATION on user /part
Hello 

i run an eggdrop v1.10.0 on Linux 6.1.106-1-MANJARO with python module. (tested without python module)
Eggdrop crash when user /part just after bot join channel, with or without script loaded.

Tested without script and with only vital module. 

Edit1: It seem happen only during join phase of the bot... If no /part during something like one minute, the /part not produce crash... 

``[18:57:13] * Last bind (may not be related): evnt:init_server``
``[18:57:12] * Wrote DEBUG``
``[18:57:12] * SEGMENT VIOLATION -- CRASHING!``

``[18:57:12] * eggdrop(lamest@fooBar) vient de rentrer``
``[18:57:13] * RandomUser (foo@Bar) est parti``
``[18:57:13] * eggdrop a quitté (Connection closed)``

Reproduce : 
Just compile an eggdrop 1.10.0, 
comment the two die line in config, 
add a serveur, 
add some channels, 
launch, 
make user /part just after bot join, 
crash...

[DEBUG.log](https://github.com/user-attachments/files/17510353/DEBUG.log)

Edit: After a few tests, random /part can produce crash

michaelortmann:
Found by: BigBadWouf
Patch by: michaelortmann
Fixes: #1708

One-line summary:
Crashfix `gotpart()`: If `ismember()` returns `0`, fallback to `get_user_by_host()`

`ismember()` can returns `0` while the bot has not (yet) synched the channel, like when it just joined
This bug affects eggdrop version 1.10.0, older versions are not affected

Fixes the following crash:
```
[22:08:48] [@] :testuser!~michael@localhost PART #test666
[22:08:48] * Please report problem to https://github.com/eggheads/eggdrop/issues
[22:08:48] * Check doc/BUG-REPORT on how to do so.
[22:08:48] * Last bind (may not be related): evnt:init_server
[22:08:48] * Wrote DEBUG
[22:08:48] * SEGMENT VIOLATION -- CRASHING!
Segmentation fault (core dumped)
```

### DIFF
--- a/src/mod/irc.mod/chan.c
+++ b/src/mod/irc.mod/chan.c
@@ -2240,7 +2240,7 @@ exit:
  */
 static int gotpart(char *from, char *msg)
 {
-  char *nick, *chname, *key;
+  char *nick, *chname, uhost[UHOSTLEN], *key;
   struct chanset_t *chan;
   struct userrec *u;
   memberlist *m;
@@ -2255,9 +2255,13 @@ static int gotpart(char *from, char *msg)
     return 0;
   }
   if (chan && !channel_pending(chan)) {
+    strlcpy(uhost, from, sizeof uhost);
     nick = splitnick(&from);
     m = ismember(chan, nick);
-    u = get_user_from_member(m);
+    if (m)
+      u = get_user_from_member(m);
+    else
+      u = get_user_by_host(uhost);
     if (!channel_active(chan)) {
       /* whoa! */
       putlog(LOG_MISC, chan->dname,
--- a/src/mod/irc.mod/chan.c
+++ b/src/mod/irc.mod/chan.c
@@ -2258,6 +2258,7 @@ static int gotpart(char *from, char *msg)
     strlcpy(uhost, from, sizeof uhost);
     nick = splitnick(&from);
     m = ismember(chan, nick);
+    // TODO: check account from rawt account-tags
     if (m)
       u = get_user_from_member(m);
     else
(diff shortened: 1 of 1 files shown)


BigBadWouf:
Hi,

Thank you, this perfectly fix the /part crash. However it's seem produce a lot call to killmember(). Don't know if related

michaelortmann:
Good find. The `killmember()` was already called before the regression.

So now, additionally this PR avoids calling `killmember()` if we just checked that the parting entity is not `ismember()`

`ismember()` already checks the very channel memberlist that `killmember()` would search through again.

src/mod/irc.mod/chan.c
```diff
@@ -2255,9 +2255,13 @@ static int gotpart(char *from, char *msg)
     return 0;
   }
   if (chan && !channel_pending(chan)) {
+    strlcpy(uhost, from, sizeof uhost);
     nick = splitnick(&from);
     m = ismember(chan, nick);
-    u = get_user_from_member(m);
+    if (m)
```
thommey, line 2262:
```suggestion
    // TODO: check account from rawt account-tags
    if (m)
```

thommey (commented):
Good fix, just add that TODO comment please whenever we use straight get_user_from_host because we want to look at those instances again in the future to work with a possible account-tag we might've gotten (needs to change raw bind to rawt etc.), not part of this PR, just as a comment, we have similar ones in other places already.

src/mod/irc.mod/chan.c
michaelortmann: done

thommey (approved):

PR #1722, opened 2024-12-18
Fix .tcl server add/remove arg parsing

### LINKED ISSUE #1721: `.tcl server add` crashes eggdrop
Responsible lines: https://github.com/eggheads/eggdrop/blob/develop/src/mod/server.mod/tclserv.c#L583-L585

`argv[2]` does not exist if the command is called with only one argument.

michaelortmann:
Found by: [DasBrain](https://github.com/DasBrain)
Patch by: michaelortmann
Fixes: #1721

One-line summary:

**Before:**
`.tcl server add/remove` -> crash
**After:**
```
.tcl server add
Tcl error: wrong # args: should be "server add host ?port? ?password?"
.tcl server remove
Tcl error: wrong # args: should be "server remove host ?port? ?password?"
```

### DIFF
--- a/src/mod/server.mod/tclserv.c
+++ b/src/mod/server.mod/tclserv.c
@@ -604,7 +604,7 @@ static int tcl_server STDVAR {
     }
   } else if (!strcmp(argv[1], "remove")) {
     if (argc < 3) {
-      Tcl_SetResult(irp, "wrong # args: should be \"server remove host ?port? ?password?\"", TCL_STATIC);
+      Tcl_SetResult(irp, "wrong # args: should be \"server remove host ?port?\"", TCL_STATIC);
       return TCL_ERROR;
     }
     ret = del_server(argv[2], argc >= 4 && argv[3] ? argv[3] : "");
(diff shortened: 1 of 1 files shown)


src/mod/server.mod/tclserv.c
```diff
@@ -582,6 +582,10 @@ static int tcl_server STDVAR {
 
   BADARGS(2, 5, " subcommand ?host ?port? ?password?");
   if (!strcmp(argv[1], "add")) {
+    if (argc < 3) {
+      Tcl_SetResult(irp, "Subcommand add needs at least a server host", TCL_STATIC);
```
DasBrain: I would use the error message `wrong # args: should be "server add host ?port? ?password?"`

src/mod/server.mod/tclserv.c
```diff
@@ -599,6 +603,10 @@ static int tcl_server STDVAR {
       Tcl_SetObjResult(irp, server);
     }
   } else if (!strcmp(argv[1], "remove")) {
+    if (argc < 3) {
+      Tcl_SetResult(irp, "Subcommand remove needs at least a server host", TCL_STATIC);
```
DasBrain: I would use the error message `wrong # args: should be "server remove host ?port? ?password?"`

src/mod/server.mod/tclserv.c
michaelortmann: done

src/mod/server.mod/tclserv.c
michaelortmann: done

PR #1727, opened 2025-01-06
Add Git commit hash (if available) to .status all

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Adds Git commit hash (if available) to `.status all`

Think of this as additional version information, good for bug reports / debugging.

```
.status all
[01:38:07] tcl: builtin dcc call: *dcc:status -HQ 1 all
[01:38:07] #-HQ# status all
I am BotA, running eggdrop v1.10.0+pythonfixes: 23 users (mem: 276k).
Online for 00:00 (terminal mode) - CPU: 00:00.01 - Cache hit: 66.7%
Configured with: 
Admin: Lamer <email: lamer@lamest.lame.org>
Config file: BotA.conf
OS: Linux 6.12.8-arch1-1
Process ID: 253975 (parent 191282)
Tcl library: /usr/lib/tcl8.6
Tcl version: 8.6.16 (header version 8.6.16)
Tcl is threaded.
TLS support is enabled.
TLS library: OpenSSL 3.4.0 22 Oct 2024 (header version OpenSSL 3.4.0 22 Oct 2024)
IPv6 support is enabled.
Threaded DNS core is enabled.
Socket table: 20/100

Git commit: 4e5db356baa1b85a9e314988b06ebe844697c8e8
[...]
```


michaelortmann:
If we had the commit hash in version info, it would show up in https://www.eggheads.org/uptime/, wouldnt that be nice?

Lord255:
as i wrote in chat, i really wish that after each release (or commit to dev branch) the version would be visible and could be distinguished, so like after a release the git version (develop) would be bumped to next version+git or something.,
storing the hash is an awesome idea, but i wish to see the version during compile and startup (banner).
example: 1.10.2 was released, git develop should be bumped either to 1.10.3+git or something.
also i saw some bumps after some added features, but not after all commits. mhm.

PR #1738, opened 2025-01-12
strncpy() + manual NULL terminating -> strlcpy()

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
`strncpy()` + manual `NULL` terminating -> `strlcpy()`

Old code uses `strncpy()` combined with manual `NULL` terminating.
This is exactly what `strlcpy()` is good for.
So lets replace the few remaining `strncpys()` and test each of those code pathes.

1st commit was tested by linking 2 bots. In this case `addbot()` is called and the new code is executed. `HANDLEN + 1` == `sizeof ptr2->bot` == `33`.

### DIFF
--- a/src/botnet.c
+++ b/src/botnet.c
@@ -88,8 +88,7 @@ void addbot(char *who, char *from, char *next, char flag, int vernum, int ssl)
     ptr = &((*ptr)->next);
   }
   ptr2 = nmalloc(sizeof(tand_t));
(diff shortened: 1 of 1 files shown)


thommey:
This is fine, but marked as WIP, are you still working on this?

michaelortmann:
i was thinking to look for more such cases, and replace them also. but we could merge this, as currently i have more interesting stuff todo :)

PR #1759, opened 2025-02-01
Crash fix add_note() - Followup to #1763

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix add_note()

In addition to finding the bug in February 2025, and fixing it, i have now also been able to automatically find it with https://github.com/AFLplusplus/AFLplusplus

```
CFLAGS="-fsanitize=address" ./configure
```
Before:
```
.note CpgevlUVBdudySxBQErFYkoOBSabbVjuiLvuPkpvFjSZpWvcoomkbSoqPDLKwckeatNBMXuhjyJwwktER: a
[14:37:05] tcl: builtin dcc call: *dcc:note -HQ 1 CpgevlUVBdudySxBQErFYkoOBSabbVjuiLvuPkpvFjSZpWvcoomkbSoqPDLKwckeatNBMXuhjyJwwktER: a
=================================================================
==197765==ERROR: AddressSanitizer
```
After:
```
.note CpgevlUVBdudySxBQErFYkoOBSabbVjuiLvuPkpvFjSZpWvcoomkbSoqPDLKwckeatNBMXuhjyJwwktER: a
[06:02:24] tcl: builtin dcc call: *dcc:note -HQ 1 CpgevlUVBdudySxBQErFYkoOBSabbVjuiLvuPkpvFjSZpWvcoomkbSoqPDLKwckeatNBMXuhjyJwwktER: a
[06:02:24] add_note(): sock not a number
.note 424242424242424242424242424242424242424242424242424242424242424242424242424242424: a
[06:02:28] tcl: builtin dcc call: *dcc:note -HQ 1 424242424242424242424242424242424242424242424242424242424242424242424242424242424: a
[06:02:28] add_note(): sock out of range
.note 42: a
[06:02:35] tcl: builtin dcc call: *dcc:note -HQ 1 42: a
I don't know anyone by that name.
```


src/misc.c
```diff
@@ -207,14 +209,11 @@ void splitc(char *first, char *rest, char divider)
...
-/*    As above, but lets you specify the 'max' number of bytes (EXCLUDING the
- * terminating null).
+/* As above, but lets you specify the 'max' number of bytes
  *
  * Example of use:
  *
- * char buf[HANDLEN + 1];
- *
- * splitcn(buf, input, "@", HANDLEN);
+ * splitcn(buf, input, "@", sizeof buf);
```
thommey: do you mean `'@'` instead of `"@"`?

src/botmsg.c
```diff
@@ -755,7 +755,7 @@ int add_note(char *to, char *from, char *msg, int idx, int echo)
 {
   #define FROMLEN 40
   int status, i, iaway, sock;
-  char *p, botf[FROMLEN + 1 + HANDLEN + 1], ss[81], ssf[20 + 1 + sizeof botf];
+  char *p, botf[FROMLEN + 1 + HANDLEN + 1], ss[21], ssf[20 + 1 + sizeof botf], *endptr;
```
thommey, line 761: why is 21 correct?

src/botmsg.c
```diff
@@ -812,14 +812,19 @@ int add_note(char *to, char *from, char *msg, int idx, int echo)
...
   rmspace(ssf);
-  splitc(ss, to, ':');
+  splitcn(ss, to, ':', sizeof ss);
   rmspace(ss);
   if (!ss[0])
     sock = -1;
-  else
-    sock = atoi(ss);
+  else {
+    sock = strtoul(ss, &endptr, 10);
```
thommey: I would prefer a(n) (unsigned) long as temporary variable, do the range check, then cast to int.
And braces around this if/else construct

src/mod/module.h
```diff
@@ -528,6 +528,7 @@ typedef void (*chanout_butfunc)(int, int, const char *, ...) ATTRIBUTE_FORMAT(pr
 /* 324 - 327 */
 #define find_member_from_nick ((memberlist * (*) (char *))global[324])
 #define get_user_from_member ((struct userrec * (*) (memberlist *))global[325])
+#define splitcn ((void (*)(char *, char *, char, size_t))global[326])
```
thommey: is one of these actually const char ptr as the input?

thommey (changes requested):

src/botmsg.c
michaelortmann: ss is used as the destination of long-> string conversion. 21 is exactly that. try to lower it by one and your compiler will likely yell at you. longer numbers will be truncated there which is fine.

src/misc.c
michaelortmann: why is compiler silent about this? they should have made a call of a function with prototype `char` called with a `char*` an error with c23 instead of destroying our mod API :P

michaelortmann: ah its in comment ;P

michaelortmann: fixed.

src/mod/module.h
michaelortmann: none of the char * params. the 3rd paramter char could be const. but i dont see any benefit in
splitcn ((void (*)(char *, char *, const char, size_t)
do you?

src/botmsg.c
michaelortmann: Done.

michaelortmann: in case you wonder about any of those bits, i took it mostly from here: https://man.openbsd.org/strtol

src/mod/module.h
michaelortmann: ready for review again.

thommey (changes requested):
add_note is called with idx=-1/-2 on purpose sometimes (e.g. dcc.c:1976 from dcc_telnet_pw()). This warning message then lands on stdout/stderr (-1/-2 to dprintf are interpreted that way which isn't intended (putlog LOG_MISC instead perhaps?).

michaelortmann:
@thommey good catch. i added checks in the same style as the rest of the functions dprintf()s.

src/botmsg.c
thommey: Add a comment saying that please :)

src/mod/module.h
```diff
@@ -534,6 +534,7 @@ typedef void (*chanout_butfunc)(int, int, const char *, ...) ATTRIBUTE_FORMAT(pr
 #define stealth_telnets (*(int *)(global[328]))
 #define parse_irc ((parsed_irc_t (*)(char *))global[329])
 #define join_str_array ((char *(*)(char **, int, char *, char *, size_t))global[330])
+#define splitcn ((void (*)(char *, char *, char, size_t))global[321])
```
thommey: Uhm ... do you mean `331`?

thommey (changes requested):

src/mod/module.h
michaelortmann: fixed merge error. ty.

vanosg (approved):

PR #1776, opened 2025-02-11
Cleanup includes

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Simple cleanup of includes that are already included in `main.h`

`main.h` is included by `src/mod/module.h` and so on.

No functional change

### DIFF
--- a/src/compat/explicit_bzero.c
+++ b/src/compat/explicit_bzero.c
@@ -47,7 +47,6 @@
 #ifndef __STDC_WANT_LIB_EXT1__ /* SunOS */
 # define __STDC_WANT_LIB_EXT1__ 1
 #endif
-#include <string.h>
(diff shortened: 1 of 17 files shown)


vanosg:
Should we take this opportunity to review the same for sys/socket.h, netinet/in.h, fcntl.h, setjmp.h, signal.h?

michaelortmann:
`fcntl.h`, `setjmp.h` and `signal.h` are not currently in eggdrop.h nor main.h. so i did not touch it to not bloat the core header files of eggdrop. instead of moving those definitions to core header files i would like to include files only the files where they are really needed.

For `sys/socket.h` and `netinet/in.h` you are correct, we can do additional clean up.

PR #1783, opened 2025-02-20
Fix and enhance memory usage regarding log table

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Enhance memory usage regarding log table - Fix regression since eggdrop 1.9.0rc1

This PR not only fixes the regression since eggdrop 1.9.0rc1, but makes eggdrop even better than before regarding amount of memory allocation for log table
During eggdrop startup, memory for log tables is allocated:
https://github.com/eggheads/eggdrop/blob/341477a3fba56ec1b3f5a5b67fee3dec35878f2c/src/misc.c#L92-L101
about `max-logs * LOGLINELEN` bytes
default `max-logs` is `20`
`LOGLINELEN` was rised from `1024` to `9000` since eggdrop 1.9.0rc1
`20 * 9000 = 180 kbytes`
Users probably raise `max-logs` to `100` or more in their configuration files so they wont have to care about it anymore, not aware of the memory usage implications.

This PR makes eggdrop to still allocate memory for `max-logs` log tables. but only about 56 bytes per table / logfile static. `szlast` (last log messages to track repeats) will be allocated (grow) dynamically until `LOGLINELEN`.

`.status` will now show the log table information, similar to socket table (and memory table, if debug enabled)

Add MIN() and MAX() to eggdrop.h

```
$ grep max-logs BotA.conf 
set max-logs 100
```
Before:
```
[michael@zen ~]$ ps u|grep egg
michael   861664  0.0  0.0  18556 13664 pts/4    S+   04:41   0:00 ./eggdrop -t BotA.conf
```
After:
```
[michael@zen ~]$ ps u|grep egg
michael   862487  0.0  0.0  17668 12956 pts/4    S+   04:42   0:00 ./eggdrop -t BotA.conf

.status
Log table: 1/100 5728 bytes
```

### DIFF
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -312,6 +312,14 @@ void tell_verbose_status(int idx)
                "Threaded DNS core is disabled.\n"
 #endif
                "Socket table: %d/%d\n", threaddata()->MAXSOCKS, max_socks);
+  int j = 0;
+  int k = max_logs * sizeof(log_t);
+  for (int i = 0; i < max_logs; i++) {
+    if (logs[i].filename)
+      j++;
+    k += logs[i].szlast_len;
+  }
+  dprintf(idx, "Log table: %d/%d %d bytes\n", j, max_logs, k);
 }
 
 /* Show all internal state variables
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -313,13 +313,13 @@ void tell_verbose_status(int idx)
 #endif
                "Socket table: %d/%d\n", threaddata()->MAXSOCKS, max_socks);
   int j = 0;
-  int k = max_logs * sizeof(log_t);
+  size_t k = max_logs * sizeof(log_t);
   for (int i = 0; i < max_logs; i++) {
     if (logs[i].filename)
       j++;
     k += logs[i].szlast_len;
   }
-  dprintf(idx, "Log table: %d/%d %d bytes\n", j, max_logs, k);
+  dprintf(idx, "Log table: %d/%d %zu bytes\n", j, max_logs, k);
 }
 
 /* Show all internal state variables
--- a/src/chanprog.c
+++ b/src/chanprog.c
@@ -414,6 +414,11 @@ void chanprog()
         nfree(logs[i].chname);
         logs[i].chname = NULL;
       }
+      if (logs[i].szlast != NULL) {
+        nfree(logs[i].szlast);
+        logs[i].szlast = NULL;
+        logs[i].szlast_len = 0;
+      }
       if (logs[i].f != NULL) {
         fclose(logs[i].f);
         logs[i].f = NULL;
--- a/src/misc.c
+++ b/src/misc.c
@@ -95,20 +95,9 @@ void init_misc()
 
   if (max_logs < 1)
     max_logs = 1;
-  if (logs)
-    logs = nrealloc(logs, max_logs * sizeof(log_t));
-  else
-    logs = nmalloc(max_logs * sizeof(log_t));
-  for (; last < max_logs; last++) {
-    logs[last].filename = logs[last].chname = NULL;
-    logs[last].mask = 0;
-    logs[last].f = NULL;
-    /* Added by cybah  */
-    logs[last].szlast[0] = 0;
-    logs[last].repeats = 0;
-    /* Added by rtc  */
-    logs[last].flags = 0;
-  }
+  logs = nrealloc(logs, max_logs * sizeof(log_t));
+  memset(logs + last, 0, (max_logs - last) * sizeof(log_t));
+  last = max_logs;
 }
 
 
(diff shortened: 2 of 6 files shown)


src/misc.c
```diff
@@ -95,20 +97,9 @@ void init_misc()
...
-    logs[last].mask = 0;
-    logs[last].f = NULL;
-    /* Added by cybah  */
-    logs[last].szlast[0] = 0;
-    logs[last].repeats = 0;
-    /* Added by rtc  */
-    logs[last].flags = 0;
-  }
+  logs = nrealloc(logs, max_logs * sizeof(log_t));
+  memset(logs + last, 0, (max_logs - last) * sizeof(log_t));
```
thommey, line 101: If the user reduces max_logs and rehashes, `max_logs - last` is negative and you don't handle that, the loop before did handle it (by not executing)

src/chanprog.c
```diff
@@ -312,6 +312,14 @@ void tell_verbose_status(int idx)
                "Threaded DNS core is disabled.\n"
 #endif
                "Socket table: %d/%d\n", threaddata()->MAXSOCKS, max_socks);
+  int j = 0;
+  int k = max_logs * sizeof(log_t);
```
thommey: size_t k

thommey (commented):
szlast is not free'd properly when the log entry is removed:
src/tclmisc.c:101-108 (logfile mask is set to 0)
src/chanprog.c:403-419 expiring log entries free stuff but not szlast

src/misc.c
michaelortmann: reducing max-logs is forbidden already:
```
.set max-logs 6
Error: can't set "max-logs": ERROR: Decreasing max-logs requires a restart
```

src/chanprog.c
michaelortmann: done

michaelortmann:
> szlast is not free'd properly when the log entry is removed: src/tclmisc.c:101-108 (logfile mask is set to 0) src/chanprog.c:403-419 expiring log entries free stuff but not szlast

done

ready for review again

thommey:
AI review:

1. **Off-by-one in grow sizing (real bug)** — `src/misc.c:601-611`. The new code computes `l = strlen(out + tsl)` and allocates `szlast_len = MIN(MAX(l, 128), LOGLINELEN)` then `strlcpy(szlast, out+tsl, szlast_len)`. `strlcpy(dst, src, n)` writes at most `n-1` chars + NUL. So a 200-char message with `szlast_len=200` is truncated to 199 chars. On the next call, `strcasecmp("200-char msg", "199-char msg")` fails — **repeat detection breaks at the growth boundary**. Fix: size on `l + 1`, e.g. `MIN(MAX(l + 1, 128), LOGLINELEN)` in both branches. (Old code allocated `LOGLINELEN = LOGLINEMAX + 1 = 9001` and used `LOGLINEMAX` as strlcpy size, so the original reserved room for the NUL.)

2. **`strlcpy` with NULL dst when `l == 0`** — if `szlast_len == 0` and `l == 0`, neither growth branch runs and `strlcpy(NULL, "", 0)` is invoked. Behavior with `size == 0` is "don't write," but passing NULL to `strlcpy` is dicey across libc variants. Likely unreachable (putlog messages are non-empty once the newline is appended), but worth either guarding with `if (logs[i].szlast)` before the strlcpy or growing unconditionally on the first write.

michaelortmann:
Fixed the off by one error. the 2nd note of the review bot is ridiculous, but its fixed also.

> AI review:
> 
>     1. **Off-by-one in grow sizing (real bug)** — `src/misc.c:601-611`. The new code computes `l = strlen(out + tsl)` and allocates `szlast_len = MIN(MAX(l, 128), LOGLINELEN)` then `strlcpy(szlast, out+tsl, szlast_len)`. `strlcpy(dst, src, n)` writes at most `n-1` chars + NUL. So a 200-char message with `szlast_len=200` is truncated to 199 chars. On the next call, `strcasecmp("200-char msg", "199-char msg")` fails — **repeat detection breaks at the growth boundary**. Fix: size on `l + 1`, e.g. `MIN(MAX(l + 1, 128), LOGLINELEN)` in both branches. (Old code allocated `LOGLINELEN = LOGLINEMAX + 1 = 9001` and used `LOGLINEMAX` as strlcpy size, so the original reserved room for the NUL.)
> 
>     2. **`strlcpy` with NULL dst when `l == 0`** — if `szlast_len == 0` and `l == 0`, neither growth branch runs and `strlcpy(NULL, "", 0)` is invoked. Behavior with `size == 0` is "don't write," but passing NULL to `strlcpy` is dicey across libc variants. Likely unreachable (putlog messages are non-empty once the newline is appended), but worth either guarding with `if (logs[i].szlast)` before the strlcpy or growing unconditionally on the first write.

Fixed the off by one error. the 2nd note of the review bot is ridiculous, but its fixed also.

thommey (approved):
LGTM

PR #1785, opened 2025-02-21
Fix debug mem accounting for tcl.c

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Fix debug mem accounting for tcl.c

Regression since eggdrop 1.9.0rc1

make debug
start eggdrop
.debug
```
File 'tcl.c     ' accounted for 8432/8472 (debug follows:)
```

### DIFF
--- a/src/tcl.c
+++ b/src/tcl.c
@@ -358,6 +358,7 @@ void add_cd_tcl_cmds(cd_tcl_cmd *table)
   while (table->name) {
     if (table->cdata) {
       info = nmalloc(sizeof *info);
+      strtot += sizeof(struct tcl_call_stringinfo);
       info->proc = table->callback;
       info->cd = table->cdata;
       Tcl_CreateObjCommand(interp, table->name, tcl_call_stringproc_cd, info, tcl_cleanup_stringinfo);


michaelortmann:
This PR fixes 1 memory account glitch

There is one more here:
https://github.com/eggheads/eggdrop/pull/438/files#diff-2d8e3362ec4e40b7a6e562a1c7968e7b67063d05f639d9da1e0fa32b414c8706R319

@thommey: Please can you take a look? You know that code better than me.

michaelortmann:
we do not leak here. this PR is merely about our own debug memory accounting.

thommey:
The nrealloc case is tricky to track as function-static memory buffer that only grows, maybe we can circumvent the memory account for it? it's never free'd and only grows on purpose and can't leak

thommey (approved):

michaelortmann:
did you suggest using malloc() instead of our own nmalloc()?
sure it would be efficient, but i have a hard time imagining we start mixing nmalloc() with malloc() and other allocation functions in code. looks more like a general design question.
that said, i am no fan of nmalloc, and because it makes the usual debugging hard, i would just assert and hard crash on memory leaks.
so instead of "random" mixing alloc funcs, i want a 100% decision which of those functions to use in eggdrop and right now that means nmalloc, doesnt it?

thommey (approved):
LGTM

PR #1786, opened 2025-02-23
remove colon from accounts, modes

vanosg:
Found by: @Robby- 
Patch by: Geo

One-line summary:
Remove colons from trailing IRC command

Inspircd v3 started putting :'s on all trailing commands, even if they cannot have a space in them (accounts, nicks, etc). This broke how Eggdrop stored account names sent from WHOX, and nicknames contained in MODE commands.

This is a blunt-force and inelegant patch but it at least fixes the identified issues... there are probably more messages that are needed to be handled but not known at this time, and maybe a larger code revamp could be done. Just not today.

https://docs.inspircd.org/faq/#why-does-my-client-not-show-mode-changesopped-users-on-joinetc-correctly-when-using-inspircd-v3

### DIFF
--- a/src/mod/irc.mod/chan.c
+++ b/src/mod/irc.mod/chan.c
@@ -1205,6 +1205,7 @@ static int got354(char *from, char *msg)
         }
         flags = newsplit(&msg);     /* Grab the flags */
         account = newsplit(&msg);   /* Grab the account name */
+        fixcolon(account);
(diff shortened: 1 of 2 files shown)


thommey:
Fine, but please add // TODO comments because we don't want to leave it like this long-term.

I disagree with all of this and think inspircd should be fixed, not eggdrop.

PR #1795, opened 2025-03-14
Patch 2

MeNearly:
Found by: MeNearly (aka ian)
Patch by: MeNearly
Fixes: fix missing "Writing channel file" translations

One-line summary: The recurrent message "Writing channel file..." was not translated, so, following the logic I determined reading the rest of code, I did the translations (w/ help of an LLM ^^).

```
.tcl putlog $::version
[16:53:06] 1.10.0 1100004
Tcl:
[17:00:00] Écriture de la liste des utilisateurs...
[17:00:00] Écriture de la liste des canaux...
```


michaelortmann:
The new CHAN_FILE_WRITING looks good.

But CHAN_FILE_COPYING is only defined and not used in code. So you could remove it. Or you could fix the code to use it. Looks to me as if someone forgot to remote the string from the italian language file some long time ago.

For the german file, can you please change Kanalliste to Channeldatei?

and merci for your 2nd PR! :)

MeNearly:
* change Kanalliste to Channeldatei => OK, I'll do that
* someone forgot to remote the string from the italian language => I suspected this to
Unfortunately, I did not find any relevant mention of "copy" anywhere in the source code. And further, I can't remember of any partyline feature that copy the channel file...
I suggest to remove it. Do you agree ?

[Edit : done]

michaelortmann:
fyi: here the italian language file was added for eggdrop 1.9.1:
https://github.com/eggheads/eggdrop/commit/322d3568a3f963fb9ba91a15868fee0a4dcd6e12

the numbering fix also came to my mind,
but we used 0x90d in the past.
so the question is, if we want to avoid strings from older / non updated installation language files to show up as the new 0x90d.
like:
/home/michael/usr/src/eggdrop1.4.0/src/lang.h:#define CHAN_BOGUSBAN	get_language(0x90d)
@Geo: is repurposing 0x90d fine for you?

MeNearly:
@michaelortmann 
> so the question is, if we want to avoid strings from older / non updated installation language files to show up as the new 0x90d.

If I understand well, you're suggesting some could only update the src folder alone ?

michaelortmann:
i mean, someone could update their eggdrop, like with make install, but not update language files, keep the old language files in their destination installation directory.

michaelortmann:
that said, for me, i would be ok with assuming, new language files are installed as well, ans thus fix numbering, reusing old numbers. but i cant speak for anyone else or eggheads. so i asked geo to answer that general question.

MeNearly:
@michaelortmann OK, I didn't get it [Usually I do a `git pull` ^^]
[Edit] OK, I'll wait

michaelortmann:
i am also no native english speaker (and no french and no italian), so we fight a bit with language barrier here probably :))

MeNearly:
I suggest to watch some "RobWords" on YT: Rob is a native english man, living in Germany ;op

MeNearly:
@vanosg What do you think ?

skralg (approved):
Looks good to me

vanosg:
This one is really a toss-up... on the one hand, something that has been gone since 1.4 isn't really a concern for us anymore. However, to inspire good coding practices in the future and help people not have to review old source to see if a value is allowed or not, I think we'll comment out 0x90d and move this to the end- unless I'm misunderstanding what was presented in the discussion of this PR

MeNearly:
@vanosg Hi.
The discussion was about the actual core.italian.lang file which still have : 
`0x911,Scrivo il file dei canali...
0x915,Faccio una copia del file dei canali...`
And it's the only lang file.
And we talked about my suggestion of numbering them as 0x90d [and 0x90e ?], while restoring the translations of "Writing channel file...".

vanosg:
Ah, thanks for the clarification. So yes- lets not use 0x90d, let's just use 0x90e.  And we'll just remove those two random Italian strings. Good catch, thanks.

MeNearly:
@vanosg Well, it's up to you 🙂
— ian

vanosg:
Sorry, one more request- can you add 
```
//0x90d - removed 
```
in the correct place in each file? Thanks so much!

MeNearly:
@vanosg Is it ok ?

vanosg:
Please pull the most current version of the develop branch, run ``git merge <insert eggheads remote name here>/develop`` in this branch on your local machine, and then push back to this PR. Thanks!

MeNearly:
checks seem to be stuck :/

PR #1798, opened 2025-03-23
Require owner perms for python

vanosg:
Found by: thommey
Patch by: Geo

One-line summary:
Require +n flag to use .python command

Also checks the `must-be-owner` config setting to see if all owners or only perm owners can use `.python`

set must-be-owner 1
```
You have no messages.
*** Geo joined the party line.
.tcl
What?  You need '.help'
.python
What?  You need '.help'
```
set must-be-owner 0
```
.tcl
Tcl: 
.python
Python Error: invalid syntax (<string>, line 0)
```

### DIFF
--- a/src/mod/python.mod/pycmds.c
+++ b/src/mod/python.mod/pycmds.c
@@ -65,7 +65,7 @@ static void cmd_python(struct userrec *u, int idx, char *par) {
   Py_ssize_t n;
   int i;
 
-  if (!(isowner(dcc[idx].nick)) && (must_be_owner)) {
+  if (!isowner(dcc[idx].nick) && must_be_owner) {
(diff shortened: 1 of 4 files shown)


src/mod/python.mod/pycmds.c
```diff
@@ -65,6 +65,11 @@ static void cmd_python(struct userrec *u, int idx, char *par) {
   Py_ssize_t n;
   int i;
 
+  if (!(isowner(dcc[idx].nick)) && (must_be_owner)) {
```
thommey:
```suggestion
  if (!isowner(dcc[idx].nick) && must_be_owner) {
```

then it's fine

michaelortmann: can we automate this? let the CI check for superflous brackets?

PR #1802, opened 2025-03-25
Update m4/ax_check_compile_flag.m4 from upstream

michaelortmann:
Found by: michaelortmann
Patch by: michaelortmann
Fixes: 

One-line summary:
Update `m4/ax_check_compile_flag.m4` from upstream
Also update `src/mod/python.mod/configure.ac:AC_PREREQ`


vanosg:
@thommey this jumps the autoconf requirement ahead by a decade... I don't think it's an issue but do you have any opinions before I merge?

Issue #1831, opened 2025-08-26
time zone versus settings (scientific grammar/terminology)


dchmelik:
The eggdrop.conf file incorrectly calls a time setting (EST) a zone (ET).  Time zones (such as Eastern Time (ET)) are locations that have time settings (Eastern Standard Time (EST) & Eastern Daylight Time (EDT) the two there, etc., with another example in your configuration being related to what's perhaps Central European Summer Time, i.e., what USA calls daylight saving time, Europe calls Summer time).

vanosg:
Closed in [ab433b4](https://github.com/eggheads/eggdrop/commit/ab433b49ff7651093e6e261ef229da5c3b9628db)

PR #1837, opened 2025-10-02
Fix and update timezone default and doc

### LINKED ISSUE #461: Make timezone setting easier
Right now we use $offset and $timezone and set that into the TZ env variable instead of just letting users set env(TZ) to Europe/Berlin and similar. I don't know if $offset and $timezone are used in scripts. if they are, we could figure them out from the env(TZ) setting.
I see little reason to allow people to set offset/timezone only but not the shell environment timezone that affects everything eggdrop does.

michaelortmann:
Found by: https://github.com/thommey and https://github.com/dchmelik
Patch by: https://github.com/michaelortmann
Fixes: #461 and #1831

One-line summary:
Fix and update timezone default and doc

**Test 1**
myeggdrop.conf:
```
#set timezone "America/New_York"
#set env(TZ) "$timezone"
```
`Local time is now 02:53`
**Test 2**
myeggdrop.conf:
```
set timezone "America/New_York"
set env(TZ) "$timezone"
```
`Local time is now 20:55`


michaelortmann:
@thommey: This fix is similar to what you ask in issue #461, but not quite the very same. I aim to fix both Issues with this PR. `$timezone` is used in scripts. In the ones we ship by default:
https://github.com/eggheads/eggdrop/blob/d317ac228eb6682fdd6386007e09d627efd618ca/scripts/ques5.tcl#L366-L367

I was thinking about switching the timezone and env settings in eggdrop.conf:
instead of:
```
#set timezone "Europe/Berlin"
#set env(TZ) "$timezone"
```
we could:
```
#set env(TZ) "Europe/Berlin"
# Set variable timezone for tcl scripts. This variable is deprecated and will be removed in a future release.
#set timezone $env(TZ)
```
But then i realized, if users only want to set the timezone variable and not the env(TZ) this make sthings more complicated for them.
So my new proposal is:
```
# Variable timezone for tcl scripts is deprecated and will be removed in a future release. So in a future release set env(TZ) "Europe/Berlin" directly instead.
#set timezone "Europe/Berlin"
#set env(TZ) "$timezone"
```
And when we later remove variable timezone, we patch ques5.tcl to use env(TZ) directly.
I hope someone comes up with a better doc wording here.

vanosg:
Playing with the pre-PR version, I just noticed this interesting timestamp gem:
```
[13:50:32] TLS: handshake successful. Secure connection established.
[13:50:32] Connected to testnet.foo.chat
[18:50:33] CAP: Current negotiations with testnet.foo.chat: message-tags sasl userhost-in-names
```

michaelortmann:
@vanosg this bug also exists in branch develop, its not a PR bug. anyway. took me a while to find and fix this bug. but here we are :) ready for review again!

vanosg:
I understand the terminology basis for changing the value of the timezone config setting, but how will this affect old non-updated configs being used after this change is made?

michaelortmann:
if an old config is used, it will work as before 
For example the following config will work as before:
```
set timezone "EST"
set offset "5"
set env(TZ) "$timezone$offset"
```

eggdrop.conf
```diff
@@ -32,23 +32,17 @@ set admin "Lamer <email: lamer@lamest.lame.org>"
...
-# positive if the local timezone is west of the Prime Meridian and
-# negative if it is east. The value (in hours) must be between -23 and
-# 23. For example, if the timezone is UTC+1, the offset is -1.
-set offset "5"
+# This setting defines which timezone your bot is in. It's used for internal
+# routines as well as for logfile timestamping and scripting purposes. The
+# timezone string specifies an abbreviation for the timezone and must be three
+# or more alphabetic characters. For example, timezone "Europe/Berlin"
+# automatically adjusts for standard time and daylight saving time. See
+# https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#Time_zone_abbreviations.
```
thommey:
```suggestion
# https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List
# and use the full timezone identifiers
```

thommey (commented):
@vanosg this removes the offset Tcl variable - do you think that is an issue?

Otherwise I much prefer this approach with timezone names instead of manually figuring out offets unnecessarily.

doc/sphinx_source/using/core.rst
```diff
@@ -38,21 +38,15 @@ You can change the basic Eggdrop appearance and behavior in this section.
...
-  set env(TZ) "$timezone $offset" (disabled by default)
+  set timezone "America/New_York"
+    This setting defines which timezone your bot is in. It's used for internal
+    routines as well as for logfile timestamping and scripting purposes. The
+    timezone string specifies an abbreviation for the timezone and must be three
+    or more alphabetic characters. For example, timezone "Europe/Berlin"
+    automatically adjusts for standard time and daylight saving time. See
+    https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#Time_zone_abbreviations.
+
+  set env(TZ) "$timezone" (disabled by default)
```
vanosg, line 49: Is "disabled by default" still relevant here?

vanosg:
> @vanosg this removes the offset Tcl variable - do you think that is an issue?
> 
> Otherwise I much prefer this approach with timezone names instead of manually figuring out offets unnecessarily.

I have always hated how Eggdrop handles timezones- offset was added for scripting only. Taking it out could cause issues? Maybe we just hardcode it to 0 in the codebase so nothing crashes, and implement this one as part of 1.11.0?

doc/sphinx_source/using/core.rst
michaelortmann: yes, eggdrop conf will still have this commented out, thus: disabled by default:

`#set env(TZ) "$timezone"`

michaelortmann:
if tcl scripts use "offset" they would change behaviour, but they should not crash because offset was always just a tcl variable set in eggdrop.conf so scripts could never depend on its existence. to not break behaviour of tcls, we could (maybe should?) calculate and set the tcl variable offset based on the set timezone. this would be nice, wouldnt it?

michaelortmann:
also, static offsets (like we currently have in eggdrop.conf) are not a worse alternative to timezone names, they are not equivalent/static. many timezones have different offsets throughout the year. no one should be damned to update offset manually to stay synced to their timezone.

due to changing offsets, we can argue that the idea of using offset in tcls was borked and never really worked all time anyway. therefore removing offset alltogether is justified. what do you say? calculate and set offset whenever timezone is set/changed or remove it?

vanosg:
@thommey  thoughts?

src/chanprog.c
```diff
@@ -39,7 +39,7 @@ extern char ver[], botnetnick[], firewall[], motdfile[], userfile[], helpdir[],
 extern time_t now, online_since;
 extern int backgrd, term_z, con_chan, cache_hit, cache_miss, firewallport,
            default_flags, max_logs, conmask, protect_readonly, make_userfile,
-           noshare, ignore_time, max_socks;
+           noshare, ignore_time, max_socks, now2_last;
```
thommey: now2_last should be time_t

doc/sphinx_source/using/core.rst
```diff
@@ -38,21 +38,15 @@ You can change the basic Eggdrop appearance and behavior in this section.
...
-    time to get Coordinated Universal Time (UTC aka GMT). The offset is
-    positive if the local timezone is west of the Prime Meridian and
-    negative if it is east. The value (in hours) must be between -23 and
-    23. For example, if the timezone is UTC+1, the offset is -1.
-
-  set env(TZ) "$timezone $offset" (disabled by default)
+  set timezone "America/New_York"
+    This setting defines which timezone your bot is in. It's used for internal
+    routines as well as for logfile timestamping and scripting purposes. The
+    timezone string specifies an abbreviation for the timezone and must be three
```
thommey: this isn't accurate anymore, it is not restricted to 3 letters

eggdrop.conf
```diff
@@ -32,23 +32,18 @@ set admin "Lamer <email: lamer@lamest.lame.org>"
...
-
-# The offset setting specifies the time value to be added to the local
-# time to get Coordinated Universal Time (UTC aka GMT). The offset is
-# positive if the local timezone is west of the Prime Meridian and
-# negative if it is east. The value (in hours) must be between -23 and
-# 23. For example, if the timezone is UTC+1, the offset is -1.
-set offset "5"
+# This setting defines which timezone your bot is in. It's used for internal
+# routines as well as for logfile timestamping and scripting purposes. The
+# timezone string specifies an abbreviation for the timezone and must be three
```
thommey: same as docs, not an abbreviation anymore

eggdrop.conf
```diff
@@ -32,23 +32,18 @@ set admin "Lamer <email: lamer@lamest.lame.org>"
...
+# or more alphabetic characters. For example, timezone "Europe/Berlin"
+# automatically adjusts for standard time and daylight saving time. See
+# https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List
```
thommey: Add the list to the rst docs as well, not just eggdrop.conf :)

thommey (changes requested):
almost done, very nice change!

doc/sphinx_source/using/core.rst
michaelortmann: line 45 says:
https://github.com/eggheads/eggdrop/pull/1837/changes#diff-a215cc92885bbd01347d9fbdc923b6c208fc5bc1a33d71325b10d2f5dd65758cR45
so, the doc is correct, it must be **three or more**

eggdrop.conf
michaelortmann: both files eggdrop.conf and core.rst have now same doc text.

eggdrop.conf
michaelortmann: fixed.

doc/sphinx_source/using/core.rst
michaelortmann: fixed

src/chanprog.c
michaelortmann: good catch.

PR #1838, opened 2025-10-03
Fix tdns leak and init attr only once

michaelortmann:
Found by: https://github.com/michaelortmann
Patch by: https://github.com/michaelortmann
Fixes: 

One-line summary:
Fix tdns leak and init attr only once

Also cleaned up code structure, moved tdns init code from `main.c` to `dns.c:init_tdns()`

before:
```
$ valgrind --leak-check=full ./eggdrop -t BotA.conf 2>&1|grep dns.c
.tcl dnslookup google.com foo
.die
==11075==    by 0x24E202: core_dns_ipbyhost (dns.c:620)
==11075==    by 0x24E973: tcl_dnsipbyhost (dns.c:369)
==11075==    by 0x24E640: tcl_dnslookup (dns.c:739)
```
https://github.com/eggheads/eggdrop/blob/d317ac228eb6682fdd6386007e09d627efd618ca/src/dns.c#L620
after:
```
$ valgrind --leak-check=full ./eggdrop -t BotA.conf 2>&1|grep dns.c
.tcl dnslookup google.com foo
.die
```

### DIFF
--- a/src/dns.c
+++ b/src/dns.c
@@ -630,20 +630,19 @@ void core_dns_ipbyhost(char *host)
     nfree(dtn);
     return;
   }
-  dtn->next = dns_thread_head->next;
-  dns_thread_head->next = dtn;
   strlcpy(dtn->host, host, sizeof dtn->host);
   if (pthread_create(&(dtn->thread_id), &attr, thread_dns_ipbyhost, (void *) dtn)) {
     putlog(LOG_MISC, "*", "core_dns_ipbyhost(): pthread_create(): error = %s", strerror(errno));
     call_ipbyhost(host, &addr, 0);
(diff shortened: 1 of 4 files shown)


thommey (changes requested):
The `pthread_mutex_destroy()` you add to `core_dns_ipbyhost()` should probably also be added to `core_dns_hostbyip()`, right?

michaelortmann:
> The `pthread_mutex_destroy()` you add to `core_dns_ipbyhost()` should probably also be added to `core_dns_hostbyip()`, right?

yes. so i added them.

ready for review again.

thommey (approved):
LGTM

PR #1846, opened 2025-10-22
Crash fix heap-use-after-free

michaelortmann:
Fixes: 

One-line summary:
Fix a pointer in dcc table linking into dcc table. We must not do that because dcc table can be realloced

Between dns resolve start and finish, realloc can happen here:
https://github.com/eggheads/eggdrop/blob/d317ac228eb6682fdd6386007e09d627efd618ca/src/dccutil.c#L89
like when someone connects to the eggdrrop at the right moment
Its a heap-use-after-free
When investigating the code i realized the dns_info.ip pointer is redundant, and the fix equals a cleanup.

```
==444524==ERROR: AddressSanitizer: heap-use-after-free on address 0x7d0d8c608f58 at pc 0x559353b0b799 bp 0x7ffc6235d370 sp 0x7ffc6235d360
READ of size 4 at 0x7d0d8c608f58 thread T0
    #0 0x559353b0b798 in dns_dcchostbyip /home/michael/projects/eggdrop/src/dns.c:153
[...]
0x7d0d8c608f58 is located 3672 bytes inside of 4560-byte region [0x7d0d8c608100,0x7d0d8c6092d0)
freed by thread T0 here:
    #0 0x7efd8ef1fa45 in realloc /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:81
    #1 0x559353b3a356 in n_realloc /home/michael/projects/eggdrop/src/mem.c:388
    #2 0x559353afdde4 in increase_socks_max /home/michael/projects/eggdrop/src/dccutil.c:89
[...]
==444524==ABORTING
```

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -1283,7 +1283,6 @@ static void dcc_telnet(int idx, char *buf, int i)
     return;
   }
 
-  dcc[i].u.dns->ip = &dcc[i].sockname;
   dcc[i].sock = sock;
   dcc[i].port = port;
 #ifdef TLS
--- a/src/dns.c
+++ b/src/dns.c
@@ -146,11 +146,11 @@ static void dns_dcchostbyip(sockname_t *ip, char *hostn, int ok, void *other)
         (dcc[idx].u.dns->dns_type == RES_HOSTBYIP) && (
 #ifdef IPV6
         (ip->family == AF_INET6 &&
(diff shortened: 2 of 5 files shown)


thommey (approved):
good fix, deleting code is great :)
LGTM

PR #1848, opened 2025-10-29
Add eggdrop version check python script

michaelortmann:
Fixes: 

One-line summary:
Add eggdrop version check python script

The script version checks every 24h
The other example scripts never used `eggdrop.tcl.set()` and `threading`, so this is a nice addition

eggdrop.conf:
```
loadmodule python
pysource scripts/check_version.py
```
```
> ./eggdrop -t eggdrop.conf
[...]     
Module loaded: python          
Version check: eggdrop update found: latest v1.10.1 current v1.10.0
Loaded check_version.py
```

### DIFF
--- a/src/mod/python.mod/scripts/check_version.py
+++ b/src/mod/python.mod/scripts/check_version.py
@@ -24,7 +24,8 @@ def check_version():
             putlog(
                 f"Version check: eggdrop update found: latest {latest_version} current {current_version}"
             )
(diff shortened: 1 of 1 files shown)


vanosg:
I don't think this is something that we need to include, but would be a great addition for an external repo to house!

michaelortmann:
moved to https://codeberg.org/mortmann/snippets/src/branch/main/check_version.py

PR #1859, opened 2025-11-21
Enhance tdns node mutex lock

michaelortmann:
Found by: github.com/michaelortmann/
Patch by: github.com/michaelortmann/
Fixes: 

One-line summary:

Quick summary of what is happening now:
- Each tdns thread has its own data node in a linked list.
- Each tdns thread finishes by writing its data to its node, using a mutex lock on that node, then calling close() on a pipe to signal eggdrops main thread select() that data is ready.
- In eggdrops mainloop net.c (the changed code of this PR) we loop over the linked list, check which node(s) fds were closed and handle those nodes.

By putting the debug print of the node->strerror inside the FD_ISSET code-block,  we only need to lock the relevant nodes.

tested for regression with:
```
valgrind --tool=helgrind ./eggdrop -nt BotA.conf
[...]
.tcl dnslookup google.de foo
```

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -1060,11 +1060,11 @@ int sockread(char *s, int *len, sock_list *slist, int slistmax, int tclonly)
 #ifdef EGG_TDNS
   dtn_prev = dns_thread_head;
   for (dtn = dtn_prev->next; dtn; dtn = dtn->next) {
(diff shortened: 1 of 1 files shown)


thommey (approved):
LGTM

PR #1872, opened 2025-12-31
Arbitrary mode support, phase 1

thommey:
Found by: various
Patch by: thommey

One-line summary: Understand mode info from raw 005 isupport and parse modes correctly

Phase 1 is parsing mode info from ISUPPORT raw 005 and using it to properly split modes (primarily for bind mode / pushmode) in gotmode() and got324().
It is not intended to overhaul eggdrop's mode enforcement or understanding of/reaction to modes.
The goal is to introduce proper parsing without breaking anything first.

**FULL FF-MERGE, NO SQUASHING**

Commit history is clean, review individual commits instead of full diff.

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -31,6 +31,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include "chan.h"
+#include "src/eggdrop.h"
 #include "tandem.h"
 #include "modules.h"
 
@@ -267,6 +268,60 @@ char *newsplit(char **rest)
     *o++ = 0;
   *rest = o;
   return r;
+}
+
+// WARNING: modifies original text
+// Split IRC text into words (without the "from" component, which could also start with ':')
+// - replace all ' ' with \0, splitting into words
+// - if a word starts with ':' it is the last word, it can contain spaces
+// - return argc/argv structure (pointers into original text)
+struct parsed_irc parse_irc(char *text)
+{
+  struct parsed_irc result = {.argc = 0};
+
+  while (*text) {
+    while (*text == ' ') {
+      *text++ = '\0';
+    }
+    if (!*text) {
+      break;
+    }
+    if (result.argc == MAX_IRC_TOKENS - 1) {
+      putlog(LOG_MISC, "*", "parse_irc() error: too many tokens, PLEASE REPORT THIS BUG");
+      result.argv[result.argc++] = text;
+      break;
+    } else if (*text == ':') {
+      *text++ = '\0';
+      result.argv[result.argc++] = text;
+      break;
+    } else {
+      result.argv[result.argc++] = text;
+      while (*text && *text != ' ') {
+        text++;
+      }
+    }
+  }
+
+  return result;
+}
+
+char *join_str_array(char **argv, int argc, char *delim, char *outbuf, size_t outbufsiz)
+{
+  size_t written = 0;
+
+  if (!argc) {
+    outbuf[0] = '\0';
+    return outbuf;
+  }
+
+  for (int i = 0; i < argc; i++) {
+    written += snprintf(outbuf + written, outbufsiz - written, "%s%s", argv[i], i == argc - 1 ? "" : delim);
+    if (written >= outbufsiz) {
+      written = outbufsiz - 1;
+      break;
+    }
+  }
+  return outbuf;
 }
 
 /* maskhost(), modified to support custom mask types, as defined
--- a/src/mod/irc.mod/chan.c
+++ b/src/mod/irc.mod/chan.c
@@ -2860,6 +2860,11 @@ static void update_chanmodes(mode_info_t *modes, int is_prefix)
       }
     }
   }
+  if (!is_prefix) {
+    // assume that if +e/+I are list-type modes that they are exempts and invites
+    use_exempts = (MODE_TYPE('e') == MODETYPE_LIST);
+    use_invites = (MODE_TYPE('I') == MODETYPE_LIST);
+  }
 }
 
 // CHANMODES=eIbq,k,flj,CFLMPQScgimnprstuz
--- a/src/mod/irc.mod/chan.c
+++ b/src/mod/irc.mod/chan.c
@@ -958,16 +958,21 @@ static void recheck_channel(struct chanset_t *chan, int dobans)
 }
 
 /* got 324: mode status
- * <server> 324 <to> <channel> <mode>
+ * <server> 324 <to> <channel> <mode> [<mode params>...]
  */
-static int got324(char *from, char *msg)
+static int got324(char *from, char *origmsg)
 {
-  int i = 1, ok = 0;
-  char *p, *q, *chname;
+  int i = 1, ok = 0, nextarg = 3;
+  char *chname, *chg, *arg, buf[511];
+  struct parsed_irc msg;
   struct chanset_t *chan;
 
-  newsplit(&msg);
-  chname = newsplit(&msg);
+  strlcpy(buf, origmsg, sizeof buf);
+  msg = parse_irc(buf);
+  if (msg.argc < 3)
+    return 0;
+  chname = msg.argv[1];
+  chg = msg.argv[2];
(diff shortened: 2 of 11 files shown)


src/misc.c
```diff
@@ -278,6 +278,52 @@ char *newsplit(char **rest)
...
+      while (*text && *text != ' ') {
+        text++;
+      }
+    }
+  }
+
+  return result;
+}
+
+char *join_str_array(char **argv, int argc, char *delim)
```
michaelortmann: can we have last parameter type char or is there a need for delimiters longer than 1 char?

src/proto.h
```diff
@@ -236,6 +236,8 @@ void check_logsize(void);
 void splitc(char *, char *, char);
 void splitcn(char *, char *, char, size_t);
 void remove_crlf(char **);
+parsed_irc_t parse_irc(char *);
+char *join_str_array(char **argv, int argc, char *delim);
```
michaelortmann: remove variable names?

src/misc.c
michaelortmann: on the other hand, i can imagine things delimiters like ", ". but i think for generic misc functions exported via module API, we should make the output buffer a parameter and len, not a static array. could be useful for threaded code or for code that needs to join into a buffer longer than 512 chars.

thommey: I think keeping it open for multi-char delimiters is more flexible and doesn't make it any more difficult to use.
Caller-provided buffer with length limitation makes sense, I'll change that.

src/proto.h
thommey: done

src/misc.c
thommey: done

vanosg:
https://ircv3.net/specs/extensions/multi-prefix   might make sense to think about at the same time as you build out some of these data structures

thommey:
> https://ircv3.net/specs/extensions/multi-prefix might make sense to think about at the same time as you build out some of these data structures

good point but a separate thing, out of scope for this PR :)

src/mod/irc.mod/chan.c
```diff
@@ -2865,6 +2968,14 @@ static int irc_isupport(char *key, char *isset_str, char *value)
     }
   } else if (!strcmp(key, "BOT")) {
     botflag005 = value[0];
+  } else if (!strcmp(key, "CHANMODES")) {
+    if (process_chanmodes(isset ? value : "beI,k,l,imnpstcCMRrDuNdTaq")) {
```
vanosg: is it worth using the default isupport values here, rather than hardcoding?

src/mod/irc.mod/chan.c
```diff
@@ -2865,6 +2968,14 @@ static int irc_isupport(char *key, char *isset_str, char *value)
...
+      putlog(LOG_MISC, "*", "Error: isupport unable to parse CHANMODES=%s", isset ? value : "(unset)");
+    }
+  } else if (!strcmp(key, "PREFIX")) {
+    if (process_prefix(isset ? value : "(ohv)@%+")) {
```
vanosg: is it worth using the default isupport values here, rather than hardcoding?

src/mod/irc.mod/irc.h
```diff
@@ -33,6 +33,30 @@
 #define REVENGE_KICK 1          /* Kicked victim        */
 #define REVENGE_DEOP 2          /* Took op              */
 
+/* Order and values are relied upon, do not change */
```
vanosg, line 36: Can we add a comment that 1-4 are CHAMODES and 5 is PREFIX, for clarity

vanosg (commented):
This has been another useless vanosg code review. Thank you for your time.

vanosg:
Limited manual testing:
```
[23:52:34] [@] :iridium.libera.chat 005 BeerBot CHANMODES=eIbq,k,flj,CFLMPQRSTcgimnprstuz CHANLIMIT=#:250 PREFIX=(ov)@+ MAXLIST=bqeI:100 MODES=4 NETWORK=Libera.Chat STATUSMSG=@+ CASEMAPPING=rfc1459 NICKLEN=16 MAXNICKLEN=16 CHANNELLEN=50 TOPICLEN=390 :are supported by this server

.tcl isupport isset CHANMODES
Tcl: 1

.tcl isupport get CHANMODES
Tcl: eIbq,k,flj,CFLMPQRSTcgimnprstuz

.tcl set use-exempts
Tcl: 1
.tcl set use-invites
Tcl: 1
```

```
[00:14:54] [@] :dallas.tx.us.undernet.org 005 BeerBot MAXNICKLEN=15 TOPICLEN=160 AWAYLEN=160 KICKLEN=160 CHANNELLEN=200 MAXCHANNELLEN=200 CHANTYPES=#& PREFIX=(ov)@+ STATUSMSG=@+ CHANMODES=b,k,l,imnpstrDdRcCPM CASEMAPPING=rfc1459 NETWORK=UnderNet :are supported by this server

.tcl isupport get CHANMODES
Tcl: b,k,l,imnpstrDdRcCPM

.tcl set use-exempts
Tcl: 0
.tcl set use-invites
Tcl: 0
```

And what I assume is still TODO for the next phase:
```
[00:14:54] [@] :dallas.tx.us.undernet.org 005 BeerBot MAXNICKLEN=15 TOPICLEN=160 AWAYLEN=160 KICKLEN=160 CHANNELLEN=200 MAXCHANNELLEN=200 CHANTYPES=#& PREFIX=(ov)@+ STATUSMSG=@+ CHANMODES=b,k,l,imnpstrDdRcCPM CASEMAPPING=rfc1459 NETWORK=UnderNet :are supported by this server

.chanset #eggtest chanmode +z
Successfully set modes { chanmode { +z } } on #eggtest.
[00:23:37] #-HQ# chanset #eggtest chanmode { +z }
```

vanosg:
Suggested commit entry:

This commit handles parsing mode info from ISUPPORT raw 005 and using it to properly split modes (primarily for bind mode / pushmode) in gotmode() and got324(). Each collected mode is packed in an argc/argv struct holding pointers into the original text. Each individual flag provided by CHANMODES is stored in a mode_info struct, along with the type of mode it is ( MODETYPE_FLAG (no params), MODETYPE_LIMIT(param on set only), MODETYPE_KEY (param always), MODETYPE_LIST (param always) ).

This commit is not intended to overhaul eggdrop's mode enforcement or understanding of/reaction to modes.
The goal is to introduce proper parsing without breaking anything first.

The sole functional change is using CHANMODES to set use_invites and use_exempts (hence the removal of these config settings.

src/mod/irc.mod/chan.c
thommey: That fallback case only happens on CHANMODES or PREFIX without a value which makes no sense. Changed it so we instead fall back to "" to avoid NULL pointer issues, but which fails parsing later.
Better than hardcoding a seeming default, agreed.

src/mod/irc.mod/chan.c
thommey: see above

src/mod/irc.mod/irc.h
thommey: done

thommey:
> Suggested commit entry:
> 
> This commit handles parsing mode info from ISUPPORT raw 005 and using it to properly split modes (primarily for bind mode / pushmode) in gotmode() and got324(). Each collected mode is packed in an argc/argv struct holding pointers into the original text. Each individual flag provided by CHANMODES is stored in a mode_info struct, along with the type of mode it is ( MODETYPE_FLAG (no params), MODETYPE_LIMIT(param on set only), MODETYPE_KEY (param always), MODETYPE_LIST (param always) ).
> 
> This commit is not intended to overhaul eggdrop's mode enforcement or understanding of/reaction to modes. The goal is to introduce proper parsing without breaking anything first.
> 
> The sole functional change is using CHANMODES to set use_invites and use_exempts (hence the removal of these config settings.

bind mode / pushmode are unaffected by this change, they still only get the modes they previously did.

PR #1877, opened 2026-02-16
Remove amalgamations (irc.mod first)

thommey:
Found by: thommey
Patch by: thommey

One-line summary:
Remove C file amalgamations (huge .c file by #include of other .c files) in modules

This drastically improves IDE support. irc_proto.h is the internal header file for irc.mod interdependencies. The key strategy to minimal build system changes is incremental linking of multiple .o files into a single .o file:
```
../irc.o: irc.o mode.o ...
	$(CC) -r -nostdlib -o ../irc.o irc.o mode.o ...
```

Disclaimer: 90% AI generated (Claude Opus 4.6)

### DIFF
--- a/src/mod/irc.mod/Makefile
+++ b/src/mod/irc.mod/Makefile
@@ -16,7 +16,7 @@ static: ../irc.o
 modules: ../../../irc.$(MOD_EXT)
 
 ../irc.o: $(OBJS) irc.h irc_proto.h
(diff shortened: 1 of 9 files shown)


michaelortmann:
instead of removing the static specifier for functions, we should first check, if the function could be moved to stay static. for example, `check_tcl_pubm()` should simply be moved from `irc.c` to `chan.c` and stay static there.

PR #1880, opened 2026-02-22
Extban support

vanosg:
Patch by: Geo

One-line summary:
Add extban support

This PR does the following:
- Adds the '.+extban' command which adds the ability to insert an extban flag to a ban. This is essentially a wrapper for the .+ban command. This was allegedly done for backwards compatibility, but maybe it's not needed? 
 
- Stores the EXTBAN 005 flags

- Stores the ACCOUNTEXTBAN 005 flag

- Removes the previous banmask "fixing" that was preventing Eggdrop from creating extban-formatted banmasks. These banmaks are now able to be stored in the banlist just like a normal ban.

- Handles extbans in two ways- enforceable, and non-enforceable
- - Enforceable extbans are modes Eggdrop can enforce with a kick based on things like account, realname, etc. These bans are treated like normal bans in the banlist and are subject to all the same rules (ie, dynamic bans, ban-time, etc)
- - - Enforceable bans should be activated in the normal ways- when a matching ban is placed, when a matching user joins, etc
- - - Currently supports only ACCOUNTEXTBAN flag and the U flag. 
- - Non-enforceable bans are modes that Eggdrop can't enforce and are instead closer to server modes than bans (things like blocking a user from sending notices, ctcps, etc). These bans are treated like sticky bans.
- - - Currently supports the following non-enforceable modes obtained from https://defs.ircdocs.horse/defs/extbans.html:  p, q, Q, A, B, c, N, and T

- Any extban flag can be used create and store an internal ban in Eggdrop, but Eggdrop will only activate an extban on channel if the server allows that flag to be created (ie, stored in EXTBAN 005 flag)
 
 
TO DOs:

[ ] ~~Add in more enforceable flags~~- will do in future PR
[X] Add Tcl variables
[X] Switch from enforceable-default to sticky-default
[ ] ~~Add long-form accountextban ~~- will do in future PR


src/mod/channels.mod/userchan.c
```diff
@@ -220,9 +220,55 @@ static int u_equals_mask(maskrec *u, char *mask)
...
+      }
+    }
+  }
+
+  for (; rec; rec = rec->next) {
+    if (extban_parse(rec->mask, &type, &arg)) {
+      me = module_find("server", 0, 0);
+      if (me && me->funcs && me->funcs[SERVER_GET_ISUPPORT]) { 
+        accountflag = isupport_get("ACCOUNTEXTBAN", strlen("ACCOUNTEXTBAN"));
+        return 0;
```
thommey: why return 0 regardless of accountflag?

src/mod/irc.mod/chan.c
```diff
@@ -92,6 +93,56 @@ static void update_idle(char *chname, char *nick)
...
+  if (!m || !m->account[0]) {
+    return 0;
+  }
+
+
+  me = module_find("server", 0, 0);
+  if (me && me->funcs && me->funcs[SERVER_GET_ISUPPORT]) {
+    v = (const char *)isupport_get("ACCOUNTEXTBAN", strlen("ACCOUNTEXTBAN"));
+  }
+  if (type && (type == v[0])) {
```
thommey: v could be NULL

src/mod/irc.mod/chan.c
```diff
@@ -113,6 +166,11 @@ static void setaccount(char *nick, char *account)
           check_tcl_account(m->nick, m->userhost, get_user_from_member(m), chan->dname, account);
         }
         strlcpy(m->account, account, sizeof m->account);
+
+        egg_snprintf(user, sizeof user, "%s!%s", m->nick, m->userhost);
+        if (u_match_mask(global_bans, user) || u_match_mask(chan->bans, user)) {
+          refresh_ban_kick(chan, user, m->nick);
```
thommey, line 203: `refresh_ban_kick` only executes the kick, should this possibly be (maybe in addition) `check_this_ban` to set the ban if it is dynamic?

src/mod/irc.mod/chan.c
```diff
@@ -92,6 +93,56 @@ static void update_idle(char *chname, char *nick)
...
+
+  comma = strchr(value, ',');
+  types = comma ? comma + 1 : value;
+  for (; *types; types++)
+    if (*types == flag)
+      return 1;
+  return 0;
+}
+
+/* XXXXXXX Document this little clusterf */
```
thommey: Document pls :P

src/mod/channels.mod/channels.c
```diff
@@ -55,6 +57,100 @@ static int gfld_chan_thr, gfld_chan_time, gfld_deop_thr, gfld_deop_time,
...
+      // Prefix
+      return;
+    }
+  }
+  // What do we do if no , present?
+  return;
+}
+
+/*
+static int extban_flag_is_supported(char flag)
```
thommey: do we need this commented out code still?

src/mod/channels.mod/userchan.c
```diff
@@ -220,9 +220,55 @@ static int u_equals_mask(maskrec *u, char *mask)
...
+    if (bang) {
+      size_t nicklen = (size_t)(bang - mask);
+
+      if (nicklen >= sizeof nick) {
+        nicklen = sizeof nick - 1;
+      }
+      memcpy(nick, mask, nicklen);
+      nick[nicklen] = 0;
+      if (nick[0]) {
+        m = find_member_from_nick(nick);
```
thommey, line 239: this searches the nickname in any channel instead of the channel we called u_match_mask for, what do you think about creating a new matching function that also takes a channel so we have a specific `m`?

src/mod/channels.mod/channels.h
```diff
@@ -65,6 +65,15 @@ struct udef_struct {
                                 * structures.                          */
 };
 
+/* List of extban flags that Eggdrop can actively enforce.
+ * Any extban flag not in this list is treated as sticky by default.
+ */
+static int extban_is_enforceable_flag(char flag, char account_extban_flag)
+{
+  return (flag == 'U' || (account_extban_flag && flag == account_extban_flag));
```
thommey: U is missing in the PR description

src/mod/channels.mod/cmdschan.c
```diff
@@ -174,13 +207,49 @@ static void cmd_pls_ban(struct userrec *u, int idx, char *par)
...
+            (me->funcs[IRC_CHECK_THIS_BAN]) (chan, s, sticky || extban_default_sticky);
+        } else
+          dprintf(idx, "The %c extban is not enabled on this server. Eggdrop will save this ban but only set it when on a server that has the %c flag enabled.\n", extbanflag, extbanflag);
+      }
     }
   }
 }
 
+
+static void cmd_pls_extban(struct userrec *u, int idx, char *par)
```
thommey, line 227: Do you want to keep this then?
If yes, I would consider adding cmd_mns_extban as well to correspond.

vanosg: .-ban works for all bans, extban or not. I acknowledge it is confusing, but I don't know what this command would add since you're either using a number or a mask, both of which work via .-ban

I'll add that I implemented cmd_pls_extban because it takes needing to know the extban prefix and specific banmask formatting out of the users hands... I thought this is a bonus. .+ban should still work if you want to do it manually as well.

src/mod/channels.mod/userchan.c
```diff
@@ -220,9 +220,55 @@ static int u_equals_mask(maskrec *u, char *mask)
...
+    if (extban_parse(rec->mask, &type, &arg)) {
+      me = module_find("server", 0, 0);
+      if (me && me->funcs && me->funcs[SERVER_GET_ISUPPORT]) { 
+        accountflag = isupport_get("ACCOUNTEXTBAN", strlen("ACCOUNTEXTBAN"));
+        return 0;
+      }
+      /* unknown account state never matches extbans */
+      if (!m || !m->account[0]) {
+        continue;
+      }
```
vanosg: @thommey I think this is old code and incorrect, should this come out?

src/mod/channels.mod/userchan.c
vanosg: This was supposed to be a 'just in case' errorcheck- if ACCOUNTEXTBAN isn't set, skip. But it "should" always be set via the defaults of ISUPPORT, no?

thommey: I mean that just returns 0 for the whole function always, doesn't seem like the program flow you want?

src/mod/channels.mod/userchan.c
thommey: the sanity check makes sense I think, I believe "no account" is internally "*", not sure if you handle that well down below

src/mod/channels.mod/channels.h
vanosg: "Currently supports only ACCOUNTEXTBAN flag and the U flag." :P

src/mod/channels.mod/userchan.c
vanosg: indeed, sir.

src/mod/channels.mod/userchan.c
vanosg: I think we could include a channel, but do we need to? I don't have a strong reason for this, but I feel finding any record for a nick is better than limiting it to the channel, just in case for some reason the record hasn't been created/already been deleted for the channel? It's not a strong case, but I don't know why we would limit it to just the channel

src/mod/irc.mod/chan.c
vanosg: fun fact, check_this_ban doesn't appear to be used anywhere else in code? is there some other newer code being used?

src/mod/channels.mod/cmdschan.c
thommey: .+ban also works on all bans, no? Maybe you can explain more why .+extban adds value but .-extban would not

src/mod/channels.mod/userchan.c
thommey: it's the consistency thing, this only grabs `m->account` from a "random" channel, I think when we did accounts we made sure it is always copied properly and should always be the same on all channels, then this is fine

src/mod/irc.mod/chan.c
thommey: It is used through irc.mod exports, see `IRC_CHECK_THIS_BAN`, but not (yet) by irc.mod directly

src/mod/channels.mod/cmdschan.c
vanosg: For sure- as a novice user, I might not know that $ and ~ are different on different servers, or what flags are allowed and not allowed on a server. .+extban checks all this; .+ban will just set what you set. In short, there is optionality in how to build a ban, but deleting a ban is still pretty simple and intuitive.

src/mod/channels.mod/userchan.c
vanosg: Ah, I think i fussed about this over there, too haha. nick will always be nick regardless of the channel. Tell me if you're very motivated to create a new function just for this (other places?), otherwise I'd just leave it as is because I don't see the harm (said every dev who ever introduced a vulnerability)

vanosg:
```
[00:57:15] Failed connect to dsfgsdgt34.dfsgdf (DNS lookup failed)
.+extban
You must be connected to a server with EXTBAN support to use +extban.
```

thommey (approved):
fine work!

PR #1909, opened 2026-05-16
webui.mod enhancement

michaelortmann:
Found by: thommey and Claude Code
Patch by: michaelortmann
Fixes: 

One-line summary:

Issue: https://text.is/66JQ3
This PR Fixes:
1. Blocker — src/tls.c ssl_init() X509 use-after-free on error.
- Fix use after free / fingerprint logging, implemented a TODO left by #1467
2. Blocker — src/tcldcc.c webui listen type works even without the webui module.
- Init `webui_dcc_telnet_hostresolved()` hook to `null_func` instead of 0
- But we dont need any fix here, because the listen type webui is only set for webui listen ports to begin with.
3. Should-fix — src/tls.c:1075-1104 HTTP-on-SSL handler hijacks every TLS listener, not just webui.
- Fixed SSL_R_HTTP_REQUEST logic to only apply to webui ports
- Response status HTTP 200 -> 400 for error msg "Error: HTTP request received on an SSL port, please try HTTPS"
- Leaking program name and version when stealth telnets is set to 1 is not a bug but a feature. nginx does the same.
- The write()/SSL_shutdown() was only suffering from a bad comment. I removed the comment, because the connection is properly closed with SO_LINGER + close(), verified with strace and tcpdump
4. killsock(): dont kill stdout, so that we keep stdout logging until the very last cycle of .die waving killsock(), it just doesnt make sense to kill stdout.
5. puts(): fix openssl error logging
6. puts(): fix memleak
7. Fix error logging for SSL_write() by doing ERR_clear_error() before that write call
8. Stop writing to sock after close notify during write and enhance logging
9. Fix fatal(): end all logging to dcc for killed sock
10. Fix memleak by calling changeover_dcc() before dcc[i].u is reused. this bug existed since at least eggdrop 1.10.1.

### DIFF
--- a/src/dcc.c
+++ b/src/dcc.c
@@ -1346,6 +1346,8 @@ void dcc_telnet_hostresolved2(int i, int idx) {
     ident_target_port = getenv("EGGDROP_TEST") ? 1113 : 113;
 
   snprintf(userhost, sizeof userhost, "telnet@%s", dcc[i].host);
+  changeover_dcc(i, &DCC_IDENTWAIT, 0);
+
   /* Skip ident lookup if disabled */
   if (identtimeout <= 0) {
     dcc[i].u.ident_sock = dcc[idx].sock;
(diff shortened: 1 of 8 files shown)


michaelortmann:
ready for review

PR #1920, opened 2026-07-12
Task/normalizeextban

### LINKED ISSUE #1914: Extbans placed by users on a channel through ChanServ are not properly expired, stays active forever
Extbans placed by users on a channel *through ChanServ* are not properly expired, stays active forever.
- The channel is configured with a `ban-time` of `1` minute for speedier testing/reproduction.
- Set the ban: `/CS BAN #channel +b m:*!*@example.com` (observe ChanServ is setting the ban)
- Wait approximately 1 to 2 minutes (as expiration doesn't happen that precisely to the minute).
- Observe that the ban is not being removed by the bot.

This is possibly a regression caused by #1880 as we didn't have this issue before.

This is tested with InspIRCd 4.10.1 and the default setting of `<options:extbanformat="name">` in [`inspircd.conf`](https://github.com/inspircd/inspircd/blob/insp4/docs/conf/inspircd.example.conf).

Due to the above mentioned InspIRCd setting, an interesting observation I made is, if the extban is set manually using their letter, for example for `mute` it is `m`, then the ircd transforms `/MODE #channel +b m:*!*@example.com` into `/MODE #channel +b mute:*!*@example.com` which is functionally the same, but then the ban does expire properly and is removed by Eggdrop as expected. However, if it is set via ChanServ, this transformation doesn't happen and the bot does not expire the ban and remains active forever.

vanosg:
Found by: Robby
Patch by: Geo
Fixes: #1914 

One-line summary:
Have extbans follow normal ban expiration

The original intent, as included in the documentation ;) , was to only expire extbans that eggdrop itself could take some action on (ie, removing a user from the channel due to a matching hostmask). Because there are many conflicting extban modes and no clear intent for ircds to add flags similar to ACCOUNTEXTBAN, discussion moved to treating extbans just like normal bans, and respecting the dynamicban channel setting. If a user wants an extban to be set that is not removed even though dynamicbans is set, then it needs to have the sticky flag added to it so that it is not removed.

### DIFF
--- a/src/mod/channels.mod/channels.c
+++ b/src/mod/channels.mod/channels.c
@@ -116,14 +116,14 @@ int extban_parse(const char *mask, char *type, const char **arg) {
   }
 
 /* If EXTBAN is unknown (e.g. before connect), retain the historical fallback
- * so stored prefixed extbans can still be parsed.
+ * so stored prefixed extbans can still be parsed. With any non-alnum prefixchar.
  */
-  if ((!value || !value[0]) && isalnum((unsigned char) mask[1]) &&
-      mask[2] == ':') {
+  if ((!value || !value[0]) && !isalnum((unsigned char) mask[0]) &&
+      isalnum((unsigned char) mask[1]) && mask[2] == ':') {
     if (type)
-      *type = mask[0];
+      *type = mask[1];
     if (arg)
-      *arg = mask + 2; /* a:x!y@z -- mask+2 = x */
+      *arg = mask + 3; /* ~a:x!y@z -- mask+2 = x */
     return 1;
   }
   return 0;
@@ -157,7 +157,7 @@ static int is_extban_mask(const char *mask)
 }
 
 /* Extban prefix from ISUPPORT EXTBAN, if present.
- * EXTBAN grammar is [prefix],<types> 
+ * EXTBAN grammar is [prefix],<types>
  */
 static void get_extban_prefix(char *prefix)
 {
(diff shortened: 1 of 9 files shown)


src/mod/channels.mod/channels.c
```diff
@@ -116,14 +116,14 @@ int extban_parse(const char *mask, char *type, const char **arg) {
   }
 
 /* If EXTBAN is unknown (e.g. before connect), retain the historical fallback
- * so stored prefixed extbans can still be parsed.
+ * so stored prefixed extbans can still be parsed. With any non-alnum prefixchar.
```
vanosg: update comment

src/mod/channels.mod/channels.c
```diff
@@ -116,14 +116,14 @@ int extban_parse(const char *mask, char *type, const char **arg) {
...
-  if ((!value || !value[0]) && isalnum((unsigned char) mask[1]) &&
-      mask[2] == ':') {
+  if ((!value || !value[0]) && !isalnum((unsigned char) mask[0]) &&
+      isalnum((unsigned char) mask[1]) && mask[2] == ':') {
     if (type)
-      *type = mask[0];
+      *type = mask[1];
     if (arg)
-      *arg = mask + 2; /* a:x!y@z -- mask+2 = x */
+      *arg = mask + 3; /* ~a:x!y@z -- mask+2 = x */
```
vanosg: update comment

PR #1926, opened 2026-08-02
Make SIGILL fatal instead of looping in its signal handler

### LINKED ISSUE #1640: SIGILL -> endless loop
eggdrop can run into an endless loop in its ill signal handler.

for a minimal demo, i introduced a malicious division by zero into eggdrops source `dcc.c:dcc_remove_lost():for (i = 1 / 0; i < dcc_total; i++) {`, compiled and run it. result:

```
[04:12:53] triggering bind quotepong_servermsg
[04:12:53] triggered bind quotepong_servermsg, user 0.004ms sys 0.000ms
[04:12:53] -NOTICE- *** Found your hostname: localhost
[04:12:53] * Please REPORT this BUG!
[04:12:53] * Check doc/BUG-REPORT on how to do so.
[04:12:53] * Last bind (may not be related): quotepong_servermsg
[04:12:53] * Please REPORT this BUG!
[04:12:53] * Check doc/BUG-REPORT on how to do so.
[04:12:53] * Last bind (may not be related): quotepong_servermsg
[04:12:53] * Please REPORT this BUG!
[04:12:53] * Check doc/BUG-REPORT on how to do so.
[04:12:53] * Last bind (may not be related): quotepong_servermsg
[...]
```

Default action for SIGILL is core dump / stack trace. Our current signal handler tries to ignore the signal. Effect is, we can no longer debug SIGILL, no stack trace, to useful information about the cause or even what happened is printed.

willbuckner:
Found by: michaelortmann 
Patch by: willbuckner
Fixes: #1640

One-line summary: Make SIGILL fatal instead of looping in its signal handler

got_ill() logged the signal and returned. Returning from a SIGILL handler re-executes the faulting instruction, so a real illegal instruction re-raised the signal immediately, spamming the log in an endless loop, and no core dump or stack trace was ever produced.

Treat SIGILL like SIGBUS/SIGSEGV: write the DEBUG file, log the crash and re-raise the signal with its default action restored so the bot dies with a core dump.

Restore the default action explicitly instead of relying on SA_RESETHAND alone: at least on macOS the handler stays installed when the signal is re-raised from within the handler, which would turn the re-raise into another endless loop.

This removes the (now pointless) sigill Tcl evnt bind; the bot no longer survives SIGILL.

Test cases demonstrating functionality (if applicable): N/A

### DIFF
--- a/doc/sphinx_source/using/tcl-commands.rst
+++ b/doc/sphinx_source/using/tcl-commands.rst
@@ -3487,7 +3487,6 @@ The following is a list of bind types and how they work. Below each bind type is
 
           sighup            - called on a kill -HUP <pid>
           sigterm           - called on a kill -TERM <pid>
-          sigill            - called on a kill -ILL <pid>
           sigquit           - called on a kill -QUIT <pid>
           save              - called when the userfile is saved
           rehash            - called just after a rehash
--- a/doc/tcl-commands.doc
+++ b/doc/tcl-commands.doc
@@ -3743,7 +3743,6 @@ the Tcl proc, and an explanation.
 
       sighup            - called on a kill -HUP <pid>
       sigterm           - called on a kill -TERM <pid>
-      sigill            - called on a kill -ILL <pid>
       sigquit           - called on a kill -QUIT <pid>
       save              - called when the userfile is saved
       rehash            - called just after a rehash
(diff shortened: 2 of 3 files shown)


vanosg:
ok, I'll bite... Wcc?

willbuckner:
yes :P

